PR #4, opened 2026-05-12
fix: LSI writes are always synchronous regardless of gsi_propagation_…

LeeroyHannigan:
…delay_ms

LSI updates were incorrectly subject to the GSI propagation delay, causing queries with `ConsistentRead: true` to return stale results. In DynamoDB, LSIs are always strongly consistent with the base table.
  
Fix: skip the delay check in `sync_indexes` for LSI index types. One-line change, LSIs are now always written within the same transaction as the base table write, matching DynamoDB semantics.


jcshepherd (approved):

pdf-amzn:
This was folded in as a part of a larger merge, correct? Or did this one get lost in the shuffle and we still need to land it?

LeeroyHannigan:
Its in, went in with the larger merge.

pdf-amzn:
got it, closing

PR #11, opened 2026-05-13
Fix HTTP→HTTPS redirect to preserve client Host header & path

pdf-amzn:
The plain-HTTP redirect was using the server's local bind address
(e.g., 10.2.3.4:8000) in the Location header. When accessing via SSH
port forwarding (127.0.0.1:8000 → remote), this broke the redirect by
sending the browser to the remote host's real IP.

Now reads the HTTP request to extract the Host header and request path,
preserving the client's view of the hostname. Falls back to local_addr
only if the Host header is absent.

Also adds the Secure flag to the session cookie (TLS is mandatory).

### DIFF
--- a/crates/server/src/console/pages/auth_pages.rs
+++ b/crates/server/src/console/pages/auth_pages.rs
@@ -82,7 +82,7 @@ pub async fn login_submit<
 
     let (token, _csrf) = state.sessions.create(identity).await;
     let cookie = format!(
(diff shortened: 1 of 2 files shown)


pdf-amzn:
Abandoning this work. I haven't reproduced the set of circumstances that caused me to want this change in the first place, and there's a slight chance that someone could craft a request that would cause the redirect to go somewhere else.

pdf-amzn:
An interesting question to consider later is whether we should just not listen/redirect at all, and trust browsers to try https even if a user specified http.

PR #29, opened 2026-05-18
fix(storage): return winning item on UpdateItem insert race loss

pdf-amzn:
When INSERT ON CONFLICT DO NOTHING returns 0 rows (another transaction
inserted between our SELECT FOR UPDATE and INSERT), UpdateItem previously
returned ConditionFailed(None). This meant ReturnValuesOnConditionCheck-
Failure=ALL_OLD would return no item for the race-loss case.

Now mirrors PutItem's existing pattern: after rows_affected() == 0,
fetch the winner row and include it in the ConditionFailed error. Applied
to both the sort-key and PK-only code paths.


LeeroyHannigan:
I forgot updateItem... We should probably also check transactions returnvalues

jcshepherd (approved):

PR #32, opened 2026-05-18
docs: Adding AGENTS.md

jcshepherd:


pdf-amzn (approved):
We should add something about `cargo fmt` and `clippy` in a followup.

Approved.

jcshepherd:
I pulled in a blurb about cargo fmt and clippy from PR 34 (demohan@).

PR #44, opened 2026-05-19
ci: add GitHub Actions for fmt, clippy, and test checks

LeeroyHannigan:
Three separate workflows so each appears as its own status check on PRs:
- Format: cargo fmt --all -- --check
- Clippy: cargo clippy --all-targets -- -D warnings
- Test: cargo test --workspace

Uses rust-cache for faster CI runs on clippy and test.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,0 +1,16 @@
+name: Clippy
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo clippy --all-targets -- -D warnings
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -9,8 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: "1.85.0"
           components: clippy
       - uses: Swatinem/rust-cache@v2
       - run: cargo clippy --all-targets -- -D warnings
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -9,9 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: "1.85.0"
           components: clippy
       - uses: Swatinem/rust-cache@v2
       - run: cargo clippy --all-targets -- -D warnings
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -1,0 +1,15 @@
+name: Format
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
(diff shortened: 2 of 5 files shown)


LeeroyHannigan:
This PR is dependent on #41 being merged first.

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo clippy --all-targets -- -D warnings
```
jcshepherd, line 16: What is the effect of -D warnings? (Unsure of the semantics of deny vs forbid.)

jcshepherd (commented):
Looks good but would like to understand -D warnings on clippy: specifically, is it CI breaking or not?

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,16 @@
+name: Clippy
+on:
+  pull_request:
+  push:
```
jcshepherd, line 4: Do these actions happen immediately before the push, or after?

.github/workflows/clippy.yml
LeeroyHannigan: -D promotes warnings to errors. Still allows overrides.
-F would mean we cant use annotations to override clippy errors such as: `#[allow(clippy::too_many_arguments)]`, as it would treat them as failures also.

.github/workflows/clippy.yml
LeeroyHannigan: After. Runs when a PR get created/updated. And runs when we merge to main, but does not rollback if main has failed.

jcshepherd: So before and after? :-)  (Before a PR is merged to main, and after.) Cool.

jcshepherd (approved):
Okay with me, but I'll let you do the merge once you and Paul are on the same page about CI.

.github/workflows/fmt.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
pdf-amzn, line 11: Should this be @v6?

.github/workflows/clippy.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [main]
+
+jobs:
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: "1.85.0"
```
pdf-amzn, line 14: So this is interesting. I think our current main HEAD (2f65e1b) generates warnings (that block due to -D) in GitGub with this toolchain but does not generate warnings locally on my Linux box which has Rust 1.95.0 installed. Thoughts? Perhaps we want `clippy` and `fmt` to use Rust latest but `test` to use 1.85.0 (the oldest version we advertise support for)?

pdf-amzn: huh. Never mind. The latest attempt to run the checks passed. So perhaps something was out of sync somewhere. Merging this before things get out of sync again.

LeeroyHannigan: Clippy was failing because it was missing something locally for me, that the LATEST version caught here. I thought it was better to fix those, and use LATEST as the action, rather than pinning a version that will shoot us in the foot later.

pdf-amzn (approved):

.github/workflows/clippy.yml
pdf-amzn: I'm assuming that this is pinning us to rust 1.85.0, yes?

PR #45, opened 2026-05-19
perf: disable SQLx connection pool ping, add 30min max_lifetime

LeeroyHannigan:
Disable `test_before_acquire` to eliminate a full round-trip on every connection checkout. At high QPS connections are never idle long enough to go stale, making the ping pure overhead (~0.85ms per request).

Add `max_lifetime` of 30 minutes to gracefully retire long-lived connections, handling failover scenarios without per-request cost.


jcshepherd:
Can you add items to the Taskei board to:
(1) Pull the connection lifetime out to a constant.
(2) Gameday an intermittent PSQL outage (or network blip) and confirm xdb recovery afterwards.

jcshepherd (approved):

PR #47, opened 2026-05-19
Fix run-tests for MacOS

pdf-amzn:


jcshepherd (approved):
Built and tested locally (on ALinux).

PR #89, opened 2026-05-21
fix(storage-postgres): build socket-aware connection URLs (#52)

### LINKED ISSUE #52: Init-generated connection string fails at daemon startup with "empty host" when --pg-host is a Unix socket path
## Summary

`extenddb init` writes a PostgreSQL connection string into `extenddb.toml` and into the catalog `settings.data_database_connection_string` field that fails to parse at daemon startup. Following the documented peer-authentication path in `docs/getting-started.md` produces an install the daemon refuses to start.

## Repro

Ubuntu 24.04, fresh Postgres 16:

```bash
sudo apt-get install -y postgresql postgresql-contrib
echo 'local all extenddb scram-sha-256' \
  | sudo tee -a /etc/postgresql/16/main/pg_hba.conf
sudo systemctl reload postgresql

cargo build --release
./target/release/extenddb init \
    --pg-host /var/run/postgresql \
    --extenddb-pass 'extenddb-test-pw'
# init succeeds, prints admin credentials, writes extenddb.toml

./target/release/extenddb serve --config extenddb.toml
# Server failed to start: daemon process exited during startup.

sudo journalctl -t extenddb -n 5
# extenddb fatal: Failed to connect to postgres: Connection error:
# error with configuration: empty host
```

## What `init` wrote

`extenddb.toml`:
```toml
[storage.postgres]
connection_string = "postgresql://extenddb:***@/var/run/postgresql:5432/extenddb_catalog"
```

`extenddb_catalog.settings`:
```
data_database_connection_string |
postgresql://extenddb:***@/var/run/postgresql:5432/extenddb
```

The `@/var/run/postgresql:5432/` segment is the problem. Unencoded forward slashes in the host portion make the URL ambiguous, and the `sqlx`/`libpq` parser the daemon uses rejects it with `empty host`. The init code apparently builds its connection from discrete `--pg-host`/`--pg-port` flags rather than from the URL it later writes, so init succeeds and the daemon fails.

## Workaround

After hitting this, the user has to:

1. Edit `extenddb.toml` to change the connection string.
2. `psql extenddb_catalog -c "UPDATE settings SET value=... WHERE key='data_database_connection_string';"` to fix the data-database connection string the daemon reads separately.

After both edits, the daemon starts.

## Suggested fix

The connection-string assembly should either percent-encode the host portion (`postgresql://extenddb:***@%2Fvar%2Frun%2Fpostgresql:5432/extenddb_catalog`) or use the `host=` parameter form (`postgresql://extenddb:***@localhost:5432/extenddb_catalog?host=/var/run/postgresql`). A test that round-trips a generated URL through `sqlx::postgres::PgConnectOptions::from_url` would catch this.

## Environment

- ExtendDB: v0.1.0 (HEAD as of release tag)
- OS: Ubuntu 24.04.4 LTS, aarch64
- PostgreSQL: 16.13 (Ubuntu 16.13-0ubuntu0.24.04.1)
- Rust: 1.94.1

flightlesstux:
## What

Fix `extenddb init --pg-host <socket-path>` producing a connection URL that the daemon's PostgreSQL driver rejects at startup with `error with configuration: empty host`.

## Why

Closes #52

`init` was writing URLs of the form `postgresql://user:pw@/var/run/postgresql:5432/db` into both `extenddb.toml` and the catalog `data_database_connection_string` row. The unencoded forward slashes in the host segment break standard URL parsing, so `sqlx::PgPool::connect` rejected the URL at daemon startup. Init succeeded, `serve` failed, and users had to hand-edit both the TOML and the catalog row to recover.

## Changes

- `crates/storage-postgres/src/config.rs`: new `build_connection_url()` helper. When `host` starts with `/` it emits the documented PostgreSQL socket form `postgresql://user:pw@localhost:5432/db?host=/var/run/postgresql`. TCP hosts keep the classic form. `parse_connection_string()` now honors a `?host=<path>` query parameter with last-value-wins libpq semantics; an empty value falls back to the URL host so existing parse callers stay backwards compatible.
- `crates/storage-postgres/src/bootstrapper.rs`: `app_connection_url()` delegates to the new builder. No other behavior change.

## Testing done

Added 10 unit tests in `crates/storage-postgres/src/config.rs`:

- TCP and socket URL build shapes.
- TCP URL parse path unchanged.
- Socket URL parse extracts the `?host=` value.
- Empty `?host=` falls back to the URL host.
- Unknown query parameters are ignored.
- Round-trip through the custom parser for both TCP and socket URLs.
- Round-trip through `sqlx::postgres::PgConnectOptions::from_str` for both URL shapes. This is the same parser the daemon uses, so it directly verifies the fix at the failure layer.

Results:

\`\`\`
cargo test --workspace --lib        382/382 passed
cargo clippy -p extenddb-storage-postgres --all-targets -- -D warnings   clean
cargo fmt -p extenddb-storage-postgres -- --check  (on the two changed files)   clean
\`\`\`

Full \`cargo test --workspace\` (including pytest/external integration suites) was not exercised in this branch because it requires a running PostgreSQL instance and provisioned test credentials.

## Checklist

- [x] All workspace lib tests pass (\`cargo test --workspace --lib\`)
- [x] Changed files are formatted (\`cargo fmt\`)
- [x] Clippy is clean on the modified crate (\`-D warnings\`)
- [x] Added unit tests for new functionality
- [ ] No documentation changes needed (URL form change is internal; observable behavior is identical for TCP setups and now actually works for socket setups)
- [ ] No breaking changes

### DIFF
--- a/crates/storage-postgres/src/bootstrapper.rs
+++ b/crates/storage-postgres/src/bootstrapper.rs
@@ -15,6 +15,7 @@ use sqlx::postgres::{PgConnectOptions, PgPoolOptions};
 use tokio::sync::OnceCell;
 
 use crate::CATALOG_VERSION;
+use crate::config::build_connection_url;
 use crate::migrations;
 
 /// Utilities for bootstrapping a PostgreSQL backend store.
@@ -82,11 +83,10 @@ impl PostgresBootstrapper {
 
     /// Build the connection URL for the application user and a named database.
     fn app_connection_url(&self, database: &str) -> String {
-        format!(
-            "postgresql://{}:{}@{}:{}/{}",
-            self.config.app_user,
-            self.config.app_password,
-            self.config.host,
+        build_connection_url(
+            &self.config.app_user,
+            &self.config.app_password,
+            &self.config.host,
             self.config.port,
             database,
         )
--- a/crates/storage-postgres/src/config.rs
+++ b/crates/storage-postgres/src/config.rs
@@ -45,9 +45,32 @@ pub struct ConnParts {
     pub database: String,
 }
 
+/// Build a `PostgreSQL` connection URL from discrete components.
+///
+/// When `host` starts with `/` (a Unix socket directory), the host is moved into
+/// the `?host=` query parameter and the URL host is set to `localhost`. This is
+/// the PostgreSQL-documented form for socket connections and is accepted by
+/// `sqlx::postgres::PgConnectOptions::from_url` and libpq. For TCP hosts the
+/// classic `postgresql://user:password@host:port/database` form is emitted.
+pub fn build_connection_url(
+    user: &str,
+    password: &str,
+    host: &str,
+    port: u16,
+    database: &str,
+) -> String {
+    if host.starts_with('/') {
+        format!("postgresql://{user}:{password}@localhost:{port}/{database}?host={host}")
+    } else {
+        format!("postgresql://{user}:{password}@{host}:{port}/{database}")
+    }
+}
+
 /// Parse host, port, user, password, and database from a `PostgreSQL` connection string.
 ///
-/// Handles the standard `postgresql://user:pass@host:port/db` format.
+/// Handles the standard `postgresql://user:pass@host:port/db` format and the
+/// socket-style `postgresql://user:pass@localhost:port/db?host=/path` form
+/// produced by [`build_connection_url`] when the host is a Unix socket path.
 ///
 /// # Errors
 ///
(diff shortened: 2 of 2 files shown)


jcshepherd:
Hi Ercan! Thanks for the PR. I'll let the reviewers weigh in on this approach vs https://github.com/ExtendDB/extenddb/pull/92 .

jcshepherd:
Resolved with https://github.com/ExtendDB/extenddb/pull/92 . Thanks!

PR #103, opened 2026-05-22
fix: update item wcu

### LINKED ISSUE #95: [Bug] update_item WCU incorrect when ReturnValues not set
## Describe the bug

Adding multiple KB of data to an item still results in reported write capacity units of 1 when return values does not include return new items. this is because the value is calculated based on the size of the new_item [here](https://github.com/ExtendDB/extenddb/blob/2f65e1bdfe396364417af75387a0e9e7eb19d668/crates/engine/src/update_item.rs#L190), but that value is only a Some when the api needs to return new values. A similar issue likely occurs when old_value is not set.

## To Reproduce

1. add an item
2. run update item and add an attribute with a 4KB string, do not include the return new images
3. notice reported write capacity units remains at 1
4. run again, include the return new images
5. noticed reported write capacity is 4

## Expected behavior

DynamoDB reports 4 WCU consumed

## Actual behavior

ExtendDB reports 1 WCU consumed

## Environment

- ExtendDB version: 0.1.0
- Operating system: macOS 26
- Rust version (if building from source): 1.94
- Client SDK/driver and version: CLI

## Checklist

- [x] I can reproduce this with the latest release

LeeroyHannigan:
## What
  
Always fetches the new item from storage in `update_item` so WCU is calculated from the actual item size, regardless of `ReturnValues`.

## Why

Closes #95

WCU was calculated from `new_item` which was only populated when `ReturnValues` included the new image. Without it, WCU always reported 1 regardless of actual item size.

## Testing done

- Verified against real DynamoDB: 4KB update reports 5 WCU with and without ReturnValues
- Verified ExtendDB matches (5.0 WCU no return, 9.0 WCU with return for 8KB item)
- `cargo test --workspace` passes
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --all -- --check` clean

## Checklist

---

### DIFF
--- a/crates/engine/src/update_item.rs
+++ b/crates/engine/src/update_item.rs
@@ -152,10 +152,6 @@ pub async fn handle_update_item(
         input.return_values,
         ReturnValues::AllOld | ReturnValues::UpdatedOld
     );
-    let return_new = matches!(
-        input.return_values,
-        ReturnValues::AllNew | ReturnValues::UpdatedNew
-    );
 
     let view_type = stream_capture::stream_view_type(&key_info);
     let stream = view_type.map(|vt| extenddb_storage::StreamCapture {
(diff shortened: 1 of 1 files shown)


jcshepherd (approved):

LeeroyHannigan:
Closed by #139

PR #109, opened 2026-05-22
feat: add SQLite storage backend (extenddb-storage-sqlite)

ivanbarlog:
## Summary

- Introduces a new `extenddb-storage-sqlite` crate implementing the full DynamoDB-compatible storage layer backed by SQLite via `sqlx`
- Implements all required traits: `TableEngine`, `DataEngine`, `MetadataEngine`, `StreamEngine`, `BackupEngine`, `WorkerStore`, plus all catalog/management/auth stores
- Self-registers via the `inventory` crate alongside the existing PostgreSQL backend — no changes needed in the engine or server crates

## Key design decisions vs the PostgreSQL backend

| Concern | SQLite approach |
|---|---|
| Placeholders | `?` positional (not `$N`) |
| Concurrency | Single WAL-mode pool; writer serialization replaces `FOR UPDATE` |
| GSI/LSI updates | Synchronous on every write (no async propagation queue) |
| Stream sequencing | `seq_counters` table with atomic UPDATE+SELECT (no PostgreSQL sequences) |
| Parallel scan | `rowid % total_segments` (vs `hashtext(pk) % total_segments`) |
| Pagination | `(pk > ? OR (pk = ? AND sk > ?))` OR-expansion (SQLite lacks tuple comparison) |
| String begins-with upper bound | `char(1114111)` (not `chr()` which is PostgreSQL) |
| Pool | Single `self.pool` for all catalog + data ops |

## Test plan

- [ ] `cargo build -p extenddb-storage-sqlite` — clean build, zero errors
- [ ] `cargo clippy --all-targets -- -D warnings` — zero warnings/errors
- [ ] `cargo test --workspace` — 379 tests pass across all crates
- [ ] Wire up `backend = "sqlite"` in `extenddb.toml` and run `devtools/run-tests --extenddb --pytest` against a live SQLite instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -852,6 +852,7 @@ dependencies = [
  "extenddb-server",
  "extenddb-storage",
  "extenddb-storage-postgres",
+ "extenddb-storage-sqlite",
  "libc",
  "rand 0.9.4",
  "rcgen",
@@ -976,6 +977,34 @@ dependencies = [
  "async-trait",
  "base64 0.22.1",
  "bcrypt",
+ "crc32fast",
+ "extenddb-auth",
+ "extenddb-core",
+ "extenddb-storage",
+ "futures",
+ "inventory",
+ "rand 0.9.4",
+ "serde",
+ "serde_json",
+ "sqlx",
+ "time",
+ "tokio",
+ "toml",
+ "tracing",
+ "uuid",
+ "zeroize",
+]
+
+[[package]]
+name = "extenddb-storage-sqlite"
+version = "0.1.0"
+dependencies = [
+ "aes-gcm",
+ "anyhow",
+ "async-trait",
+ "base64 0.22.1",
+ "bcrypt",
+ "bigdecimal",
  "crc32fast",
  "extenddb-auth",
  "extenddb-core",
@@ -1654,6 +1683,7 @@ version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2e99fb7a497b1e3339bc746195567ed8d3e24945ecd636e3619d20b9de9e9149"
 dependencies = [
+ "cc",
  "pkg-config",
  "vcpkg",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "crates/engine",
     "crates/storage",
     "crates/storage-postgres",
+    "crates/storage-sqlite",
     "crates/auth",
     "crates/server",
     "crates/bin",
@@ -24,6 +25,7 @@ extenddb-core = { path = "crates/core" }
 extenddb-engine = { path = "crates/engine" }
 extenddb-storage = { path = "crates/storage" }
 extenddb-storage-postgres = { path = "crates/storage-postgres" }
+extenddb-storage-sqlite = { path = "crates/storage-sqlite" }
 extenddb-auth = { path = "crates/auth" }
 extenddb-server = { path = "crates/server" }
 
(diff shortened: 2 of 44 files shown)


LeeroyHannigan:
Thanks for putting this together, the design table is well thought out and the approach to SQLite's limitations (WAL serialization, OR-expansion for pagination, positional placeholders) shows you've done your homework.

That said, we're not in a position to take on new storage backends right now. We're still working through DynamoDB behavioral parity on the existing Postgres backend, and we don't yet have a clear story for how multiple backends should be organized, tested, and maintained long-term (shared conformance harness, CI matrix, feature parity guarantees between backends, etc.).

Once we've nailed that down, we'll have a better foundation for community-contributed backends like this. I'd suggest keeping your fork alive and we can revisit when the time is right.

ivanbarlog:
Goal was clear:

> /goal implement extension for sqlite for extenddb

jcshepherd:
This is awesome and I'll find some time to look through it. It's especially valuable because it's an existence proof of the practicality of adding additional backends: thanks very much for that alone!

Like Lee said, we need to get our ducks in a little better order for properly qualifying and releasing backends, but I hope you'll bear with us while we do that so that this contribution can be part of the project.

One head's up: we are leaning in the direction of keeping the reference PostgreSQL backend implementation in the main extenddb repo, and vending additional backends as library crates from their own repos, partly to prevent the main project from acquiring a bunch of dependencies on different db drivers, and partly as a forcing function to ensure proper decoupling between the request handling and routing aspects of ExtendDB from the db-specific implementation details.

I will write a proper RFC to outline this and get consensus on it, but wanted you to be aware of how we're thinking about it.

Thanks!

Iamrodos:
I'm really keen on SQLite making the cut here. I have been waiting for someone to start it. 

The local dev and CI story is one of ExtendDB's strongest use cases, and for that I'd reach for an in-process backend over a running Postgres every time.

Whilst it makes sense having Postgres staying as the in-core reference, two things jump out at me:

1. If SQLite lives in its own repo, that's a fork to keep alive and extra friction for what's actually a primary use case. Feels a bit odd to push a core use case out to the bring-your-own tier.

2. So @jcshepherd I'd love the **RFC to pin down a promotion path**. What gets a backend into core, especially when it's serving a primary use case like this one.

Thanks for doing this in the open.

LeeroyHannigan:
> I'm really keen on SQLite making the cut here. I have been waiting for someone to start it.
> 
> The local dev and CI story is one of ExtendDB's strongest use cases, and for that I'd reach for an in-process backend over a running Postgres every time.
> 
> Whilst it makes sense having Postgres staying as the in-core reference, two things jump out at me:
> 
>     1. If SQLite lives in its own repo, that's a fork to keep alive and extra friction for what's actually a primary use case. Feels a bit odd to push a core use case out to the bring-your-own tier.
> 
>     2. So @jcshepherd I'd love the **RFC to pin down a promotion path**. What gets a backend into core, especially when it's serving a primary use case like this one.
> 
> 
> Thanks for doing this in the open.

Every backend now gets into core. We will use feature flags for each backend. Then each will be pushed to docker and crates.io to allow users to easily consume it. We're still weeding out the details, but we're not far away from landing additional backends and docker.

Iamrodos:
@LeeroyHannigan @jcshepherd has this been superseded by #182?

The design memo there notes the SQL logic is "ported from PR #109 and corrected/retargeted", and #182 lands in-tree under RFC-0002.

If that's right, probably worth closing this one out. Either way, plenty of us are keen for SQLite support in some form and it's good to see it moving.

LeeroyHannigan:
@Iamrodos yes, this was superseded by #182 

@ivanbarlog thank you for your contribution, the work for SqlLite was already in progress, however at that time, like I mentioned, we had planned on separate repositories for each backend. Then we made the decision for in-tree backends.

PR #128, opened 2026-05-25
fix: persist GSI queue to PostgreSQL for crash safety

### LINKED ISSUE #125: [Bug] Data Loss / Inconsistency on process exit
## Describe the bug

If I restart the process I can see missing data in the GSIs. It appears delayed writes to GSIs are not crash safe and I do not see any repair or recovery mechanism.

In addition, I see that GSI write delays can be over 10 minutes. Possibly longer but I did not actively monitor. This is an unacceptable data inconsistency compared to managed DynamoDB.

## To Reproduce

1. Create a table with 3 GSIs
2. Load 20GB of data
3. Put an item. Inspect table data and GSI data with queries on an interval to check delay for GSI propagation.
4. Load another 20GB of data.
5. Kill and restart the process. wait. GSIs never become consistent with table.

## Expected behavior

AWS DynamoDB has back pressure from the GSI to the table when latency breaches a certain level to have a bounded latency. ExtendDB as a replacement must offer a reasonable level of bounded latency.

GSI permanent inconsistency on crash or restart should be treated as a small step below a data loss condition and should never happen.

## Actual behavior

No back pressure for GSI catchup leads to unbounded propagation delays. Restarts result in permanent inconsistency and loss of data in the GSI without a repair or recovery mechanism.

## Environment

- ExtendDB version: 140a1e5
- Operating system: macOS 26
- Rust version (if building from source): 1.94
- Client SDK/driver and version: http

## Additional context

Effectively lost data would need a user to scan and re-touch every item to get GSIs back to consistent.

## Checklist

- [ ] I have searched existing issues for duplicates
- [ ] I have included the ExtendDB version and environment details
- [ ] I can reproduce this with the latest release

LeeroyHannigan:
## What

Replaces the in-memory `VecDeque` GSI propagation queue with a PostgreSQL-backed **persistent, self-describing** queue (`gsi_pending`). Pending GSI updates are enqueued in the same transaction as the base write and applied by background workers, surviving process crash/restart with no lost updates.

Key design:
- New `gsi_pending` table (data migration `002_gsi_pending.sql`).
- **Self-describing rows:** each row carries an `index_context` snapshot (base key schema, attribute definitions, target index defs) captured at enqueue, so workers apply with **zero catalog reads**, no metadata cache. A GSI's key schema/projection are immutable after creation, so the snapshot is never stale.
- **Crash-safe apply:** a worker claims, applies, and deletes each row in a single transaction (`SELECT … FOR UPDATE SKIP LOCKED` → index writes → `DELETE` → `COMMIT`). A crash or error before commit rolls the unit back and the row is reprocessed.
- **Per-key FIFO ordering:** rows are partitioned by a stable hash of the base key; each worker owns one partition and drains it in `id` order, so successive updates to one item apply in order (matching the old in-memory queue) while distinct keys propagate concurrently.
- Propagation delay enforced by `ready_at`; the worker sleeps until the next due row rather than polling.
- Dropped-index race (table deleted mid-flight) handled per index via a `SAVEPOINT` (SQLSTATE 42P01 → skip, row consumed).
- Data migrations are tracked in the data DB's `schema_history` (registry), with safe adoption of pre-tracking deployments.

## Why

Closes #125. The in-memory queue lost all pending GSI updates on crash/restart, causing permanent GSI inconsistency with no recovery path.

## Note on this branch

Merges latest `main`. Beyond the original change it addresses review feedback (removes the index-metadata cache in favour of self-describing rows; removes the dead `index_name` field) and hardens crash safety and per-key ordering, with integration-test coverage.

## Testing done

- `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --workspace` — pass
- New PG integration tests (`tests/test_gsi_async_queue.py`): crash recovery
  (SIGKILL → recovered after restart), per-key FIFO ordering, dropped-index
  consume; plus data-migration tracking (`test_cli_lifecycle.py`)
- Python integration suites — pass;

## Checklist

- [x] Clippy is clean

## Breaking changes

None to the API. `002_gsi_pending.sql` adds the `gsi_pending` table (with a `worker_partition` column); it is applied via the tracked data migration on `init`/`migrate`. Existing deployments gain crash-safe GSI propagation
transparently.

### DIFF
--- a/crates/storage-postgres/data_migrations/002_gsi_pending.sql
+++ b/crates/storage-postgres/data_migrations/002_gsi_pending.sql
@@ -1,0 +1,34 @@
+-- Copyright 2026 ExtendDB contributors
+-- SPDX-License-Identifier: Apache-2.0
+-- Persistent queue for async GSI propagation.
+--
+-- Rows are inserted atomically within the base write transaction and consumed
+-- by background workers in a single claim+apply+delete transaction, so an
+-- uncommitted claim rolls back and the row is reprocessed after a crash.
+--
+-- Each row is self-describing: `index_context` carries the base key schema,
+-- attribute definitions, and the target index definitions captured at enqueue
+-- time, so workers apply updates with zero catalog reads. A GSI's key schema
+-- and projection are immutable after creation, so the snapshot can never be
+-- stale relative to the live index definition.
+--
+-- `worker_partition` is a stable hash of the base table key. All updates to a
+-- given base item share a partition and are consumed by a single worker in `id`
+-- order, preserving per-key FIFO ordering (so successive updates to the same
+-- item never race in the index and converge to the latest state).
+
+CREATE TABLE IF NOT EXISTS gsi_pending (
+    id BIGSERIAL PRIMARY KEY,
+    table_id TEXT NOT NULL,
+    worker_partition INTEGER NOT NULL,
+    old_item JSONB,
+    new_item JSONB,
+    index_context JSONB NOT NULL,
+    ready_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- Serves the worker claim: WHERE worker_partition = $1 AND ready_at <= NOW()
+-- ORDER BY id. The leading partition column also keeps each worker's scan
+-- confined to its own rows.
+CREATE INDEX IF NOT EXISTS idx_gsi_pending_claim
+    ON gsi_pending (worker_partition, ready_at, id);
--- a/crates/storage-postgres/src/data/delete_item.rs
+++ b/crates/storage-postgres/src/data/delete_item.rs
@@ -9,7 +9,7 @@ use extenddb_storage::StreamCapture;
 use extenddb_storage::error::StorageError;
 use extenddb_storage::util::{SortKeyValue, parse_sk, pk_to_text, sk_column, sk_info};
 
-use super::index::{enqueue_async_indexes, fetch_indexes_for_table, pk_hash, sync_indexes};
+use super::index::{enqueue_async_indexes, fetch_indexes_for_table, sync_indexes};
 use super::query::check_condition;
 use super::tx_helpers::write_stream_record_in_tx;
 use super::{data_table_name, json_to_item};
@@ -128,7 +128,6 @@ impl PostgresEngine {
                         .transpose()?;
                     sync_indexes(
                         &mut tx,
-                        &key_info.table_id,
                         &key_info.key_schema,
                         &key_info.attribute_definitions,
                         &indexes,
@@ -155,26 +154,26 @@ impl PostgresEngine {
                     )
                     .await?;
                 }
+                // Persist async GSI work inside the same transaction — one row
+                // per async index, each honoring its own propagation delay.
+                let async_enqueued = enqueue_async_indexes(
+                    &mut tx,
+                    key_info,
+                    &indexes,
+                    old_item_for_idx.as_ref(),
+                    None,
+                    sys_delay,
+                )
+                .await?;
+
                 tx.commit()
                     .await
                     .map_err(|e| StorageError::Internal(e.to_string()))?;
 
-                // Enqueue async GSI updates after commit (D-4).
-                if let Some(ref q) = self.gsi_queue {
-                    enqueue_async_indexes(
-                        q,
-                        pk_hash(pk_text.as_ref()),
-                        &key_info.account_id,
-                        &key_info.table_name,
-                        &key_info.table_id,
-                        &key_info.key_schema,
-                        &key_info.attribute_definitions,
-                        &indexes,
-                        old_item_for_idx.as_ref(),
-                        None,
-                        sys_delay,
-                    )
-                    .await;
+                if async_enqueued > 0
+                    && let Some(ref q) = self.gsi_queue
+                {
+                    q.notify_workers();
                 }
 
                 if return_old {
@@ -266,7 +265,6 @@ impl PostgresEngine {
                         .transpose()?;
                     sync_indexes(
                         &mut tx,
-                        &key_info.table_id,
                         &key_info.key_schema,
                         &key_info.attribute_definitions,
                         &indexes,
@@ -293,26 +291,26 @@ impl PostgresEngine {
                     )
                     .await?;
                 }
+                // Persist async GSI work inside the same transaction — one row
+                // per async index, each honoring its own propagation delay.
+                let async_enqueued = enqueue_async_indexes(
+                    &mut tx,
+                    key_info,
+                    &indexes,
+                    old_item_for_idx.as_ref(),
+                    None,
+                    sys_delay,
+                )
+                .await?;
+
                 tx.commit()
                     .await
                     .map_err(|e| StorageError::Internal(e.to_string()))?;
 
-                // Enqueue async GSI updates after commit (D-4).
-                if let Some(ref q) = self.gsi_queue {
-                    enqueue_async_indexes(
-                        q,
-                        pk_hash(pk_text.as_ref()),
-                        &key_info.account_id,
-                        &key_info.table_name,
-                        &key_info.table_id,
-                        &key_info.key_schema,
-                        &key_info.attribute_definitions,
-                        &indexes,
-                        old_item_for_idx.as_ref(),
-                        None,
-                        sys_delay,
-                    )
-                    .await;
+                if async_enqueued > 0
+                    && let Some(ref q) = self.gsi_queue
+                {
+                    q.notify_workers();
                 }
 
                 if return_old {
--- a/crates/storage-postgres/src/gsi_queue.rs
+++ b/crates/storage-postgres/src/gsi_queue.rs
@@ -1,38 +1,78 @@
 // Copyright 2026 ExtendDB contributors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Async GSI update queue (D-4, D-5, D-6).
+//! Persistent GSI update queue (D-4, D-5, D-6).
 //!
-//! Base table writes commit independently. For GSIs with a non-zero
-//! propagation delay, index updates are enqueued here and applied after a
-//! random delay, simulating real `DynamoDB` eventual consistency.
+//! Base table writes insert a row into `gsi_pending` within the same
+//! transaction as the item mutation. Each row is **self-describing**: it
+//! carries the base key schema, attribute definitions, and the target index
+//! definitions captured at enqueue time (`GsiApplyContext`), so workers apply
+//! updates with **zero catalog reads**. A GSI's key schema and projection are
+//! immutable after creation, so the snapshot can never be stale relative to
+//! the live index definition; if the index is dropped, the apply is skipped.
 //!
-//! Each queue partition is consumed by a single worker task, guaranteeing
-//! per-key FIFO ordering. Workers are event-driven via `Notify` and sleep
-//! when the queue is empty.
+//! Workers claim, apply, and delete each row inside a **single transaction**
+//! (`SELECT ... FOR UPDATE SKIP LOCKED` → index writes → `DELETE` → `COMMIT`).
+//! A crash or transient error before `COMMIT` rolls the whole unit back, so the
+//! pending row reappears and is reprocessed — nothing is lost once enqueued.
+//!
+//! Rows are partitioned by a stable hash of the base table key, and each worker
+//! owns one partition, so all updates to a given base item are applied in `id`
+//! order by a single worker (**per-key FIFO**) while distinct keys propagate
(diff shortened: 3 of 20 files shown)


crates/storage-postgres/src/gsi_queue.rs
```diff
@@ -24,223 +26,405 @@ use crate::data::{
...
+        String,
+        Option<serde_json::Value>,
+        Option<serde_json::Value>,
+    )> = sqlx::query_as(
+        "DELETE FROM gsi_pending \
+         WHERE id IN ( \
+             SELECT id FROM gsi_pending \
+             WHERE ready_at <= NOW() \
+             ORDER BY id \
+             LIMIT $1 \
```
jcshepherd: Is LIMIT applied before or after ORDER BY?

crates/storage-postgres/src/data/delete_item.rs
```diff
@@ -155,26 +156,23 @@ impl PostgresEngine {
...
-                    .map_err(|e| StorageError::Internal(e.to_string()))?;
-
-                // Enqueue async GSI updates after commit (D-4).
-                if let Some(ref q) = self.gsi_queue {
-                    enqueue_async_indexes(
-                        q,
-                        pk_hash(pk_text.as_ref()),
-                        &key_info.account_id,
-                        &key_info.table_name,
+                if has_async_indexes(&indexes, sys_delay) {
```
jcshepherd: This could probably be called once at the beginning (right after `indexes` is populated).

crates/storage-postgres/data_migrations/002_gsi_pending.sql
```diff
@@ -0,0 +1,16 @@
+-- Copyright 2026 ExtendDB contributors
+-- SPDX-License-Identifier: Apache-2.0
+-- Persistent queue for async GSI propagation.
+-- Inserted atomically within the base write transaction, consumed by
+-- background workers. Survives process crash/restart.
+
+CREATE TABLE IF NOT EXISTS gsi_pending (
```
jcshepherd, line 20: I don't recall where it's buried, but somewhere there is a catalog version identifier that should be updated when the metadata/system table schema is updated, so that extenddb migrate will know that there is a migration to perform. I believe it's in storage-postgres somewhere.

jcshepherd (commented):
Couple initial questions/comments. Probably the one I'm most concerned with is the evaluation order of LIMIT and ORDER BY. If I were a gambler, I'd wager results are LIMITed before ORDER BY, which may not given you the ordering guarantees you want.

crates/storage-postgres/src/gsi_queue.rs
jcshepherd: I did a bit of reading ... It looks like Postgres will sort the entire resultset first, and then apply the limit. So if "ready_at" is far enough in the past on a busy table, the sort could be expensive, though the index on ready_at _should_ help mitigate that. My guess is that BATCH_SIZE is small enough to discourage the planner from believing that a full table scan would be cheaper than an index scan.

amrith: After

amrith:
I'm uncomfortable with the caching of index information. In my steering rules, I have a hard veto on anything other than auth data in a cache. this PR violates that and that's something I want to think through very closely. There are also many minor (lesser) nits. 

Why are we still populating index_name? Per steering: "#[allow(dead_code)] is a review blocker unless accompanied by a TODO(phase-N) comment." The underscore prefix is the same thing — a quieter way to suppress the warning.

Shouldn't there be migration version tracking?

LeeroyHannigan:
> I'm uncomfortable with the caching of index information. In my steering rules, I have a hard veto on anything other than auth data in a cache. this PR violates that and that's something I want to think through very closely. There are also many minor (lesser) nits.
> 
> Why are we still populating index_name? Per steering: "#[allow(dead_code)] is a review blocker unless accompanied by a TODO(phase-N) comment." The underscore prefix is the same thing — a quieter way to suppress the warning.
> 
> Shouldn't there be migration version tracking?

Cache removed, there's no metadata cache, and the queue no longer holds a `catalog_pool` at all. Each `gsi_pending` row is now self-describing: it carries an `index_context` snapshot (base key schema, attribute definitions, and the target index definitions) captured at enqueue, so workers apply with zero catalog reads. 

That snapshot is the work item's payload, written transactionally with the base write, not a lookup-aside cache, so it doesn't fall under the veto. It's also safe by construction: a GSI's key schema and projection are immutable after creation, so the snapshot can't go stale, and it removes the 30s staleness window the cache had.

amrith:
If I understand the code correctly, I have three issues that I'd like fixed before we merge this.

1. `delete_table.rs` — orphaned `gsi_pending` rows:

When a table is deleted, its gsi_pending rows are left behind. Workers handle this gracefully via the savepoint skip on 42P01, but it creates unnecessary work (claim →  deserialize → attempt → fail → skip → delete). Consider adding DELETE FROM gsi_pending WHERE table_id = $1 in the data-pool transaction that drops the tables. Not blocking, but avoids log noise and wasted cycles, especially if many items were written just before deletion.

2. `enqueue_gsi_pending` — delay uses system default:                                          

delay_ms is always sys_delay (the system default). The old per-index effective_delay(idx, system_default) override is lost. Since GsiApplyContext bundles all async indexes into one row with one ready_at, any per-index propagation delay override is effectively ignored. 
                                              
3. Deterministic vs randomized delay:                                                          
 
The old code applied random_range(1..=delay_ms). The new code uses a fixed NOW() + delay_ms. This makes GSI propagation deterministic rather than simulating DynamoDB's eventual-consistency jitter.

LeeroyHannigan:
Pushed a revision addressing all feedback, and merged latest `main` (the branch was 19 commits behind) so the PR is conflict-free again.

**@amrith #2 (per-index delay) + #3 (jitter):** the root cause was bundling every async GSI into one `gsi_pending` row with a single `ready_at`, which can't represent differing per-index delays and dropped the jitter. Now **one row per async index**, each carrying its own `effective_delay`. Jitter is restored as a per-enqueue random value — bounded to `[delay/2, delay]` rather than the original `1..=delay`, so the configured propagation delay stays a meaningful lower bound (a literal `1..=delay` can fire in ~1 ms, making the delay almost meaningless and untestable). Crucially, `ready_at` is clamped `GREATEST(NOW()+jitter, MAX(ready_at) in partition)` so jitter can never reorder same-partition rows — per-key FIFO convergence is preserved (verified by the ordering test).

**@amrith #1 (orphaned rows):** `delete_table` (sync path) and the control-plane poller (deferred DELETING path) now `DELETE FROM gsi_pending WHERE table_id = $1` in the same transaction that drops the tables, so workers no longer claim→deserialize→skip orphans.

**@jcshepherd / @amrith (migration tracking):** the `schema_history` ledger existed, but `extenddb migrate` only ran *catalog* migrations and returned early on a version match — so `002_gsi_pending.sql` was never applied on an existing deployment (the table would be missing and async-GSI writes would fail). `migrate` now also runs the data-DB migrations (idempotent, independent of the catalog version) and reports/gates pending ones via a new `Bootstrapper::pending_data_migrations`. Verified end-to-end: refuses without `--yes`, applies `002` (adopting a pre-tracking `001`) with it, idempotent on re-run.

**@jcshepherd (LIMIT vs ORDER BY):** confirmed by @amrith ("After") — Postgres applies `ORDER BY` before `LIMIT`, and the `(worker_partition, ready_at, id)` index keeps the claim cheap. No change needed.

**@amrith (caching veto / dead `index_name`):** already resolved earlier on this branch via self-describing rows (no metadata cache; `catalog_pool` dropped). The per-index rows keep that — workers apply with zero catalog reads. The `#[allow(clippy::too_many_arguments)]` introduced during this work has also been removed (use a `&TableKeyInfo` param instead).

**Validation:** `cargo fmt --check`, `clippy -D warnings`, unit tests, CLI lifecycle + GSI async queue integration (incl. new tests for per-index delay, orphan cleanup, and a migrate-applies-pending-data-migration regression), and the item/GSI/multipart/transaction/table/conditional endpoint suites — all green.

crates/storage-postgres/src/gsi_queue.rs
```diff
@@ -1,246 +1,466 @@
...
+    };
+
+    let delay_interval = jitter_delay_ms(delay_ms) as f64 / 1000.0;
+    sqlx::query(
+        "INSERT INTO gsi_pending \
+         (table_id, worker_partition, old_item, new_item, index_context, ready_at) \
+         VALUES ($1, $2, $3, $4, $5, GREATEST( \
+             NOW() + make_interval(secs => $6), \
+             COALESCE( \
+                 (SELECT MAX(ready_at) FROM gsi_pending WHERE worker_partition = $2), \
```
jcshepherd, line 223: If MAX(ready_at) wins, doesn't that mean you'll now have multiple records in the same worker partition with the same ready_at, and therefore they will be applied in non-deterministic order (even for the same key)? Maybe consider nudging MAX(ready_at) up?

crates/storage-postgres/src/gsi_queue.rs
```diff
@@ -1,246 +1,466 @@
...
+/// A crash or error before the `COMMIT` rolls all of it back — the row stays in
+/// `gsi_pending` and is retried. Each worker owns one partition and processes it
+/// in `id` order, so updates to a given base item apply in order (per-key FIFO);
+/// distinct partitions run concurrently. Only rows whose `ready_at` has elapsed
+/// are eligible; the propagation delay is enforced by that timestamp.
+async fn process_batch(worker_id: u64, q: &GsiQueue) -> Result<usize, StorageError> {
+    let mut processed = 0usize;
+
+    for _ in 0..BATCH_SIZE {
+        let mut tx = q
```
jcshepherd, line 312: It might be worth looking into whether you could use savepoints here to enable the whole batch to be processed in a single transaction, but if you need to rollback (you hit a failure with one GSI update) you can rollback to your savepoint (e.g., your last-good update), rather than rollback the whole batch. Similar safety to what you have here but without the overhead of a transaction per row.

crates/storage-postgres/src/gsi_queue.rs
```diff
@@ -1,246 +1,466 @@
...
+    let context: GsiApplyContext =
+        serde_json::from_value(ctx_json).map_err(|e| StorageError::Internal(e.to_string()))?;
 
-    let mut tx = pool
-        .begin()
+    // One index per row. Guard the apply with a savepoint so a dropped-index
+    // race (42P01) can be recovered and the row still consumed; aborting the
+    // subtransaction otherwise poisons the outer transaction and blocks the
+    // subsequent DELETE.
+    sqlx::query("SAVEPOINT gsi_apply")
```
jcshepherd, line 388: Oh, I see you're using savepoints here. Does that preclude using them in the batch loop as well?

jcshepherd (approved):
A couple minor comments below, but nothing blocking.

crates/storage-postgres/src/gsi_queue.rs
LeeroyHannigan: Apply order within a partition is determined by ORDER BY id, not `ready_at`, the clamp only governs eligibility timing (when a row becomes visible to a worker), not tiebreaking. Since id is the monotonic enqueue sequence, rows with an identical `ready_at` still apply in strict FIFO, so per-key ordering holds. Happy to add a +1ms nudge to `MAX(ready_at)` as belt-and-suspenders, but it's redundant given the id ordering.

crates/storage-postgres/src/gsi_queue.rs
LeeroyHannigan: Each row is processed in its own transaction here: `begin → claim one row (ORDER BY id LIMIT 1 FOR UPDATE SKIP LOCKED) → apply → delete → commit`  so a failure rolls back only that row and it's independently retried, while `SKIP LOCKED` keeps the other workers non-blocking. Folding the whole batch into one transaction with a per-row savepoint would cut commit overhead, but it trades away that simple per-row crash isolation. At `BATCH_SIZE=100` and the current GSI volume the commits aren't a bottleneck, so I'd rather take it as a follow-up than restructure this path now.

crates/storage-postgres/src/gsi_queue.rs
LeeroyHannigan: The SAVEPOINT `gsi_apply` here is scoped to one specific purpose, not general per-index isolation: there's a single target index per row, and the savepoint lets us recover from a dropped-index race (42P01, the index table was deleted out from under the apply) so the row is still consumed instead of the aborted subtransaction poisoning the outer txn and blocking the DELETE. Any other error deliberately propagates and rolls the row's transaction back for retry. So it's a targeted recovery guard rather than the batch-isolation mechanism from the other thread.

PR #131, opened 2026-05-26
fix: reject redundant parentheses in KeyConditionExpression and FilterExpression

### LINKED ISSUE #130: [Bug] Redundant parentheses not rejected in KeyConditionExpression and mislabelled in FilterExpression
## Describe the bug

DynamoDB rejects redundant parentheses in any expression with `The expression has redundant parentheses;`. ExtendDB only does this for `ConditionExpression` (added in #71, which closed #64). The other two expression contexts are inconsistent:

- `KeyConditionExpression` does not reject redundant parentheses at all. `((pk = :v))` is accepted, and `((pk = :pk)) AND (sk = :sk)` fails with an unrelated parse error instead of the redundant-parentheses message.
- `FilterExpression` does reject them (it shares the condition parser) but reports the error as `Invalid ConditionExpression: ...` rather than `Invalid FilterExpression: ...`.

## To Reproduce

1. Start ExtendDB and create a table with a composite key (`pk` hash, `sk` range).
2. Run a Query with `KeyConditionExpression` of `((pk = :v))`. It is accepted instead of rejected.
3. Run a Scan with `FilterExpression` of `((a = :v))`. It is rejected, but the message reads `Invalid ConditionExpression: ...`.

## Expected behavior

Both contexts reject redundant parentheses, matching DynamoDB and the existing `ConditionExpression` behaviour:

- `Invalid KeyConditionExpression: The expression has redundant parentheses;`
- `Invalid FilterExpression: The expression has redundant parentheses;`

## Actual behavior

- `KeyConditionExpression`: `((pk = :v))` is accepted; `((pk = :pk)) AND (sk = :sk)` returns a generic parse error.
- `FilterExpression`: rejected, but labelled `Invalid ConditionExpression: ...`.

## Environment

- ExtendDB version: current `main`
- Deployment method: source
- Platform: n/a - parser-level validation, deterministic from source and reproducible on any platform.
- Client SDK/driver: any. The validation happens server-side, before SDK marshalling matters.

## Additional context

`ConditionExpression` already rejects these correctly since #71. This is about bringing the other two expression contexts to the same behaviour.

## Checklist

- [x] I can reproduce this on current `main`

hicksy:
## What

Extends redundant-parentheses rejection to `KeyConditionExpression` and `FilterExpression`, so all three expression contexts behave the same way.

- Redundant-paren detection moves into one shared token-level check. Each parser prefixes the error with its own expression type.
- `KeyConditionExpression` now rejects forms like `((pk = :v))`.
- `FilterExpression` errors are now labelled `Invalid FilterExpression:` instead of `Invalid ConditionExpression:`.
- `ConditionExpression` behaviour is unchanged. The detection it gained in #71 now lives in the shared check.

## Why

Closes #130.

DynamoDB rejects redundant parentheses in every expression with `The expression has redundant parentheses;`. #71 added that for `ConditionExpression` only. `KeyConditionExpression` accepted them, and `FilterExpression` rejected them under the wrong label. This brings the remaining two contexts in line.

## Testing done

- `cargo test --workspace` passes. Added unit tests covering rejection and acceptance of parenthesised forms across `ConditionExpression`, `FilterExpression` and `KeyConditionExpression`, including the existing condition and transaction condition-check paths.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.
- Verified against a local server with the AWS SDK: a Query with `((pk = :v))` and a Scan with `((a = :v))` are now rejected with the documented messages, and `PutItem`/`UpdateItem`/`DeleteItem` and transaction condition checks still reject `((a = :v))` with `Invalid ConditionExpression: The expression has redundant parentheses;`.

## Notes

Redundant-paren detection now runs as a single pass at parse entry rather than during the descent. For an expression that is both redundantly parenthesised and otherwise malformed, the redundant-parentheses error may now surface where a different syntax error would have surfaced previously. Both paths still reject with a `ValidationException`, and no behavioural contract pins the message for doubly-invalid input.

## Checklist

- [x] Clippy is clean (`cargo clippy --all-targets -- -D warnings`, matching CI)
- [x] I have updated documentation if behavior changed (no user-facing docs describe redundant-paren handling; this increases DynamoDB parity)

## Breaking changes

Expressions that were previously accepted but are invalid on DynamoDB are now rejected. `KeyConditionExpression` values with redundant parentheses such as `((pk = :v))` now return a `ValidationException` instead of being accepted, matching DynamoDB. What filter and condition expressions accept is unchanged; only the `FilterExpression` error label changes.

---

### DIFF
--- a/crates/core/src/expression/key_condition.rs
+++ b/crates/core/src/expression/key_condition.rs
@@ -258,6 +258,9 @@ pub enum SortKeyCondition {
 ///
 /// Returns `ValidationException` for syntax errors or unsupported constructs.
 pub fn parse_key_condition(tokens: &[Token]) -> Result<KeyCondition, DynamoDbError> {
+    parser_common::check_redundant_parens(tokens)
+        .map_err(|body| validation_err(&format!("Invalid KeyConditionExpression: {body}")))?;
+
     // Strip outer parentheses: "(pk = :pk AND sk > :sk)" → "pk = :pk AND sk > :sk"
     let tokens = if tokens.len() >= 2
         && tokens[0] == Token::LParen
@@ -650,6 +653,35 @@ mod tests {
     use super::*;
     use crate::expression::tokenize;
     use std::collections::HashMap;
+
+    #[test]
+    fn key_condition_redundant_parens_rejected_with_canonical_message() {
+        for expr in ["((pk = :v))", "((#pk = :pk)) AND (#sk = :sk)"] {
+            let tokens = tokenize(expr).unwrap();
+            let err = parse_key_condition(&tokens).unwrap_err();
+            assert!(
+                matches!(&err, DynamoDbError::ValidationException(msg)
+                    if msg == "Invalid KeyConditionExpression: The expression has redundant parentheses;"),
+                "expr {expr}: got {err:?}"
+            );
+        }
+    }
+
+    #[test]
+    fn key_condition_valid_parens_accepted() {
+        for expr in [
+            "(pk = :v)",
+            "(#pk = :pk) AND (#sk = :sk)",
+            "(#pk = :pk AND #sk = :sk)",
+            "(#pk = :pk AND (#sk = :sk))",
+        ] {
+            let tokens = tokenize(expr).unwrap();
+            assert!(
+                parse_key_condition(&tokens).is_ok(),
+                "expr {expr} should parse"
+            );
+        }
+    }
 
     #[test]
     fn resolve_pk_sk_swaps_when_reversed() {
--- a/crates/core/src/expression/parser.rs
+++ b/crates/core/src/expression/parser.rs
@@ -525,4 +525,22 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn redundant_parens_use_canonical_message() {
+        for expr in [
+            "((a = :v))",
+            "(((a = :v)))",
+            "((a = :v AND b = :v2))",
+            "((NOT (a = :v)))",
+        ] {
+            let tokens = tokenize(expr).unwrap();
+            let err = parse_condition(&tokens).unwrap_err();
+            assert!(
+                matches!(&err, DynamoDbError::ValidationException(msg)
+                    if msg == "Invalid ConditionExpression: The expression has redundant parentheses;"),
+                "expr {expr}: got {err:?}"
+            );
+        }
+    }
 }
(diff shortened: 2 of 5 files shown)


pdf-amzn:
Oops! The comment originally here was intended for PR 133

hicksy:
Thanks for taking a look! 

Before I action anything - this PR (#131) is the redundant-parentheses fix for #130, and I can't see where it touches `resolver.rs`, `validate_begins_with_operands` or `begins_with`. 

Those three points all read like the `begins_with` operand-type work in #133. 

Have I got that right, or is there something in the paren change you're pointing at that I've not picked up?

pdf-amzn:
oh, ugh, I think the scripting I've been using to queue these up for checking botched this. Either that or I just copy-pasted wrong, although I remember having to double check because I thought something funny was going on. Will recheck

pdf-amzn:
OK, I've given this a proper look. I'm concerned about the O(n^2) behavior of `check_redundant_parens`, since it scans forward independently for every "(". Can this be done with a one-pass or two-pass algorithm? Otherwise looks good.

FWIW, GenAI gave me this code, the `if i + 1 < close` looks strange, but maybe it works well enough to prevent out-of-bounds accesses there? I haven't taken the time to think that through properly.

```
pub fn check_redundant_parens(tokens: &[Token]) -> Result<(), String> {
    // First pass: compute matching ')' index for each '('.
    let mut match_of: Vec<usize> = vec![0; tokens.len()];
    let mut stack: Vec<usize> = Vec::new();
    for (i, token) in tokens.iter().enumerate() {
        match token {
            Token::LParen => stack.push(i),
            Token::RParen => {
                if let Some(open) = stack.pop() {
                    match_of[open] = i;
                }
            }
            _ => {}
        }
    }
    // Second pass: check for redundant parens.
    for (i, token) in tokens.iter().enumerate() {
        if *token == Token::LParen && match_of[i] > 0 {
            let close = match_of[i];
            if i + 1 < close
                && tokens[i + 1] == Token::LParen
                && match_of[i + 1] == close - 1
            {
                return Err("The expression has redundant parentheses;".to_owned());
            }
        }
    }
    Ok(())
}
```

hicksy:
Yep, that's better. 

Switched `check_redundant_parens` to two passes (7657f35): one stack pass mapping each `(` to its matching `)`, then a single scan for nested groups. 

The per-paren rescan is removed, behaviour and error messages are unchanged, and the existing paren tests still pass. 

Also added a short comment on the `i + 1 < close` line.

pdf-amzn (approved):

pdf-amzn (approved):

PR #133, opened 2026-05-26
fix: validate begins_with operand types before evaluation

### LINKED ISSUE #132: [Bug] begins_with operand-type check skipped on FilterExpression for empty or non-matching scans
## Describe the bug

DynamoDB rejects `begins_with(path, value)` up front when the value isn't a string or binary. ExtendDB only checks this during row evaluation (added in #72, closed #65), so on a Scan or Query that doesn't match any rows, the check never runs and the request succeeds.

When the filter does evaluate against a row, the check fires but reports `Invalid ConditionExpression: ...` instead of `Invalid FilterExpression: ...` - the same `FilterExpression` labelling issue fixed for redundant parentheses in #130 / #131.

PutItem/UpdateItem/DeleteItem always evaluate the expression, so `ConditionExpression` isn't affected.

## To Reproduce

1. Create an empty table with `pk` (S).
2. Scan with `FilterExpression` `begins_with(#a, :n)`, `#a` -> `pk`, `:n` -> `{N:"1"}`. Accepted, should be rejected.
3. Put an item with `pk` set and re-run. Rejected, but labelled `Invalid ConditionExpression: ...`.

## Expected behavior

Both rejected up front with `ValidationException`. Same message body as #72, with the `Invalid FilterExpression:` prefix.

## Environment

- ExtendDB version: current `main`
- Deployment method: source

## Checklist

- [x] I can reproduce this on current `main`

## Notes

Happy to tackle this, but the fix cuts across the evaluator and the handlers, so worth a steer on the approach before I open a PR.

LeeroyHannigan:
## What
  
Validates `begins_with()` operand types at parse time (before row evaluation) in Scan and Query `FilterExpression`. Previously, the type check only ran during per-row evaluation (#72), so empty result sets silently accepted invalid types. 

Also fixes the error label: `FilterExpression` errors now report `Invalid FilterExpression:` instead of `Invalid ConditionExpression:`.

## Why

Closes #132

DynamoDB rejects `begins_with(path, value)` upfront when `value` is not S or B. ExtendDB only checked during evaluation, so Scan/Query on empty tables or non-matching partitions returned success instead of `ValidationException`. The error label mismatch (`ConditionExpression` vs `FilterExpression`) was a secondary issue from the same root cause.

## Testing done

- 14 new unit tests in `crates/core/src/expression/resolver.rs` covering all 8 invalid types, 2 valid types, and nested expressions (AND/OR/NOT)
- 13 new integration tests in `tests/python/test_begins_with_type_check.py` covering empty table scan, query filter, condition expression, and error label correctness
- Verified exact error messages match real DynamoDB
- `cargo test --workspace` — all pass

## Checklist

ADR / RFC: n/a

## Breaking changes

Expressions that were previously accepted on empty tables are now rejected. `begins_with(path, :number_value)` in `FilterExpression` now returns `ValidationException` regardless of whether any rows match, matching DynamoDB behaviour.

---

### DIFF
--- a/crates/core/src/expression/mod.rs
+++ b/crates/core/src/expression/mod.rs
@@ -28,7 +28,7 @@ pub use projection::{apply_projection, parse_projection};
 pub use reserved_words::validate_no_reserved_words;
 pub use resolver::{
     ExpressionMaps, collect_key_condition_refs, resolve_element_name, resolve_name_ref,
-    resolve_path, validate_unused_attributes,
+    resolve_path, validate_begins_with_operands, validate_unused_attributes,
 };
 pub use tokenizer::{Token, tokenize, tokenize_for, tokenize_with_limit};
 pub use update_evaluator::apply_update;
--- a/crates/core/src/expression/resolver.rs
+++ b/crates/core/src/expression/resolver.rs
@@ -12,6 +12,7 @@ use std::collections::{BTreeMap, HashMap};
 use crate::error::DynamoDbError;
 use crate::types::AttributeValue;
 
+use super::ast::Expr;
 use super::ast::PathElement;
 
 /// Resolved expression attribute names and values.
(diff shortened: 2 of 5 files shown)


pdf-amzn (approved):
Review notes:

_ => "UNKNOWN" unreachable arm.
The type_code match has a _ => "UNKNOWN" fallback, but all AttributeValue variants are already covered (S, B handled by the outer check; N, Bool, Null, L, M, SS, NS, BS in the inner match). This arm is dead code. If a new variant is added to AttributeValue, this would silently produce "UNKNOWN" instead of a compile error.

Minor. Should be unreachable!() or removed, but not a functional issue today.

Unit tests are thorough but repetitive (14 tests with identical structure, varying only the AttributeValue variant). A parameterized approach would be more maintainable.

It'd be good to make those changes either as an update to this PR or in a follow-on PR. These are nonblocking; approving.

pdf-amzn:
I meant to paste more feedback here, and accidentally pasted it to the wrong PR.

**Overall:** Nice fix — tests pass against both ExtendDB and real DynamoDB, error messages match exactly. A few non-blocking suggestions:

---

**`crates/core/src/expression/resolver.rs` — `validate_begins_with_operands`**

**1. Incomplete AST traversal**

The function recurses into `And`, `Or`, and `Not` but returns `Ok(())` for all other `Expr` variants via the `_ => Ok(())` catch-all. If the AST has other compound nodes (e.g. parenthesized groups, `Between`, `In`) that can contain a nested `begins_with`, those wouldn't be validated. Worth confirming the parser always flattens compound expressions into `And`/`Or` nodes — if so, a comment noting that assumption would help future readers.

---

**2. No upfront validation in the ConditionExpression (write) path**

The integration test `test_condition_expression_rejects_invalid_type` passes, but only because PutItem always evaluates the condition against the existing/absent item. The scan/query paths got explicit upfront validation, but the write paths (PutItem, UpdateItem, DeleteItem) rely on the per-row evaluator catching it. If ExtendDB ever short-circuits condition evaluation on writes (e.g. for performance), this would silently regress. Consider adding the same `validate_begins_with_operands` call in the write paths for consistency, or at minimum a comment explaining why it's safe to omit.

---

**3. Unit test repetition (minor/stylistic)**

The 14 unit tests follow the same pattern with only the `AttributeValue` variant changing. A `#[test_case(...)]` parameterized approach (via the `test-case` crate) would reduce ~80 lines of boilerplate to ~20 while keeping the same coverage. Not blocking, just a suggestion for maintainability.

---

Regardless, this is a step forward; I'm willing to approve as is. You can land this as is and create a tracking issue for the remaining work, or amend this PR.

LeeroyHannigan:
Closed by #139

Issue #141, opened 2026-05-28
Add TiDB storage backend


ngaut:
## Summary

Track the proposal to add TiDB as a first-class ExtendDB storage backend.

## Motivation

ExtendDB currently ships with PostgreSQL storage. A TiDB backend would give operators a MySQL-compatible, horizontally scalable SQL storage option while preserving the DynamoDB-compatible API surface.

## Scope

The RFC should cover:

- Backend selection and configuration model
- Storage trait and backend contract implications
- TiDB schema, migration, and operational requirements
- TiDB-native behavior for secondary indexes, snapshot-consistent reads, TTL, and backup/restore
- Compatibility expectations and rollout plan

Related implementation prototype: #140

PR #140, opened 2026-05-28
Add TiDB storage backend

ngaut:
## Summary
- Add an optional TiDB storage backend and wire it through workspace, binary, config, bootstrap, serve, backup/restore, TTL, and tests.
- Make backend defaults and bootstrap options backend-owned instead of passing raw Postgres-shaped CLI arguments through shared code.
- Leverage TiDB-native capabilities: generated-column secondary indexes, snapshot-consistent reads, native TTL where stream REMOVE records are not required, and BR-backed backup/restore.
- Update docs, config samples, install scripts, and CLI lifecycle tests for generic `--storage-*` backend flags and TiDB guidance.

## Verification
- `cargo fmt --all -- --check`
- `git diff --check HEAD`
- `cargo test -j12 -p extenddb-storage-tidb`
- `cargo test -j12 -p extenddb --no-default-features --features tidb --all-targets`
- `cargo test -j12 -p extenddb --all-targets`
- `cargo test -j12 -p extenddb-storage-postgres`
- `cargo clippy -j12 -p extenddb --no-default-features --features tidb --all-targets -- -D warnings`
- `/tmp/extenddb-docs-venv/bin/python docs/build-docs.py`

## Known Existing Issue
- `cargo clippy -j12 -p extenddb --all-targets -- -D warnings` still fails on `crates/storage-postgres/src/data/update_item.rs:168` (`clippy::collapsible_else_if`). This is an unrelated pre-existing Postgres data-plane lint left untouched.

PR #142, opened 2026-05-28
RFC: TiDB storage backend

ngaut:
## What

Adds `docs/rfcs/0000-tidb-storage-backend.md`, a draft RFC for adding TiDB as a first-class optional ExtendDB storage backend.

## Why

The implementation prototype in #140 was marked deferred pending an RFC. This RFC tracks the design discussion requested by the maintainer and links to tracking issue #141.

Related: #140
Tracks: #141

## Testing done

- `git diff --check`
- ASCII-only RFC check with `perl -ne 'print "$ARGV:$.:$_" if /[^\\x00-\\x7F]/' docs/rfcs/0000-tidb-storage-backend.md`
- `/tmp/extenddb-docs-venv/bin/python docs/build-docs.py`

## Checklist

- [x] Code is formatted (`cargo fmt --check`) - documentation-only PR
- [x] This PR is the RFC for the proposed storage/backend and CLI/config surface changes

ADR / RFC: this PR

### DIFF
--- a/docs/rfcs/0000-tidb-storage-backend.md
+++ b/docs/rfcs/0000-tidb-storage-backend.md
@@ -1,0 +1,358 @@
+# RFC-NNNN: TiDB storage backend
+
+- Status: Draft
+- Author: @ngaut
+- Created: 2026-05-28
+- FCP ends: TBD
+- Tracking issue: #141
+
+## Summary
+
+Add TiDB as a first-class, optional ExtendDB storage backend. The backend uses
+TiDB's MySQL-compatible SQL surface for normal DynamoDB operations and delegates
+database-native responsibilities, such as secondary index maintenance,
+transactional snapshots, TTL cleanup, and physical backup/restore, to TiDB
+instead of reimplementing them in ExtendDB.
+
+## Motivation
+
+ExtendDB currently has one production storage backend. PostgreSQL is a good
+default for local development and small installations, but some operators need a
+horizontally scalable SQL store with MySQL protocol compatibility, distributed
+transactions, online DDL, native TTL, and native backup/restore. TiDB is a good
+fit for that operator profile while preserving ExtendDB's DynamoDB-compatible
+wire protocol.
+
+The target users are:
+
+- teams that already operate TiDB and want DynamoDB-compatible APIs on top of
+  their existing database platform;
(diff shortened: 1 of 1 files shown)


navareamol:
Thank you for submitting the RFC for TiDB storage backend. We're reviewing your submission and debating on certain feature sets in the RFC. We will circle back with detailed response by 7/9.

LeeroyHannigan:
Thanks for the thorough RFC, @ngaut this is well thought through and we're happy to accept TiDB as an optional backend along these lines. The backend-additive, backend-owned-implementation approach is the right call.

A few DynamoDB-parity points to fold in before/with implementation, split into what must match vs. what's fine to differ as long as it's documented:

### Must match DynamoDB (observable-behavior breaks otherwise)

- TTL + Streams: relying on native TiDB TTL means expired items won't emit `REMOVE` stream records, that's a breaking gap, not just a caveat. When streams are enabled, TTL deletes must go through an ExtendDB path that emits DynamoDB-compatible `REMOVE` records.

- Key ordering/equality: DynamoDB compares String keys by case-sensitive UTF-8 byte order and Numbers as exact decimals (up to 38 significant digits). The generated index-key columns must use a binary collation (not MySQL's default case-insensitive collation) and a decimal type for numeric keys, or Query ordering and key equality will diverge.

- Number fidelity: items must round-trip numbers as exact decimals, JSON storage must not coerce them to floats.

- Conditional/transactional writes: must stay atomic (no write-skew) and keep surfacing `ConditionalCheckFailedException` / `TransactionCanceledException` as today.

### Fine to differ, just call it out

- GSI consistency: maintaining indexes transactionally makes GSI reads strongly consistent instead of eventually consistent. That's acceptable (stricter than DynamoDB) please document it as an intentional backend difference, and still reject ConsistentRead=true on a GSI so the API surface matches.

- Point-in-time window: snapshot-backed reads are bounded by TiDB's GC lifetime, so the effective PITR/export window is shorter than DynamoDB's 35 days. Fine to ship with a documented limit.

Also please keep sparse-index/index-key type rules and projection (KEYS_ONLY/INCLUDE/ALL) enforced at read time, since items are stored once.

Overall: accepted in principle, let's nail down the TTL-stream path and key-encoding details, then move to implementation behind the opt-in feature. Thanks again for the careful write-up!

PR #146, opened 2026-05-31
fix: use compound condition for index ExclusiveStartKey pagination

### LINKED ISSUE #145: [Bug] LSI/GSI pagination with ExclusiveStartKey fails when index sort key has duplicate values
## Describe the bug

When querying a Local Secondary Index (LSI) with `Limit` and paginating via `ExclusiveStartKey`, the second page returns 0 items if multiple items share the same index sort key value. The SQL WHERE clause for ExclusiveStartKey consumption only compares the index sort key (`AND sk_col > $last_sk_value`) without using the base table sort key as a tie-breaker.

This was partially addressed in PR #66 which fixed `LastEvaluatedKey` **output** to include base table keys, but the **input** consumption of `ExclusiveStartKey` in `query_scan.rs` still ignores the base table sort key.

## To Reproduce

1. Create a table with an LSI:
```bash
aws dynamodb create-table \
  --table-name TestTable \
  --attribute-definitions \
    AttributeName=pk,AttributeType=S \
    AttributeName=sk,AttributeType=S \
    AttributeName=taskType,AttributeType=S \
  --key-schema AttributeName=pk,KeyType=HASH AttributeName=sk,KeyType=RANGE \
  --local-secondary-indexes '[{"IndexName":"TaskTypeLSI","KeySchema":[{"AttributeName":"pk","KeyType":"HASH"},{"AttributeName":"taskType","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]' \
  --billing-mode PAY_PER_REQUEST \
  --endpoint-url https://localhost:8000
```

2. Insert 5 items with same partition key and same LSI sort key (`taskType=EXPORT`), differing only in base table sort key (`sk=task1..task5`):
```bash
for i in 1 2 3 4 5; do
  aws dynamodb put-item --table-name TestTable \
    --item "{\"pk\":{\"S\":\"user1\"},\"sk\":{\"S\":\"task$i\"},\"taskType\":{\"S\":\"EXPORT\"}}" \
    --endpoint-url https://localhost:8000
done
```

3. Query the LSI without pagination — returns all 5 items:
```bash
aws dynamodb query --table-name TestTable --index-name TaskTypeLSI \
  --key-condition-expression "pk = :pk" \
  --expression-attribute-values '{":pk":{"S":"user1"}}' \
  --endpoint-url https://localhost:8000
# Returns Count: 5
```

4. Query with `--limit 2` — returns 2 items + `LastEvaluatedKey`:
```bash
aws dynamodb query --table-name TestTable --index-name TaskTypeLSI \
  --key-condition-expression "pk = :pk" \
  --expression-attribute-values '{":pk":{"S":"user1"}}' \
  --limit 2 \
  --endpoint-url https://localhost:8000
# Returns Count: 2, LastEvaluatedKey: {"pk":{"S":"user1"},"sk":{"S":"task2"},"taskType":{"S":"EXPORT"}}
```

5. Query page 2 using the `LastEvaluatedKey` as `--exclusive-start-key`:
```bash
aws dynamodb query --table-name TestTable --index-name TaskTypeLSI \
  --key-condition-expression "pk = :pk" \
  --expression-attribute-values '{":pk":{"S":"user1"}}' \
  --exclusive-start-key '{"pk":{"S":"user1"},"sk":{"S":"task2"},"taskType":{"S":"EXPORT"}}' \
  --limit 2 \
  --endpoint-url https://localhost:8000
# Returns Count: 0 — BUG
```

## Expected behavior

Page 2 should return the next 2 items (task3, task4). DynamoDB uses the base table primary key as a tie-breaker when index sort keys are equal, applying a compound condition:

```
(index_sort_key > last_index_sort_key)
OR (index_sort_key = last_index_sort_key AND base_table_sort_key > last_base_table_sort_key)
```

This is documented DynamoDB behavior — any LSI on a low-cardinality attribute (status, type, category) will have duplicate sort key values and must paginate correctly through them.

## Actual behavior

Page 2 returns 0 items. The generated SQL uses only `AND index_sk > $last_index_sk`, which matches nothing when all items share the same index sort key value.

Additionally, `ScanIndexForward=false` does not sub-sort by base table sort key when index sort keys are equal.

## Environment

- ExtendDB version: 0.1.0 (commit e50af5b, includes PR #66)
- Operating system: macOS
- Rust version: 1.94 (built from source)
- Client SDK/driver and version: AWS CLI v2 / AWS SDK for Java 1.x
- Deployment method: source

## Logs / stack trace

No error is thrown — the query silently returns an empty result set instead of the expected items.

## Additional context

**Affected code:** `crates/storage-postgres/src/data/query_scan.rs` lines 117-128

**Relationship to PR #66:** That PR fixed the *output* side (LastEvaluatedKey now correctly includes base table keys for index queries). This bug is about the *input* side — when consuming `ExclusiveStartKey`, the base table sort key is not used in the WHERE clause.

**Suggested fix:** When building the WHERE clause for ExclusiveStartKey on index queries, use a compound condition:
```sql
-- Forward scan (ScanIndexForward=true):
AND (index_sk > $last_index_sk
     OR (index_sk = $last_index_sk AND base_table_sk > $last_base_table_sk))

-- Reverse scan (ScanIndexForward=false):
AND (index_sk < $last_index_sk
     OR (index_sk = $last_index_sk AND base_table_sk < $last_base_table_sk))
```

**Impact:** Any paginated query on an LSI/GSI where multiple items share the same index sort key will return incomplete results. This is a very common DynamoDB access pattern (e.g., querying by status, type, or category).

## Checklist

- [x] I can reproduce this with the latest release

whoami3014:
## What

Fix index (LSI/GSI) pagination when items share the same index sort key value. Extends the fix to cover all index types, scan operations, and addresses all review feedback.

1. **Query pagination compound condition** — use base table key as tie-breaker when index sort keys are equal
2. **Scan pagination for indexes** — same compound condition for index Scan operations
3. **Hash-only GSI pagination** — use `base_pk` for positioning when index has no sort key
4. **GSI on hash-only base table** — use `base_pk` as tie-breaker when base table has no range key
5. **LSI vs GSI ordering** — LSI base SK follows ScanIndexForward; GSI base SK always ASC
6. **Cache base table key info** — `RwLock<HashMap>` cache on `PostgresEngine`, bounded to 500 entries, avoids per-request catalog queries
7. **`PaginationBinds` enum** — bind values are built in the same code branch that generates SQL placeholders, consumed via `match` in `execute_query_sql`, preventing bind-order divergence
8. **`build_pagination_where()` helper** — self-contained SQL generation function, eliminates `let _ = param_idx`

## Why

Closes #145

PR #66 fixed the *output* side (`LastEvaluatedKey` includes base table keys). This PR fixes the *input* side — consuming `ExclusiveStartKey` correctly for pagination.

## Testing done

- `cargo check --workspace` — clean
- `cargo fmt --check` — clean
- `cargo test --workspace --lib` — 489 passed
- Python integration tests — **55 passed**, 0 failed
- Covers: LSI pagination (forward/reverse), hash-only GSI, GSI on hash-only base table, scan pagination, repeated pagination (cache path)

## Checklist

---

### DIFF
--- a/crates/storage-postgres/src/data/query_scan.rs
+++ b/crates/storage-postgres/src/data/query_scan.rs
@@ -46,6 +46,10 @@ fn build_pagination_where(
             } else {
                 ""
             };
+            // LSI: base SK follows ScanIndexForward because items share the
+            // same partition and the base SK is part of the composite sort order.
+            // GSI: base SK is always ">" (ascending) because it's only a
+            // uniqueness tie-breaker, not a user-visible sort dimension.
             let base_cmp = if is_lsi { cmp } else { ">" };
             format!(
                 " AND ({sk_col}{collate} {cmp} ${p1} OR \
@@ -297,7 +301,7 @@ impl PostgresEngine {
                 let base_pk_attr = &key_info.base_key_schema[0].attribute_name;
                 let base_pk = start_key
                     .get(base_pk_attr.as_str())
-                    .map(|v| pk_to_text(v))
+                    .map(pk_to_text)
                     .transpose()?
                     .map(|c| c.into_owned());
                 match (base_pk, &base_sk_info) {
--- a/crates/storage-postgres/src/data/query_scan.rs
+++ b/crates/storage-postgres/src/data/query_scan.rs
@@ -4,17 +4,89 @@
 //! `query` and `scan` implementations for the `PostgreSQL` backend.
 
 use extenddb_core::expression::{ExpressionMaps, KeyCondition};
-use extenddb_core::types::{Item, ScalarAttributeType, TableKeyInfo};
+use extenddb_core::types::{Item, KeySchemaElement, ScalarAttributeType, TableKeyInfo};
 use extenddb_storage::error::StorageError;
 use extenddb_storage::util::{
-    encode_netstring_composite, pk_to_text, sk_column, sk_column_n, sk_info,
+    encode_netstring_composite, parse_sk, pk_to_text, sk_column, sk_column_n, sk_info,
 };
 
 use super::query::{
-    build_key, build_sk_sql, execute_query_sql, execute_scan_sql, resolve_expr_to_av,
+    PaginationBinds, build_key, build_sk_sql, execute_query_sql, execute_scan_sql,
+    resolve_expr_to_av,
 };
 use super::{all_sort_key_info, data_table_name, index_table_name, json_to_item};
 use crate::PostgresEngine;
+
+/// Build the WHERE clause fragment for ExclusiveStartKey pagination.
+///
+/// Self-contained: takes `param_idx` (the next available placeholder number),
+/// returns a complete SQL fragment. No mutable state leaks out.
+fn build_pagination_where(
+    param_idx: u32,
+    sk_info_val: Option<(&str, ScalarAttributeType)>,
+    base_sk_info: &Option<(String, ScalarAttributeType)>,
+    is_index: bool,
+    is_lsi: bool,
+    forward: bool,
+) -> String {
+    if let Some((_, sk_type)) = sk_info_val {
+        let sk_col = sk_column(sk_type);
+        let collate = if sk_type == ScalarAttributeType::S {
+            " COLLATE \"C\""
+        } else {
+            ""
+        };
+        let cmp = if forward { ">" } else { "<" };
+
+        if let Some((_, base_sk_type)) = base_sk_info {
+            // Index with base SK tie-breaker
+            let base_col = format!("base_{}", sk_column(*base_sk_type));
+            let base_collate = if *base_sk_type == ScalarAttributeType::S {
+                " COLLATE \"C\""
+            } else {
+                ""
+            };
+            let base_cmp = if is_lsi { cmp } else { ">" };
+            format!(
+                " AND ({sk_col}{collate} {cmp} ${p1} OR \
+                 ({sk_col}{collate} = ${p1} AND {base_col}{base_collate} {base_cmp} ${p2}))",
+                p1 = param_idx,
+                p2 = param_idx + 1
+            )
+        } else if is_index {
+            // Index with no base SK — use base_pk as tie-breaker
+            format!(
+                " AND ({sk_col}{collate} {cmp} ${p1} OR \
+                 ({sk_col}{collate} = ${p1} AND base_pk COLLATE \"C\" > ${p2}))",
+                p1 = param_idx,
+                p2 = param_idx + 1
+            )
+        } else {
+            // Base table — simple comparison
+            format!(" AND {sk_col}{collate} {cmp} ${param_idx}")
+        }
+    } else if is_index {
+        // Hash-only index — paginate using base table PK
+        if let Some((_, base_sk_type)) = base_sk_info {
+            let base_sk_col = format!("base_{}", sk_column(*base_sk_type));
+            let base_sk_collate = if *base_sk_type == ScalarAttributeType::S {
+                " COLLATE \"C\""
+            } else {
+                ""
+            };
+            format!(
+                " AND (base_pk COLLATE \"C\" > ${p1} OR \
+                 (base_pk = ${p1} AND {base_sk_col}{base_sk_collate} > ${p2}))",
+                p1 = param_idx,
+                p2 = param_idx + 1
+            )
+        } else {
+            format!(" AND base_pk COLLATE \"C\" > ${param_idx}")
+        }
+    } else {
+        String::new()
+    }
+}
 
 impl PostgresEngine {
     /// Implementation of `DataEngine::query`.
@@ -31,13 +103,14 @@ impl PostgresEngine {
     ) -> Result<(Vec<Item>, Option<Item>), StorageError> {
         use std::fmt::Write;
 
-        let ddb_table = if let Some(idx_name) = index_name {
+        let (ddb_table, is_lsi) = if let Some(idx_name) = index_name {
             let idx_info = self
                 .fetch_index_info_by_table_id(&key_info.table_id, idx_name)
                 .await?;
-            index_table_name(&idx_info.index_id)
+            let lsi = idx_info.index_type == extenddb_core::types::IndexType::Lsi;
+            (index_table_name(&idx_info.index_id), lsi)
         } else {
-            data_table_name(&key_info.table_id)
+            (data_table_name(&key_info.table_id), false)
         };
 
         // Resolve partition key value(s) — for multi-part keys, encode
@@ -113,22 +186,30 @@ impl PostgresEngine {
             }
         }
 
-        // Pagination: exclusive start key
-        if let (Some(_), Some((_, sk_type))) = (exclusive_start_key, sk_info_val) {
-            let sk_col = sk_column(sk_type);
-            let collate = if sk_type == ScalarAttributeType::S {
-                " COLLATE \"C\""
-            } else {
-                ""
-            };
-            if forward {
-                let _ = write!(sql, " AND {sk_col}{collate} > ${param_idx}");
-            } else {
-                let _ = write!(sql, " AND {sk_col}{collate} < ${param_idx}");
-            }
-        } else if exclusive_start_key.is_some() && sk_info_val.is_none() {
-            // PK-only table with start key — no more items for this PK
+        // For index queries, fetch the base table's sort key info. This is needed
+        // for both ORDER BY (sub-sort by base SK when index SKs are equal) and
+        // pagination (compound ExclusiveStartKey condition).
+        let base_sk_info: Option<(String, ScalarAttributeType)> = if index_name.is_some() {
+            self.fetch_base_table_sk_info(&key_info.table_id).await?
+        } else {
+            None
+        };
+
+        if exclusive_start_key.is_some() && sk_info_val.is_none() && index_name.is_none() {
+            // PK-only base table with start key — no more items for this PK
             return Ok((Vec::new(), None));
+        }
+
+        if exclusive_start_key.is_some() {
+            let pagination_sql = build_pagination_where(
+                param_idx,
+                sk_info_val,
+                &base_sk_info,
+                index_name.is_some(),
+                is_lsi,
+                forward,
+            );
+            sql.push_str(&pagination_sql);
         }
 
         // ORDER BY — use COLLATE "C" for string sort keys to match DynamoDB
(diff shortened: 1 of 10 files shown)


crates/storage-postgres/src/data/query_scan.rs
```diff
@@ -293,4 +338,32 @@ impl PostgresEngine {
...
+    /// Fetch the base table's sort key attribute name and type for a given `table_id`.
+    ///
+    /// Used during index query pagination to determine the `base_sk_*` column
+    /// for the compound tie-breaker condition.
+    async fn fetch_base_table_sk_info(
+        &self,
+        table_id: &str,
+    ) -> Result<Option<(String, ScalarAttributeType)>, StorageError> {
+        let row: Option<(serde_json::Value, serde_json::Value)> = sqlx::query_as(
+            "SELECT key_schema, attribute_definitions FROM tables WHERE table_id = $1",
```
LeeroyHannigan: A catalog query per index request is expensive under load. Could this be cached on the PostgresEngine (keyed by table_id, same as the index metadata cache), or could the engine pass the original base key_schema as an additional field on TableKeyInfo? Either would avoid the per-request DB hit.

crates/storage-postgres/src/data/query_scan.rs
```diff
@@ -113,23 +113,52 @@ impl PostgresEngine {
...
-                let _ = write!(sql, " AND {sk_col}{collate} < ${param_idx}");
+                // Base table query: simple comparison on sort key.
+                let _ = write!(sql, " AND {sk_col}{collate} {cmp} ${param_idx}");
+                param_idx += 1;
             }
         } else if exclusive_start_key.is_some() && sk_info_val.is_none() {
             // PK-only table with start key — no more items for this PK
             return Ok((Vec::new(), None));
         }
+        let _ = param_idx; // suppress unused warning
```
LeeroyHannigan: This suppression at the end is a smell. The bind ordering between query_scan.rs (which writes $N placeholders) and query.rs (which calls bind_sk_value) is implicit, if anyone adds a bind between these two blocks in the future, the indices go out of sync silently at runtime. Consider either asserting the expected param count or restructuring so the SQL builder and the binder can't diverge.

tests/rust/src/lsi.rs
LeeroyHannigan, line 1: All tests exercise LSIs. GSIs have a different failure mode: a hash-only base table (no sort key) with a composite GSI would mean there's no base_sk_* column to use as a tiebreaker. Does fetch_base_table_sk_info return None in that case? If so, the compound condition isn't generated and pagination still breaks for GSIs on hash-only tables. Can you add a GSI test case, ideally one where the base table has no range key?

LeeroyHannigan:
Thank you for taking the time to create this PR and contribute to ExtendDB. I left some feedback on the code, and also have a few further points:

This fixes Query pagination but Scan pagination with the same tied-sort-key problem is unaddressed. The scan path in the same file builds its WHERE clause similarly and needs the same compound condition. Can you extend the fix to cover scans as well?

Be sure to rebase on main to resolve existing conflicts.

Thanks, 
Lee

tests/rust/src/lsi.rs
whoami3014: Thanks for flagging this. Was able to reproduce the issue: fetch_base_table_sk_info returns None for hash-only base tables, and the code was falling through to the simple sk > $N comparison which silently breaks pagination (page 2 returns 0 items).

Fixed with an else if is_index branch in build_pagination_where() that uses base_pk as the tie-breaker when there's no base SK column. Added TestGSIOnHashOnlyBaseTable covering this scenario: base table is hash-only, GSI has a composite key (hash + range), 7 items share the same GSI sort key, paginated with Limit. Also added a scan test for the same case.

crates/storage-postgres/src/data/query_scan.rs
whoami3014: Agreed. Two changes to address it:

1. Introduced a `PaginationBinds` enum. The extra bind values (base_pk, base_sk) are now built in `query_scan.rs` in the same branch that generates the SQL placeholders, then passed to `execute_query_sql` which consumes them via a `match`. The bind order is determined by the enum variant, not by implicit convention between files.

2. Extracted `build_pagination_where()` as a pure function that takes `param_idx` as input and returns a SQL fragment. `param_idx` is consumed as a function argument so there's no leftover unused variable, no `let _ = param_idx` needed.

I looked at doing the full structural fix (a builder that couples every `$N` with its value at generation time) but it touches the entire query/scan SQL construction since every conditional branch that emits a placeholder would need to simultaneously capture the value. `SortKeyValue` doesn't derive `Clone` either so the builder's value storage gets messy without changes to the `storage` crate. Would bloat this bug fix significantly. Happy to take it as a followup that covers all bind paths (positions 1-3 have the same implicit pattern from the original code).

crates/storage-postgres/src/data/query_scan.rs
whoami3014: Addressed. Added a `RwLock<HashMap<String, BaseKeyInfo>>` cache on `PostgresEngine` keyed by `table_id`. First query for a table hits the catalog, subsequent ones serve from cache. Bounded to 500 entries (clears on overflow). No TTL since key schemas are immutable for a table's lifetime and `table_id` is never reused. Simpler than `SwrCache` intentionally since there's nothing to revalidate.

LeeroyHannigan:
Thanks for the update. The pagination logic itself is correct, but the newly added `base_key_cache` has some issues worth discussing before we merge:
  
1. Eviction policy - thundering herd

When the cache hits 500 entries, cache.clear() nukes everything at once. The next N index queries all miss simultaneously and hit the catalog DB in parallel to repopulate. A per-entry TTL or LRU eviction would avoid this, but honestly I don't think we need a cache at all (see point 3).

2. Multi-instance staleness

Multiple ExtendDB instances sharing the same PostgreSQL won't see each other's table deletions, a deleted table's entry stays cached indefinitely until the 500-entry eviction fires. In practice this is low-risk since table_id is never reused and the cached data (key schema) is immutable, but it's still unbounded memory growth from zombie entries that will never be read again.

3. The cache is unnecessary, we can avoid the catalog query entirely

The base table's key schema is already available in the engine layer when it constructs `query_key_info`. Right now the engine replaces `key_info.key_schema` with the index's key schema and passes that to storage, discarding the original. If we instead pass the base key schema alongside (e.g., as an additional field on `TableKeyInfo` or a separate parameter), the storage layer doesn't need to query the catalog or cache anything.

This would eliminate the per-request DB hit, the cache, and all the complexity around eviction and staleness in one go. 

Happy to help with a sketch if that approach works for you.

whoami3014:
Thanks for the thorough review, all three points are well taken.

**Agreed the cache should go.** Your third point (pass the base key schema from the engine layer) is the right fix. The engine already holds the base table's `key_info` (with original `key_schema` + `attribute_definitions`) before it constructs the index-swapped `query_key_info`/`scan_key_info`. It just doesn't pass it downstream today.

**Thinking of adding `base_key_schema` as a field on `TableKeyInfo`:**

```rust
pub struct TableKeyInfo {
    pub table_name: String,
    pub account_id: String,
    pub table_id: String,
    pub key_schema: Vec<KeySchemaElement>,        // "active" schema (index or base)
    pub base_key_schema: Vec<KeySchemaElement>,   // always the base table's schema
    pub attribute_definitions: Vec<AttributeDefinition>,
    pub has_lsi: bool,
    pub stream_specification: Option<StreamSpecification>,
}
```

This follows the same pattern as `stream_specification`: info the storage layer needs that would otherwise require an extra lookup, carried on the struct that's already passed to every operation.

- For base table operations: `base_key_schema == key_schema` (both are the table's own schema). No special-casing needed, it's always populated.
- For index operations: `key_schema` has the index's schema, `base_key_schema` retains the original table schema.
- No trait signature changes needed since `query()` and `scan()` already receive `&TableKeyInfo`, so the base key info comes along for free.

In the engine, the construction is natural:

```rust
let query_key_info = TableKeyInfo {
    key_schema: idx.key_schema.clone(),
    base_key_schema: key_info.key_schema.clone(),
    ..key_info.clone()
};
```

In `storage-postgres`, derive `base_pk_attr_name` and `base_sk_info` from `key_info.base_key_schema` + `key_info.attribute_definitions` using the existing `sk_info()` helper, same pattern used everywhere else in the crate.

**This would let us remove:**
- `get_base_key_info()`, `fetch_base_table_sk_info()`, `resolve_base_pk_from_start_key()`, `resolve_base_pk_attr_name()`
- `BaseKeyInfo` struct, `base_key_cache` field, `BASE_KEY_CACHE_MAX_ENTRIES` constant
- The catalog query (`SELECT key_schema, attribute_definitions FROM tables WHERE table_id = $1`)

This eliminates the per-request catalog query, the cache, and all eviction/staleness concerns. The pagination SQL logic and `PaginationBinds` enum stay as-is, only the source of the base key info changes (struct field vs cache lookup).

Does this align with what you had in mind? Open to suggestions on placement if you'd prefer it structured differently.

LeeroyHannigan:
> Thanks for the thorough review, all three points are well taken.
> 
> **Agreed the cache should go.** Your third point (pass the base key schema from the engine layer) is the right fix. The engine already holds the base table's `key_info` (with original `key_schema` + `attribute_definitions`) before it constructs the index-swapped `query_key_info`/`scan_key_info`. It just doesn't pass it downstream today.
> 
> **Thinking of adding `base_key_schema` as a field on `TableKeyInfo`:**
> 
> ```rust
> pub struct TableKeyInfo {
>     pub table_name: String,
>     pub account_id: String,
>     pub table_id: String,
>     pub key_schema: Vec<KeySchemaElement>,        // "active" schema (index or base)
>     pub base_key_schema: Vec<KeySchemaElement>,   // always the base table's schema
>     pub attribute_definitions: Vec<AttributeDefinition>,
>     pub has_lsi: bool,
>     pub stream_specification: Option<StreamSpecification>,
> }
> ```
> 
> This follows the same pattern as `stream_specification`: info the storage layer needs that would otherwise require an extra lookup, carried on the struct that's already passed to every operation.
> 
>     * For base table operations: `base_key_schema == key_schema` (both are the table's own schema). No special-casing needed, it's always populated.
> 
>     * For index operations: `key_schema` has the index's schema, `base_key_schema` retains the original table schema.
> 
>     * No trait signature changes needed since `query()` and `scan()` already receive `&TableKeyInfo`, so the base key info comes along for free.
> 
> 
> In the engine, the construction is natural:
> 
> ```rust
> let query_key_info = TableKeyInfo {
>     key_schema: idx.key_schema.clone(),
>     base_key_schema: key_info.key_schema.clone(),
>     ..key_info.clone()
> };
> ```
> 
> In `storage-postgres`, derive `base_pk_attr_name` and `base_sk_info` from `key_info.base_key_schema` + `key_info.attribute_definitions` using the existing `sk_info()` helper, same pattern used everywhere else in the crate.
> 
> **This would let us remove:**
> 
>     * `get_base_key_info()`, `fetch_base_table_sk_info()`, `resolve_base_pk_from_start_key()`, `resolve_base_pk_attr_name()`
> 
>     * `BaseKeyInfo` struct, `base_key_cache` field, `BASE_KEY_CACHE_MAX_ENTRIES` constant
> 
>     * The catalog query (`SELECT key_schema, attribute_definitions FROM tables WHERE table_id = $1`)
> 
> 
> This eliminates the per-request catalog query, the cache, and all eviction/staleness concerns. The pagination SQL logic and `PaginationBinds` enum stay as-is, only the source of the base key info changes (struct field vs cache lookup).
> 
> Does this align with what you had in mind? Open to suggestions on placement if you'd prefer it structured differently.

Much better.

whoami3014:
@LeeroyHannigan Pushed revision addressing the cache feedback (0129a73).

**What changed:**
- Added `base_key_schema: Vec<KeySchemaElement>` field on `TableKeyInfo`. Always populated: equals `key_schema` for base table operations, retains the original table schema for index operations.
- Engine sets it when constructing index query/scan key info.
- Storage-postgres derives base PK/SK info directly from the field using the existing `sk_info()` helper. No catalog query, no cache.

**Removed:**
- `get_base_key_info()`, `fetch_base_table_sk_info()`, `resolve_base_pk_from_start_key()`, `resolve_base_pk_attr_name()`
- `BaseKeyInfo` struct, `base_key_cache` field, `BASE_KEY_CACHE_MAX_ENTRIES`
- The catalog query (`SELECT key_schema, attribute_definitions FROM tables WHERE table_id = $1`)

**Tests added:**
- 2 unit tests verifying `base_key_schema` is set correctly for base table and index operations
- 3 integration tests verifying pagination works end-to-end for base table query, GSI query, and GSI scan (all using the new field path)

All 58 query/scan integration tests pass, 507 unit tests pass, formatting and clippy clean. Already rebased on main, no conflicts.

Happy to squash into a single commit if preferred.

crates/storage-postgres/src/data/query_scan.rs
```diff
@@ -7,15 +7,87 @@ use extenddb_core::expression::{ExpressionMaps, KeyCondition};
...
+
+        if let Some((_, base_sk_type)) = base_sk_info {
+            // Index with base SK tie-breaker
+            let base_col = format!("base_{}", sk_column(*base_sk_type));
+            let base_collate = if *base_sk_type == ScalarAttributeType::S {
+                " COLLATE \"C\""
+            } else {
+                ""
+            };
+            let base_cmp = if is_lsi { cmp } else { ">" };
```
LeeroyHannigan, line 53: nit: Can you comment on this so others coming across the code later can understand its logic easily.

LeeroyHannigan: Also, there's no reverse-direction GSI test with tied sort keys. If this line were accidentally cmp instead of ">", reverse GSI pagination would break silently. Could you add a GSI equivalent of lsi_pagination_reverse_order?

LeeroyHannigan:
Looks good, left 2 nits, and you have clippy failures. But will be good to merge then.

crates/storage-postgres/src/data/query_scan.rs
whoami3014: Addressed both comments.

LeeroyHannigan (approved):
Thank you for your PR and you prompt addressing of the feedback.

PR #155, opened 2026-06-02
fix: store and return TableClass, SSESpecification, OnDemandThroughput

LeeroyHannigan:
## What

Store and return `TableClass`, `SSESpecification`, and `OnDemandThroughput` fields through CreateTable, DescribeTable, and UpdateTable.

- `TableClass` persists as text and is returned in `TableClassSummary` on describe
- `SSESpecification: { Enabled: true }` returns `SSEDescription` with `Status: ENABLED`, `SSEType: KMS`, and a synthesized `KMSMasterKeyArn`
- `OnDemandThroughput` persists `MaxReadRequestUnits` / `MaxWriteRequestUnits` and round-trips through create/update/describe
- Adds `OnDemandThroughput` struct to core types and `KMSMasterKeyArn` field to `SseDescription`

These are passthrough fields only, no behavioral enforcement (no throttling, no encryption, no storage tiering).

## Why

Applications that set these fields (CDK, Terraform, SDK calls) should not error out, and tools that read them back (drift detection, state reconciliation) should get what they expect.

## Testing done

- `cargo test --workspace` — all pass (5 new unit tests for serde round-trips)
- New `tests/test_config_fields.py` — 6 integration tests covering create, describe, and update for all 3 fields (all pass against ExtendDB)
- Full pytest suite: 763 passed, no new failures

## Checklist

ADR / RFC: n/a

---


crates/core/src/types/table.rs
```diff
@@ -414,6 +429,10 @@ pub struct UpdateTableInput {
     pub attribute_definitions: Option<Vec<AttributeDefinition>>,
     #[serde(rename = "StreamSpecification")]
     pub stream_specification: Option<StreamSpecification>,
+    #[serde(rename = "TableClass")]
```
jcshepherd, line 432: TableClassSummary?

crates/core/src/types/table.rs
```diff
@@ -544,3 +563,72 @@ pub struct DescribeLimitsOutput {
...
+    fn on_demand_throughput_deserializes_from_input() {
+        let input_json = r#"{"MaxReadRequestUnits": 10, "MaxWriteRequestUnits": 5}"#;
+        let odt: OnDemandThroughput = serde_json::from_str(input_json).unwrap();
+        assert_eq!(odt.max_read_request_units, Some(10));
+        assert_eq!(odt.max_write_request_units, Some(5));
+    }
+
+    #[test]
+    fn sse_description_serializes_with_kms_arn() {
+        let sse = SseDescription {
```
jcshepherd, line 594: Does SSEDescription include InaccessibleEncryptionDateTime ? Should it? (Doesn't necessarily have to be part of this review, but it looks to be missing.)

crates/storage-postgres/src/create_table.rs
```diff
@@ -386,8 +397,29 @@ impl PostgresEngine {
...
+                let enabled = spec
+                    .get("Enabled")
+                    .and_then(|v| v.as_bool())
+                    .unwrap_or(false);
+                if enabled {
+                    Some(SseDescription {
+                        status: "ENABLED".to_string(),
+                        sse_type: Some(SseType::KMS),
+                        kms_master_key_arn: Some(format!(
+                            "arn:aws:kms:us-east-1:{}:key/default",
```
jcshepherd: Is region relevant here at all? Not for ExtendDB functionally, but would customers be surprised if they've configured to run in a different region?

jcshepherd (commented):
Overall looks okay: couple minor questions.

crates/core/src/types/table.rs
LeeroyHannigan: Input is TableClass, output is TableClassSummary

crates/core/src/types/table.rs
LeeroyHannigan: Not needed as we don't simulate inaccessible keys. The field would always be null. Can add later if a customer scenario requires it.

crates/storage-postgres/src/create_table.rs
LeeroyHannigan: Definitely, nice catch. Forgot to remove this after changing it for testing.

crates/core/src/types/table.rs
jcshepherd: Hmm. Okay. It's part of the DDB public API model though, isn't it? I'm okay with taking care of it separately, but it's not clear to me that it can be left out from an API model conformance perspective. (Unless we are accepting it in the API but just ignoring it?)

jcshepherd (approved):
One more question about the SSE thingie, but shouldn't block this.

crates/core/src/types/table.rs
LeeroyHannigan: Just ignoring it, yes. Its completely pass through. We have no way to simulate a non-conformant KMS key. Because we don't do KMS encryption.

PR #158, opened 2026-06-02
fix: SET snapshot semantics and parenthesised arithmetic

LeeroyHannigan:
## What

Fix two UpdateItem expression evaluation bugs:

1. **SET snapshot semantics**: Multiple SET clauses now evaluate their RHS against the pre-update item snapshot. Previously `SET a = :v, b = a` would assign `b` the *new* value of `a` instead of the original.

2. **Parenthesised arithmetic**: `SET c = (c - :v)` now parses correctly. The SET value parser previously rejected `(` as an unexpected token.

## Why

DynamoDB evaluates the entire UpdateExpression against a single pre-update snapshot, no SET clause can observe the effect of another SET clause in the same request. This is documented behavior that applications rely on (e.g., swap patterns like `SET a = b, b = a`).

Parenthesised arithmetic is valid DynamoDB syntax and used by SDKs and ORMs that defensively wrap expressions.

## Testing done

- Verified both behaviors against **real DynamoDB** first
- `cargo test --workspace` — all pass (2 new unit tests)
- New `tests/test_set_evaluation.py` — 3 integration tests (snapshot semantics, subtraction in parens, addition in parens), all pass against ExtendDB

## Checklist

ADR / RFC: n/a

---


pdf-amzn:
This creates mutual recursion (parse_set_operand → parse_value_expr → parse_set_operand) without a depth limit. Deeply nested parens like SET c = ((((((c - :v)))))) would recurse unboundedly. In practice, DynamoDB's 4KB expression limit caps nesting at ~1000 levels, which shouldn't blow the stack on most systems. The condition parser has a 150-depth limit for reference. Should we do the same here?

LeeroyHannigan:
> This creates mutual recursion (parse_set_operand → parse_value_expr → parse_set_operand) without a depth limit. Deeply nested parens like SET c = ((((((c - :v)))))) would recurse unboundedly. In practice, DynamoDB's 4KB expression limit caps nesting at ~1000 levels, which shouldn't blow the stack on most systems. The condition parser has a 150-depth limit for reference. Should we do the same here?

Good catch on both fronts. Updated: `parse_update_from` now calls `check_redundant_parens` upfront (same function used by the condition/filter parsers), so `SET c = ((c - :v))` is rejected with the canonical "redundant parentheses" message. Verified against real DDB that single parens are accepted and double parens are rejected.
  
This also eliminates the unbounded recursion concern, since `((` is rejected before parsing begins, the `LParen` → `parse_value_expr` → `parse_set_operand` path can only recurse once (for a single valid paren group).

crates/core/src/expression/update_parser.rs
```diff
@@ -513,4 +532,19 @@ mod tests {
...
+    fn redundant_parens_rejected() {
+        let err = parse("SET c = ((c - :v))").unwrap_err();
+        assert!(
+            matches!(&err, DynamoDbError::ValidationException(msg) if msg.contains("redundant parentheses")),
+            "Expected redundant parens error, got: {err:?}"
+        );
+    }
+
+    #[test]
+    fn single_parens_accepted() {
```
jcshepherd, line 546: If you wanted bonus points, could check for error on unmatched parens.

jcshepherd (dismissed):
One suggestion for an additional unit test but looks good otherwise.

jcshepherd (approved):

PR #162, opened 2026-06-03
fix: validate expression attributes and label parse errors per parameter

yesyayen:
## What

This makes expression-attribute validation and expression parse-error messages match Amazon DynamoDB across the single-item and query/scan APIs. Three related gaps:

1. `GetItem` and `BatchGetItem` did not reject `ExpressionAttributeNames` entries that the projection never references. Every other handler (Query, Scan, PutItem, UpdateItem, DeleteItem) already did.
2. No handler rejected `ExpressionAttributeNames`/`ExpressionAttributeValues` supplied when there is no expression that could reference them. `PutItem` had a partial check, but it appended a suffix to the names message that Amazon DynamoDB does not emit.
3. Expression parse errors used a generic `Invalid expression:` prefix instead of the parameter-specific `Invalid ProjectionExpression:` / `Invalid ConditionExpression:` / `Invalid FilterExpression:` / `Invalid KeyConditionExpression:` / `Invalid UpdateExpression:`.

I also added an `ExpressionKind` enum in `extenddb-core` so the parameter label that drives both the error prefix and the value-suffix is defined once and is typo-safe, instead of being repeated as string literals at each call site.

The table shows what ExtendDB returns before this PR and what Amazon DynamoDB returns. I captured every Amazon DynamoDB message directly via the AWS CLI.

| Scenario | ExtendDB (before this PR) | Amazon DynamoDB |
|---|---|---|
| GetItem with `ProjectionExpression="id"` and `ExpressionAttributeNames={"#abc":"foo"}` (declared, never referenced) | Returns the item, no error | `ValidationException: Value provided in ExpressionAttributeNames unused in expressions: keys: {#abc}` |
| GetItem / BatchGetItem / Scan / Query with `ExpressionAttributeNames` but no expression that uses a name | Returns the item, no error | `ValidationException: ExpressionAttributeNames can only be specified when using expressions` |
| PutItem with `ExpressionAttributeNames` and no `ConditionExpression` | `ValidationException: ExpressionAttributeNames can only be specified when using expressions: ConditionExpression is null` (extra suffix) | `ValidationException: ExpressionAttributeNames can only be specified when using expressions` (no suffix) |
| Scan with `ExpressionAttributeValues` and no `FilterExpression` | Runs the scan, no error | `ValidationException: ExpressionAttributeValues can only be specified when using expressions: FilterExpression is null` |
| DeleteItem with `ExpressionAttributeValues` and no `ConditionExpression` | Deletes, no error | `ValidationException: ExpressionAttributeValues can only be specified when using expressions: ConditionExpression is null` |
| UpdateItem with `ExpressionAttributeValues` and no `UpdateExpression`/`ConditionExpression` | Runs the update, no error | `ValidationException: ExpressionAttributeValues can only be specified when using expressions: UpdateExpression and ConditionExpression are null` |
| Query with legacy `KeyConditions` and a stray `ExpressionAttributeValues` | Runs the query, no error | `ValidationException: ExpressionAttributeValues can only be specified when using expressions` (Query emits no suffix) |
| GetItem with `ProjectionExpression="status"` (a reserved word) | `ValidationException: Invalid expression: Attribute name is a reserved keyword; reserved keyword: status` | `ValidationException: Invalid ProjectionExpression: Attribute name is a reserved keyword; reserved keyword: status` |
| PutItem/DeleteItem `ConditionExpression`, Scan/Query `FilterExpression`, Query `KeyConditionExpression`, UpdateItem `UpdateExpression` with a reserved word | `Invalid expression: ...` | `Invalid <that parameter>: ...` |

## Out of scope

I kept this PR focused on the single-item and query/scan validation and the error prefix. The following gaps will be addressed in separate PRs:

- `TransactWriteItems` and `TransactGetItems` are not wired into this validation yet. Their expression parse errors and "can only be specified" handling are unchanged. This needs its own message capture pass because transactions can format errors by item index.
- Only the error prefix is corrected, not the full syntax-error body text. For example a negative list index in a projection gets the right `Invalid ProjectionExpression:` prefix but not yet the exact token/near wording Amazon DynamoDB includes.
- Nested projection correctness (for example dropping a `NULL` marker inside a list, or list-element projection) is unchanged.
- Validation of the `ExpressionAttributeNames` map entries themselves (empty key, over-long key, duplicate, over-long or empty mapped value, too many names) is unchanged.
- Expression size-limit validation is unchanged.
- Rejecting a mix of `AttributesToGet` and expression parameters inside `BatchGetItem` is unchanged.

## Why

These are conformance gaps found by comparing ExtendDB against Amazon DynamoDB with the AWS CLI.

## Testing done

Added three pytest modules; every expected message was captured from Amazon DynamoDB:

- `tests/test_unused_expression_attribute_names.py` (5 cases): unused names on GetItem/BatchGetItem, including the missing-item case that proves validation runs before the lookup, plus positive controls.
- `tests/test_expression_param_usage.py` (12 cases): names and values with no referencing expression across GetItem, BatchGetItem, Scan, DeleteItem, UpdateItem, PutItem, and Query, asserting the exact per-API suffix.
- `tests/test_expression_error_prefix.py` (7 cases): the per-parameter prefix for all five expression kinds, for both reserved-word and syntax errors.

Results:

- New tests: 24 pass against ExtendDB; the same cases pass against Amazon DynamoDB.
- Regression: `tests/` item, batch, query/scan, conditional-write, transaction, and expression suites pass (325 cases), no new failures.
- `cargo test --workspace` passes; `cargo fmt --check` clean; `cargo clippy -- -D warnings` clean for the touched crates.

## Checklist

ADR / RFC: n/a.

## Breaking changes

n/a

---

### DIFF
--- a/crates/core/src/expression/resolver.rs
+++ b/crates/core/src/expression/resolver.rs
@@ -200,6 +200,51 @@ pub fn resolve_element_name<'a>(
     }
 }
 
+/// Reject `ExpressionAttributeNames`/`Values` supplied with no expression that
+/// can reference them. Matches Amazon DynamoDB:
+///
+/// - names: `ExpressionAttributeNames can only be specified when using expressions`
+///   (no suffix), for every API.
+/// - values: `ExpressionAttributeValues can only be specified when using
+///   expressions: <params> is/are null`, listing the value-capable expression
+///   parameter(s) of the API.
+///
+/// `name_expr_present` is true when any expression that can carry a `#name` is
+/// present (projection, condition, filter, key condition, update).
+/// `value_expr_present` is true when any expression that can carry a `:value` is
+/// present (condition, filter, key condition, update; not projection).
+/// `value_expr_params` is the API's static list of value-capable expression
+/// parameter names, used to build the values suffix.
+///
+/// # Errors
+///
+/// Returns `DynamoDbError::ValidationException` when a map is supplied with no
+/// referencing expression.
+pub fn validate_expression_param_usage(
+    names: Option<&HashMap<String, String>>,
+    name_expr_present: bool,
+    values: Option<&HashMap<String, AttributeValue>>,
+    value_expr_present: bool,
+    value_expr_params: &[&str],
+) -> Result<(), DynamoDbError> {
+    if !name_expr_present && names.is_some_and(|m| !m.is_empty()) {
+        return Err(DynamoDbError::ValidationException(
+            "ExpressionAttributeNames can only be specified when using expressions".to_owned(),
+        ));
+    }
+    if !value_expr_present && values.is_some_and(|m| !m.is_empty()) {
+        let suffix = match value_expr_params {
+            [] => String::new(),
+            [one] => format!("{one} is null"),
+            many => format!("{} are null", many.join(" and ")),
+        };
+        return Err(DynamoDbError::ValidationException(format!(
+            "ExpressionAttributeValues can only be specified when using expressions: {suffix}"
+        )));
+    }
+    Ok(())
+}
+
 /// Validate that all provided ExpressionAttributeNames/Values were used.
 ///
 /// Collects `#name` and `:value` references from the given expressions,
--- a/crates/core/src/expression/resolver.rs
+++ b/crates/core/src/expression/resolver.rs
@@ -214,7 +214,8 @@ pub fn resolve_element_name<'a>(
 /// `value_expr_present` is true when any expression that can carry a `:value` is
 /// present (condition, filter, key condition, update; not projection).
 /// `value_expr_params` is the API's static list of value-capable expression
-/// parameter names, used to build the values suffix.
+/// parameter kinds, used to build the values suffix. An empty list yields no
+/// suffix (matching Query).
 ///
 /// # Errors
 ///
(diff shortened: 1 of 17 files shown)


crates/core/src/expression/resolver.rs
```diff
@@ -200,6 +200,58 @@ pub fn resolve_element_name<'a>(
...
+            "ExpressionAttributeNames can only be specified when using expressions".to_owned(),
+        ));
+    }
+    if !value_expr_present && values.is_some_and(|m| !m.is_empty()) {
+        let base = "ExpressionAttributeValues can only be specified when using expressions";
+        let msg = match value_expr_params {
+            [] => base.to_owned(),
+            [one] => format!("{base}: {one} is null"),
+            many => {
+                let joined = many
```
pdf-amzn, line 242: Is there a limit to how many things it makes sense to join here? An unbounded-length (or even very long) error message makes me nervous.

yesyayen: this is bounded by the number of items in `ExpressionKind` enum which is of length 5 (valid [expression types](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.html)). It is never derived from customer input, thus the message length does not grow with the request.

Also, the max possible value-capable param per API is `2` today.
1. UpdateItem: UpdateExpression + ConditionExpression
2. Query: KeyConditionExpression + FilterExpression 

So length isn't the concern, grammar is. If a future API ever errors with 3+ value-capable expressions, this would join as `A and B and C` instead of `A, B and C`. I believe the conformance tests we add would catch it.

pdf-amzn (approved):

PR #167, opened 2026-06-05
refactor: decouple StorageConfig from postgres feature

LeeroyHannigan:
## What

Gate `StorageConfig`'s `Default` impl and the `default_backend()` function behind `#[cfg(feature = "postgres")]`. Remove `#[serde(default)]` from the `storage` field on `AppConfig` and drop the `Default` derive from `AppConfig`.

## Why

The binary fails to compile with `--no-default-features` because `StorageConfig::default()` unconditionally references `extenddb_storage_postgres::PostgresStorageConfig`. This blocks building backend-specific binaries (e.g. `--no-default-features --features memory`) which is a prerequisite for per-backend Docker images and the multi-backend distribution strategy.

The `#[serde(default)]` fallback was never exercised in practice — `extenddb init` always generates a config with an explicit `[storage]` section. Removing it makes the contract clear: every config must declare its backend.

## Testing done

- `cargo build` (default features) — compiles ✓
- `cargo build --no-default-features` — compiles ✓ (previously failed)
- `cargo clippy --all-targets -- -D warnings` — clean for both feature sets
- `cargo test --workspace` — all pass
- Full integration pytest (945 passed, 0 failed)

## Checklist

ADR / RFC: n/a

## Breaking changes

Config files without a `[storage]` section will now fail to parse instead of silently defaulting to postgres. This only affects hand-written configs that omit the section — `extenddb init` has always generated it.

---

### DIFF
--- a/crates/bin/src/config.rs
+++ b/crates/bin/src/config.rs
@@ -6,12 +6,11 @@
 use extenddb_core::limits::LimitsConfig;
 use serde::Deserialize;
 
-#[derive(Debug, Clone, Default, Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct AppConfig {
     #[serde(default)]
     pub server: ServerConfig,
-    #[serde(default)]
     pub storage: StorageConfig,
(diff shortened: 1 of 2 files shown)


jcshepherd (approved):
Sweet.

PR #170, opened 2026-06-05
fix: run-tests doesn't need postgres connection string

jcshepherd:
## What

devtools/run-tests failed silently when it failed to parse a Postgresql connection string from config. This change checks to see if ExtendDB is even using PG for test, and skips the PG connection string env variable setup if it's not.

## Why

Enable run-tests to initiate tests against other backend impls.

## Testing done

cargo build --release --features cassandra
:
devtools/run-tests --extenddb pytest
(correctly initializes test environment and runs the pytests ... some of which even pass!)

## Checklist

- [Y] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [Y] All tests pass (`cargo test --workspace`)
- [Y] Code is formatted (`cargo fmt --check`)
- [Y] Clippy is clean (`cargo clippy -- -W clippy::pedantic`)
- [N/A] I have added or updated tests for new functionality
- [N/A] I have updated documentation if behavior changed
- [N/A] Breaking changes are noted below (if any)
- [N/A] If this changes the wire protocol, `Storage` trait, auth model, on-disk

---


pdf-amzn (dismissed):

amrith (commented):
lgtm

amrith (commented):
lgtm

devtools/run-tests
amrith, line 1: Nit: This regex will also match commented-out lines, e.g.:


> # backend = "postgres"   ← old default
> backend = "cassandra"
> 

Anchoring to the start of line would avoid the most common false-positive:

> if grep -q '^[[:space:]]*backend[[:space:]]*=[[:space:]]*"postgres"' "$CONFIG_FOR_SETTINGS" 2>/dev/null; then
>

Not a blocker since the config is auto-generated, but cheap insurance given the PR's goal is robustness for non-PG backends.

amrith (commented):
posted a comment on run-tests@349

devtools/run-tests
jcshepherd: Actually, that _should_ be blocker ... the original regex was embarrassingly lenient. Fixed.

pdf-amzn (approved):
Sigh. All this regex fun suggests this might be better off as something other than a shell script. But that's a bigger change, and this isn't the first instance of using grep to interpret the CONFIG_FOR_SETTINGS file

PR #175, opened 2026-06-17
docs: add ADR-0003 sqlx::migrate catalog/data migrations

yesyayen:
## What

Adds ADR-0003 recording the decision to adopt `sqlx::migrate` for PostgreSQL
catalog and data schema migrations.

## Why

The homegrown migration runner tracked files by name with no checksum, so an
in-place edit of an already-shipped migration produced a silent schema drift that
broke `CreateTable`. The ADR records the choice to replace it with sqlx's built-in
migration framework (which checksums each file and refuses to run when one changed
after it was applied).

Closes #

## Testing done

N/A

## Checklist

- [x] All tests pass (`cargo test --workspace`) — n/a, docs only
- [x] Code is formatted (`cargo fmt --check`) — n/a, docs only
- [x] Clippy is clean — n/a, docs only
- [ ] I have added or updated tests for new functionality — n/a
- [x] ADR captures the decision (link below)

ADR / RFC: `docs/adr/0003-catalog-migration-mechanism.md` (added in this PR)

## Breaking changes

N/A

---

### DIFF
--- a/docs/adr/0003-catalog-migration-mechanism.md
+++ b/docs/adr/0003-catalog-migration-mechanism.md
@@ -48,11 +48,11 @@ Relevant facts:
    guard test, and move the version write inside the runner. It stays fully under
    our control, but it keeps reimplementing what sqlx already does and leaves us
    maintaining a migration runner of our own.
-2. **Adopt sqlx:migrate** We would switch to sqlx and seed its `_sqlx_migrations`
+2. **Adopt sqlx, upgrade in place.** We would switch to sqlx and seed its `_sqlx_migrations`
    table to match what each existing catalog already has, so no re-initialization
    is needed. This avoids a breaking change, but getting that seeding exactly
    right for every existing catalog is delicate and error-prone.
-3. **Adopt sqlx:migrate** We switch to sqlx and accept that an existing catalog has to be
+3. **Adopt sqlx, re-initialize.** We switch to sqlx and accept that an existing catalog has to be
    torn down and re-initialized. There is no compatibility shim and no in-place
    upgrade.
 
--- a/docs/adr/0003-catalog-migration-mechanism.md
+++ b/docs/adr/0003-catalog-migration-mechanism.md
@@ -60,6 +60,8 @@ Relevant facts:
 
 Adopt `sqlx::migrate` directly for both the catalog and data database (the re-init
 option): existing catalogs re-initialize, and we do not build a compatibility shim.
+Re-initializing drops the databases, so existing tables and items are lost and must
+be recreated and reloaded after `init`.
 
 ## Rationale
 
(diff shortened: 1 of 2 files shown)


docs/adr/0003-catalog-migration-mechanism.md
```diff
@@ -0,0 +1,139 @@
...
+   is needed. This avoids a breaking change, but getting that seeding exactly
+   right for every existing catalog is delicate and error-prone.
+3. **Adopt sqlx:migrate** We switch to sqlx and accept that an existing catalog has to be
+   torn down and re-initialized. There is no compatibility shim and no in-place
+   upgrade.
+
+## Decision
+
+Adopt `sqlx::migrate` directly for both the catalog and data database (the re-init
+option): existing catalogs re-initialize, and we do not build a compatibility shim.
```
jcshepherd, line 62: What does this mean for existing user-data/tables?

docs/adr/0003-catalog-migration-mechanism.md
```diff
@@ -0,0 +1,139 @@
...
+**The breaking change**
+
+- Upgrading requires `destroy` + `init`. `destroy` drops the databases, so all
+  stored items are wiped, not just schema. We document this in the upgrade manual
+  and changelog. It is acceptable only at v0.1 with no dependent catalogs.
+
+**Version gate and the version write**
+
+- The catalog version gate stays: the server refuses to start when the binary's
+  expected version and the stored `catalog_version` disagree. It is catalog-only;
```
jcshepherd: Can you clarify what "disagree" means?

docs/adr/0003-catalog-migration-mechanism.md
```diff
@@ -0,0 +1,139 @@
...
+  the data database relies on its `_sqlx_migrations` table and has no separate
+  version. The gate is symmetric, so an older binary against a newer catalog also
```
jcshepherd: What are the implications for upgrading the binary? E.g., suppose I'm running 6 ExtendDB 1.6 front-ends on top of my PG installation. I want to upgrade to xDB 1.7 and the schema has changed. Do I need to take a complete front-end outage while I do the upgrade?

jcshepherd (commented):
A few questions below ...

docs/adr/0003-catalog-migration-mechanism.md
yesyayen: The binary has a compiled-in `CATALOG_VERSION`; the catalog DB stores a `catalog_version` row; if the two strings are not identical, startup fails. "Disagree" is just exact-string inequality (in either direction), so a version mismatch is caught at startup

docs/adr/0003-catalog-migration-mechanism.md
yesyayen: Yes, we are introducing incompatible breaking changes. The existing databases are dropped and re-created empty, so all tables and items are lost. Documented reasoning on why this is okay, let me know if we want to re-consider this decision.

this is documented under `Consequences` section,
> Upgrading requires destroy + init. `destroy` drops the databases, so all stored items are wiped, not just schema. We document this in the upgrade manual and changelog. It is acceptable only at v0.1 with no dependent catalogs.

docs/adr/0003-catalog-migration-mechanism.md
yesyayen: yeah, you're right. A schema-changing upgrade does need the front-ends to go down during upgrade.

That's because the server and the Postgres catalog are version-locked today: at startup `check_catalog_version` refuses to start unless the binary's compiled `CATALOG_VERSION` exactly matches the catalog's stored catalog_version (symmetric, either direction), and the engine's queries are written for that exact schema. So a schema bump means restarting every front-end on the matching version. Same as the current setup.

Scope of this PR is just to move us off the ad-hoc homegrown migrator onto a well-supported sqlx-based mechanism. The current one tracks migrations by filename only, so it didn't catch the recent drift and it surfaced at request time; sqlx checksums each file and the gate catches a mismatch at startup.

Backward-compatible / online (no-outage) migration is out of scope IMO. It belongs after we decouple the FE from the engines and have a proper per-engine design (expand-contract + a version-range gate, per-engine degradation, and such). I'll add it as out-of-scope / future work in the ADR so it doesn't slip through the cracks.

jcshepherd (approved):

PR #176, opened 2026-06-22
RFC-0002: Backend plugin contributor guidelines

jcshepherd:
## What

RFC for accepting and managing backend database plug-ins

## Why

A number of people inside and outside the ExtendDB organization are interested in building and contributing additional database backends for ExtendDB (in addition to the reference PostgreSQL backend). The intent of this RFC, once published is to communicate guidelines and expectations to the project and to backend implementators for developing, contributing and maintaining new backends.

## Testing done

Doc only.

## Checklist

- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [N/A] All tests pass (`cargo test --workspace`)
- [X] Code is formatted (`cargo fmt --check`)
- [X] Clippy is clean (`cargo clippy -- -W clippy::pedantic`)
- [N/A] I have added or updated tests for new functionality
- [X] I have updated documentation if behavior changed
- [N/A] Breaking changes are noted below (if any)
- [N/A] If this changes the wire protocol, `Storage` trait, auth model, on-disk

---

### DIFF
--- a/docs/rfcs/0002-backend-plugin-policy.md
+++ b/docs/rfcs/0002-backend-plugin-policy.md
@@ -1,0 +1,239 @@
+# RFC-0002: Backend Plugin Development and Maintenance Policy
+
+- Status: Draft
+- Author: @jcshepherd
+- Created: 2026-06-22
+- FCP ends: (set when entering Final Comment Period)
+- Tracking issue: #NNN
+
+## Summary
+
+This RFC establishes policies for developing, contributing, and maintaining database backend plugins for ExtendDB. It recommends a mono-repo architecture where all officially supported backends live as separate crates within the main ExtendDB repository, defines conformance requirements for backend acceptance, and establishes processes for external contributions.
+
+## Motivation
+
+ExtendDB's core value proposition is providing an Amazon DynamoDB-compatible API over multiple database backends. The reference PostgreSQL backend demonstrates feasibility. Active development of Apache Cassandra and SQLite backends shows demand for backend diversity. As the project grows, we need policies that:
+
+1. **Scale beyond the core team.** The ExtendDB organization cannot personally implement and maintain every backend indefinitely. We must enable external contributions while ensuring quality.
+
+2. **Preserve DynamoDB compatibility.** ExtendDB's value is compatibility with Amazon DynamoDB. Any backend released by the ExtendDB organization must demonstrate conformance with DynamoDB's APIs and features.
+
+3. **Support backend diversity as a feature.** Different backends serve different needs: PostgreSQL (reference implementation), Cassandra (horizontal scale), SQLite (testing/embedded), and future backends extend ExtendDB's reach in terms of the infrastructure it can operate on and the use-cases it can support.
+
+4. **Maintain velocity while managing complexity.** We must balance fast iteration for a small team with the eventual need for modular external contributions.
+
+Without formal policies, we risk inconsistent quality, fragmented implementations, and contributor confusion. This RFC provides the framework for sustainable growth.
+
+## Detailed design
+
+### Repository structure: Mono-repo with feature flags
+
+All officially supported backends live in the main `extenddb` repository as separate crates under `crates/storage-{backend}/`. Examples:
+
+- `crates/storage-postgres/` (reference implementation)
+- `crates/storage-cassandra/`
+- `crates/storage-sqlite/`
+
+Backends are selected at build time via Cargo feature flags:
+
+```toml
+[features]
+default = ["postgres"]
+postgres = ["dep:storage-postgres"]
+cassandra = ["dep:storage-cassandra"]
+sqlite = ["dep:storage-sqlite"]
+all-backends = ["postgres", "cassandra", "sqlite"]
+```
+
+Mono-repo maximizes development velocity while avoiding the operational overhead of managing N separate repositories with N CI pipelines, N release processes, N issue trackers and N documentation sites. Cross-cutting changes spanning frontend and backend require one PR, one review, one merge.
+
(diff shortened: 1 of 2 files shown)


docs/rfcs/0002-backend-plugin-policy.md
```diff
@@ -0,0 +1,239 @@
...
+
+Backends released under the ExtendDB organization must adhere to a trait-based conformance model:
+
+1. **Mandatory traits.** All backends must implement the core `Storage` traits (data plane and control plane operations) and pass 100% of conformance tests for those traits. Core traits are the minimum required for a functional DynamoDB-compatible backend.
+
+2. **Optional traits.** Backends may choose not to implement optional traits (example: Streams, Transactions) or may stub them out to return `Unimplemented` errors. However, if a backend *does* implement an optional trait, it must pass 100% of conformance tests for that trait. Partial implementations are not permitted.
+
+3. **All-or-nothing per trait.** The trait is the finest-grained unit of conformance. Backends cannot claim partial support for a trait (example: "Streams partially works but records are incomplete" is not allowed). Either the trait is fully implemented and conformant, or it is not implemented.
+
+4. **Maintain semantic correctness.** Where a backend implements an operation, it must match DynamoDB behavior including error responses, pagination, atomicity guarantees, and consistency models.
```
yesyayen: conformant backend should also match DynamoDB's isolation level (along with Atomicity and Consistency)

docs/rfcs/0002-backend-plugin-policy.md
```diff
@@ -0,0 +1,239 @@
...
+
+This model reduces friction by allowing new backends to launch with a well-defined subset of functionality without requiring support for every ExtendDB feature from day one. As backends mature, they can add optional traits incrementally, but each addition must be complete and correct. This prevents fragmented, partially-working implementations that erode ExtendDB's compatibility guarantee.
+
+#### Trait classification
+
+ExtendDB's storage abstraction consists of 13 traits covering storage operations and management functions. The following classification defines which traits are mandatory for backend acceptance and which are optional:
+
+**Mandatory storage traits:**
+- `TableEngine` - table-level operations
+- `DataEngine` - item-level data operations
```
yesyayen, line 128: Transactions (TransactGet and TransactWrite) APIs are under DataEngine trait but we categorized them as `Optional trait` above. 
I think transactions should be optional: not all backends can provide DynamoDB's isolation (serializable for TransactWriteItems).

docs/rfcs/0002-backend-plugin-policy.md
```diff
@@ -0,0 +1,239 @@
...
+
+3. **All-or-nothing per trait.** The trait is the finest-grained unit of conformance. Backends cannot claim partial support for a trait (example: "Streams partially works but records are incomplete" is not allowed). Either the trait is fully implemented and conformant, or it is not implemented.
+
+4. **Maintain semantic correctness.** Where a backend implements an operation, it must match DynamoDB behavior including error responses, pagination, atomicity guarantees, and consistency models.
+
+5. **Clear documentation.** Backends must document in their README which optional traits they implement and which they do not. Conformance test results (per-trait pass rates) must be published and tracked in CI.
+
+This model reduces friction by allowing new backends to launch with a well-defined subset of functionality without requiring support for every ExtendDB feature from day one. As backends mature, they can add optional traits incrementally, but each addition must be complete and correct. This prevents fragmented, partially-working implementations that erode ExtendDB's compatibility guarantee.
+
+#### Trait classification
```
yesyayen, line 122: [CredentialStore](https://github.com/jcshepherd/extenddb/blob/948150b1ee7ef5f482c6b2746ba1982693dbc986/crates/storage-postgres/src/credential_store.rs) is missing in this classification. 

Related: on line 90 we mark `AuthorizationStore - authorization data` as an
optional trait. Is that intentional? Since auth (SigV4 + IAM) is mandatory in
ExtendDB, I believe both CredentialStore and AuthorizationStore should be
mandatory traits

docs/rfcs/0002-backend-plugin-policy.md
```diff
@@ -0,0 +1,239 @@
...
+
+ExtendDB's storage abstraction consists of 13 traits covering storage operations and management functions. The following classification defines which traits are mandatory for backend acceptance and which are optional:
+
+**Mandatory storage traits:**
```
yesyayen, line 126: should we consider defining an explicit, stable (immutable!?) set of control-plane and data-plane operations that are mandatory for conformance, decoupled from trait boundaries, instead of using whole traits as the unit (at least for storage)?

That lets capabilities like transactions or streams be optional even when they share a trait with mandatory operations. It would need a per-operation enforcement mechanism and a policy that keeps the core set stable so backends don't silently drift or go stale as it changes

yesyayen (commented):
I'm aligned with the recommended approach of mono-repo. Comments are on inconsistencies and some suggestions.

docs/rfcs/0002-backend-plugin-policy.md
jcshepherd: Hmmm. I think "intentional" would be an overstatement. Thanks for the catch.

jcshepherd:
I added a section to go over an API-oriented acceptance criteria, versus the trait-oriented approach the document started with. They both are rough-edged: there's just not a real clean alignment between traits and API functionality, and even if we are able to refactor everything so there is, it'll probably be difficult to maintain. Cross-cutting concerns are hard to avoid over time. But, we should debate which approach seems easiest for developers to conform to and most likely to evolve gracefully.

yesyayen (approved):
I'm aligned with the recommended mono-repo approach. Left comments about inconsistencies and suggestions.

PR #182, opened 2026-06-26
feat: SQLite storage backend (in-tree) with developer mode

LeeroyHannigan:
Bring the SQLite storage backend into the workspace as crates/storage-sqlite, per RFC-0002 (backends as in-tree crates). PostgreSQL remains the default; SQLite is selected by Cargo feature. Targets local development, CI, and embedded/air-gapped use.

Consolidation:
- crates/storage-sqlite as a workspace member using workspace dependencies.
- bin features: `sqlite` and `sqlite-memory` (zero-config ephemeral in-memory).
- Uniform backend linking: each backend exposes `link()` and the bin's `backends::link_all()` references every enabled backend, so inventory registrations are never dead-stripped.

Developer mode (build-time `dev-mode` feature):
- Plain HTTP on loopback and open IAM authorization, with SigV4 still verified.
- Credential adopted from the standard AWS_* env, else AWS's documented example.
- A compile_error guard makes `dev-mode` + the Postgres backend fail to build, so an insecure production binary cannot exist.

Fixes found bringing the backend to parity:
- Resolve a relative SQLite path to absolute before daemonizing (chdir to / otherwise breaks file-backed serve in the default daemon mode).
- Remove a stale PID file before daemonizing (spurious "failed to start").
- Generate extenddb-prefixed access keys/secrets; reject non-ASCII identifiers (matches PostgreSQL).
- Hash-only GSI scan pagination on a composite-key base table no longer skips rows (include the base sort key in ORDER BY and the pagination predicate).
- Seed a self-service IAM policy on user creation.
- Restrict the database file (and WAL/SHM sidecars) to 0600.
- Parse session tags in array and object form; treat expired sessions as absent.
- Mark a restored table ACTIVE immediately.
- Bind every placeholder in delete_index_row_multi.

Testing:
- SQLite integration job added to CI, mirroring the Postgres job.
- run-tests works with non-Postgres backends; import/export allowed paths use $TMPDIR.
- Conformance green against SQLite: 721 pytest + 326 comprehensive, plus fmt, clippy -D warnings, and cargo test --workspace.

## Checklist

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] All tests pass (`cargo test --workspace`)
- [ ] Code is formatted (`cargo fmt --check`)

ADR / RFC: 

## Breaking changes

---

### DIFF
--- a/crates/storage-sqlite/src/data/index.rs
+++ b/crates/storage-sqlite/src/data/index.rs
@@ -1,0 +1,488 @@
+// Copyright 2026 ExtendDB contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! GSI/LSI index maintenance for the SQLite backend.
+//!
+//! Synchronous indexes (LSIs and zero-delay GSIs) are reconciled in the base
+//! write transaction. Async GSIs are deferred via the `gsi_pending` queue —
+//! **one self-describing row per index**: each row snapshots the base key
+//! schema, attribute definitions, and its single target index definition, so
+//! the worker applies with zero catalog reads. Each row carries its index's own
+//! (jittered) propagation delay, and `ready_at` is kept monotonic within the
+//! base key's `worker_partition` so jitter cannot reorder updates to one item.
+//! Index key columns follow D2 — `N` keys are stored as the order-preserving
+//! TEXT encoding, `S` as TEXT, `B` as BLOB — via [`sk_bound`].
+
+use extenddb_core::types::{
+    AttributeDefinition, Item, KeySchemaElement, Projection, ProjectionType, ScalarAttributeType,
+    TableKeyInfo,
+};
+use extenddb_storage::error::StorageError;
+use extenddb_storage::util::{composite_pk_to_text, parse_sk, sk_column, sk_column_n};
+use serde::{Deserialize, Serialize};
+
+use super::{BoundValue, all_sort_key_info, index_table_name, sk_bound};
+
+/// Number of partitions a base key can hash to. Updates to one base item share
+/// a partition; `ready_at` is clamped monotonic within a partition so the
+/// single drain worker (which orders by `id`) applies them in order. More
+/// partitions reduce false serialization between unrelated keys; there is no
+/// concurrency cost because SQLite has a single writer.
+const NUM_PARTITIONS: u64 = 16;
+
+/// Stable partition for a base-table key (FNV-1a, mapped to a partition). A
+/// local hash keeps the mapping fixed for a key across builds (`std`'s hasher
+/// is not guaranteed stable).
+fn partition_for(base_pk_text: &str) -> i64 {
+    let mut hash: u64 = 0xcbf2_9ce4_8422_2325; // FNV-1a offset basis
+    for byte in base_pk_text.as_bytes() {
+        hash ^= u64::from(*byte);
+        hash = hash.wrapping_mul(0x0000_0100_0000_01b3); // FNV prime
+    }
+    (hash % NUM_PARTITIONS) as i64
+}
+
+/// Jittered propagation delay (ms), uniform in `[delay_ms / 2, delay_ms]`.
+///
+/// Simulates DynamoDB's eventual-consistency variability (a fixed delay is
+/// perfectly predictable) while keeping the configured delay a meaningful lower
+/// bound — jittering all the way to ~0 would make the delay almost meaningless.
+/// `delay_ms <= 1` is returned unchanged. Callers only enqueue async indexes,
+/// so `delay_ms >= 1`.
+fn jitter_delay_ms(delay_ms: u64) -> u64 {
+    if delay_ms <= 1 {
+        delay_ms
+    } else {
+        use rand::Rng;
+        rand::rng().random_range(delay_ms / 2 + 1..=delay_ms)
+    }
+}
+
+/// A single target index definition, snapshotted at enqueue time. The
+/// propagation delay is not stored — it is already encoded in the row's
+/// `ready_at`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct GsiIndexDef {
+    pub(crate) index_id: String,
+    pub(crate) key_schema: Vec<KeySchemaElement>,
+    pub(crate) projection: Projection,
+}
+
+/// Self-describing apply context serialized into `gsi_pending.index_context`.
+/// One per row → exactly one target index, so the worker needs no catalog read.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct GsiApplyContext {
+    pub(crate) base_key_schema: Vec<KeySchemaElement>,
+    pub(crate) attribute_definitions: Vec<AttributeDefinition>,
+    pub(crate) index: GsiIndexDef,
+}
+
+/// Metadata for a single index, used on the write path and by the GSI worker.
+pub(crate) struct IndexMeta {
+    pub(super) index_id: String,
+    pub(super) index_name: String,
+    pub(super) index_type: String,
+    pub(super) key_schema: Vec<KeySchemaElement>,
+    pub(super) projection: Projection,
+    /// Per-GSI propagation delay (ms). `None` = use system default; `Some(0)` =
+    /// synchronous.
+    pub(super) propagation_delay_ms: Option<i64>,
+}
+
+/// Fetch all index metadata for a table from the catalog.
+pub(crate) async fn fetch_indexes_for_table(
+    table_id: &str,
+    pool: &sqlx::SqlitePool,
+) -> Result<Vec<IndexMeta>, StorageError> {
+    let rows: Vec<(String, String, String, String, String, Option<i64>)> = sqlx::query_as(
+        "SELECT index_id, index_name, index_type, key_schema, projection, propagation_delay_ms \
+         FROM indexes WHERE table_id = ?",
+    )
+    .bind(table_id)
+    .fetch_all(pool)
+    .await
+    .map_err(|e| StorageError::Internal(e.to_string()))?;
+
+    rows.into_iter()
+        .map(|(index_id, index_name, index_type, ks, proj, delay)| {
+            Ok(IndexMeta {
+                index_id,
+                index_name,
+                index_type,
+                key_schema: serde_json::from_str(&ks)
+                    .map_err(|e| StorageError::Internal(e.to_string()))?,
+                projection: serde_json::from_str(&proj)
+                    .map_err(|e| StorageError::Internal(e.to_string()))?,
+                propagation_delay_ms: delay,
+            })
+        })
+        .collect()
+}
+
+/// Effective GSI propagation delay (ms): per-GSI override, else system default.
+/// `Some(0)` forces synchronous; negative is treated as "use system default".
+pub(crate) fn effective_delay(idx: &IndexMeta, system_default: u64) -> u64 {
+    match idx.propagation_delay_ms {
+        Some(0) => 0,
+        Some(ms) if ms > 0 => ms as u64,
+        _ => system_default,
+    }
+}
+
+/// Enqueue async GSI propagation for a write: **one self-describing
+/// `gsi_pending` row per async index**, each honoring its own effective delay.
+///
+/// Returns the number of rows enqueued so callers can skip waking the worker
+/// when nothing was queued. Must run inside the base write transaction so the
+/// pending rows commit atomically with the item mutation.
+pub(crate) async fn enqueue_async_indexes(
+    tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>,
+    key_info: &TableKeyInfo,
+    indexes: &[IndexMeta],
+    old_item: Option<&Item>,
+    new_item: Option<&Item>,
+    system_default_delay: u64,
+) -> Result<usize, StorageError> {
+    let mut enqueued = 0usize;
+    for idx in indexes {
+        if idx.index_type == "LSI" {
+            continue; // LSIs are always synchronous.
+        }
+        let delay = effective_delay(idx, system_default_delay);
+        if delay == 0 {
+            continue; // Synchronous GSI — handled in-txn by sync_indexes.
+        }
+        let context = GsiApplyContext {
+            base_key_schema: key_info.key_schema.clone(),
+            attribute_definitions: key_info.attribute_definitions.clone(),
+            index: GsiIndexDef {
+                index_id: idx.index_id.clone(),
+                key_schema: idx.key_schema.clone(),
+                projection: idx.projection.clone(),
+            },
+        };
+        enqueue_gsi_pending(tx, &key_info.table_id, old_item, new_item, delay, &context).await?;
+        enqueued += 1;
+    }
+    Ok(enqueued)
+}
+
+/// Insert one self-describing `gsi_pending` row for a single index inside the
+/// base write transaction (zero crash window).
+///
+/// `delay_ms` is the index's effective delay; a jitter in `[delay/2, delay]`
+/// is applied. `ready_at` is clamped to `max(now + jitter, MAX(ready_at) in the
+/// base key's partition)` so a later write that draws a smaller jitter can never
+/// become eligible before an earlier one — preserving per-key FIFO when the
+/// worker drains the partition in `id` order.
+pub(crate) async fn enqueue_gsi_pending(
+    tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>,
+    table_id: &str,
+    old_item: Option<&Item>,
+    new_item: Option<&Item>,
+    delay_ms: u64,
+    context: &GsiApplyContext,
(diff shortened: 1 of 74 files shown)


crates/bin/src/backends.rs
```diff
@@ -0,0 +1,44 @@
...
+//! 2. implement the storage traits and `inventory::submit!` registrations and
+//!    expose a `pub fn link()` in the backend crate,
+//! 3. add a feature-gated `link()` call below.
+
+// Developer mode relaxes the security posture (plain HTTP on loopback, open
+// authorization). It is a dev/CI-only profile and must never be compiled into a
+// binary that also contains a production backend. Postgres is the production
+// backend, so the combination is rejected at compile time — there is no runtime
+// path by which a Postgres deployment can serve in dev mode.
+#[cfg(all(feature = "dev-mode", feature = "postgres"))]
```
jcshepherd: What would you think of making this require sqlite or sqlite-memory with dev mode, rather than prevent postgres and dev mode specifically? Most backends we add will be considered production backends, I think, but very few will be considered as suitable for dev only.

crates/bin/src/cmd_serve.rs
```diff
@@ -44,6 +44,76 @@ pub struct ServeArgs {
 /// Bind the listening socket, daemonize, then start the tokio runtime.
```
jcshepherd: This comment looks detached from the function it was associated with (run()).

crates/bin/src/cmd_serve.rs
```diff
@@ -44,6 +44,76 @@ pub struct ServeArgs {
...
+        .ok()
+        .filter(|s| !s.is_empty())
+        .unwrap_or_else(|| DEFAULT_ACCESS_KEY_ID.to_owned());
+    let secret = std::env::var("AWS_SECRET_ACCESS_KEY")
+        .ok()
+        .filter(|s| !s.is_empty())
+        .unwrap_or_else(|| DEFAULT_SECRET.to_owned());
+
+    // The credential store dispatches on the access-key prefix; keep that
+    // invariant intact rather than special-casing it for dev mode.
```
jcshepherd: I don't understand this. I mean, I understand the words but I don't understand why we need to do anything special for dev mode.

jcshepherd (commented):
Part 1 of N. Few minor comments/questions around dev mode, etc.

crates/bin/src/backends.rs
LeeroyHannigan: Done

crates/bin/src/cmd_serve.rs
LeeroyHannigan: Fixed

crates/bin/src/cmd_serve.rs
LeeroyHannigan: Dev mode only relaxes authorization (the IAM policy decision) SigV4 authentication stays identical to production, so the server must hold a credential whose secret matches what the SDK signs with, or every request fails signature verification. `seed_dev_credential` supplies that: it adopts `AWS_ACCESS_KEY_ID`/`SECRET` if present, else seeds AWS's well-known example key, keeping us a drop-in for DynamoDB Local without forking (and under-testing) the auth path.

jcshepherd: That seems orthogonal to the key type. What bad thing happens if someone uses an ASIA-type key in dev mode?

crates/bin/src/cmd_serve.rs
```diff
@@ -41,6 +41,76 @@ pub struct ServeArgs {
...
+    if !access_key_id.starts_with("AKIA") {
+        anyhow::bail!(
+            "dev credential AWS_ACCESS_KEY_ID must be AKIA-shaped (got '{access_key_id}'); \
+             use e.g. AKIAIOSFODNN7EXAMPLE, or unset it to use the default."
+        );
+    }
+
+    // The default account is created during bootstrap (by `init`, or by
```
jcshepherd: Do we have an explicit concept of a default account or admin account? If not, what is the assurance that the first account returned by list_all_accounts() is a/the admin account? Or is it okay to create the dev user under any account?

crates/bin/src/cmd_serve.rs
```diff
@@ -430,7 +547,7 @@ async fn serve_inner(
     // P120e: Create metrics collector early so workers can record health.
     let metrics = Arc::new(extenddb_core::metrics::MetricsCollector::new());
 
-    let tls_enabled = app_config.server.tls.enabled;
+    let tls_enabled = app_config.server.tls.enabled && !cfg!(feature = "dev-mode");
```
jcshepherd: Minor: I'd feel a shade more comfortable if there was a nice is_tls_enabled() utility function that wrapped this && . I haven't made it through the rest of the review but I'm seeing several checks in manage_http.rs on tls.enabled that don't appear to be updated.

crates/storage-sqlite/src/data/index.rs
```diff
@@ -0,0 +1,488 @@
...
+    // ready_at as RFC 3339 so it compares correctly (lexically) against the
+    // worker's RFC 3339 `now` cutoff and against other rows' ready_at.
+    let jittered = jitter_delay_ms(delay_ms);
+    let candidate = crate::sqlite_util::format_timestamp(
+        time::OffsetDateTime::now_utc()
+            + time::Duration::milliseconds(i64::try_from(jittered).unwrap_or(i64::MAX)),
+    );
+    // Monotonic clamp within the partition (RFC 3339 strings sort lexically).
+    let part_max: Option<String> =
+        sqlx::query_scalar("SELECT MAX(ready_at) FROM gsi_pending WHERE worker_partition = ?")
```
jcshepherd, line 214: Is ready_at indexed?

crates/bin/src/cmd_serve.rs
LeeroyHannigan: Fixed. There's now an explicit default-account concept: bootstrap records default_account_id in settings (exposed via CatalogStore::default_account_id()), and seed_dev_credential resolves the account through it rather than list_all_accounts().first(). Recorded on both bootstrap paths (file-backed init + in-memory serve-time), idempotent so it also backfills existing catalogs.

crates/bin/src/cmd_serve.rs
LeeroyHannigan: Fixed

crates/storage-sqlite/src/data/tx_helpers.rs
```diff
@@ -0,0 +1,315 @@
...
+            .map_err(|e| StorageError::Internal(e.to_string()))?;
+    }
+    Ok(())
+}
+
+/// Allocate the next monotonic stream sequence number within a transaction.
+async fn next_stream_seq(
+    tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>,
+) -> Result<i64, StorageError> {
+    sqlx::query("UPDATE seq_counters SET value = value + 1 WHERE name = 'stream'")
```
jcshepherd, line 154: Does this imply a global ordering for all stream records across all shards? Is that what we want?

jcshepherd (commented):
Left a few more comments/questions.

General questions: are there "migration scripts" to go with this PR, for setting up system tables, etc., needed by ExtendDB in MySQL?

crates/storage-sqlite/src/data/index.rs
LeeroyHannigan: Yes `idx_gsi_pending_claim` ON `gsi_pending` (`worker_partition`, `ready_at`, `id`), which matches the drain query exactly (filter partition, order by `ready_at`, tiebreak id)

crates/storage-sqlite/src/data/tx_helpers.rs
LeeroyHannigan: The counter is global/monotonic, but records are stored and consumed per shard (stream_records PK (`shard_id`, `sequence_number`)), so consumers only ever see per-shard order, which is all DynamoDB guarantees. Global monotonicity is just a cheap superset on single-writer SQLite; it doesn't imply any cross-shard ordering contract.

LeeroyHannigan:
> Left a few more comments/questions.
> 
> General questions: are there "migration scripts" to go with this PR, for setting up system tables, etc., needed by ExtendDB in MySQL?

I assume you mean SQLLite? Schema setup is one authoritative migration, `crate::schema::apply` (mirrors Postgres 001_schema.sql), applied at extenddb init (file-backed) and at serve-time bootstrap (in-memory), with extenddb migrate for existing catalogs.

eldondevat:
Just want to say I pulled this down to build and started testing with it. _Very_ excited here. 😋

LeeroyHannigan:
Thanks @eldondevat 🚀

We'll get things finalized this week and hopefully get it into your hands very soon 💪🏻

crates/storage-sqlite/src/update_table.rs
```diff
@@ -0,0 +1,373 @@
...
+                .await?;
+            }
+        }
+
+        tx.commit()
+            .await
+            .map_err(|e| StorageError::Internal(e.to_string()))?;
+
+        // Data DDL after catalog commit.
+        if let Some(updates) = &input.global_secondary_index_updates {
```
yesyayen, line 301: couple of things I'd like to discuss regarding GSI creation and backfill process,
1. [HIGH] Backfill runs under the global write lock, so it blocks writes to **all** base tables in the instance for the full duration
2. GSI is advertised ACTIVE (after Catalog table COMMIT) before base table data is backfilled, since these are non-atomic multi step update.  (DescribeTable respond with ACTIVE for the index, but Get or Query will result in resource not found since data table does not yet exist)
3. Create + backfill is non-atomic, and a process crash between them leaves an ACTIVE index row with no data table, and nothing reconciles it on restart

crates/storage-sqlite/src/data/put_item.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub(crate) async fn put_item_impl(
+        &self,
+        key_info: &TableKeyInfo,
+        item: Item,
+        return_old: bool,
+        condition: Option<&Expr>,
+        maps: &ExpressionMaps,
+        stream: Option<&StreamCapture>,
+    ) -> Result<Option<Item>, StorageError> {
+        let indexes = fetch_indexes_for_table(&key_info.table_id, &self.pool).await?;
```
yesyayen, line 37: [not a blocker] 
table's index metadata is fetched from the catalog on the write hot-path for every request (put, update, delete, and TWI). I believe we should cache this on the already cached `TableKeyInfo` and skip the read. Same pattern exists in the Postgres backend, and can be resolved by making this change on the shared layer.

crates/storage-sqlite/src/workers.rs
```diff
@@ -0,0 +1,471 @@
...
+    // backends; the worker is otherwise notify-driven and sleeps to MIN(ready_at).
+    const MAX_SLEEP: Duration = Duration::from_secs(1);
+    // Backoff after an error so a poison row cannot hot-loop the worker.
+    const ERROR_BACKOFF: Duration = Duration::from_secs(1);
+    loop {
+        // Drain everything currently due.
+        let mut errored = false;
+        loop {
+            match process_gsi_batch(&engine).await {
+                Ok(n) if n > 0 => continue,
```
yesyayen, line 370: [not a blocker] GSI async propagation, divergent failure modes across backends.

Postgres and SQLite handle async GSI propagation differently on failure. Postgres runs 4 workers partitioned by pk-hash and, on a bad update, logs-and-drops it so the queue keeps draining (liveness preserved, that one index write silently lost).

SQLite uses a single worker across all tables/accounts and, on a bad row, rolls back the claimed batch and re-claims the same rows forever (no drop). So one poison row permanently stalls all GSI propagation instance-wide and the queue grows unbounded, not just head-of-line delay. 

Neither of these matches DynamoDB; flagging the divergence as a discussion point.

crates/storage-sqlite/src/schema.rs
```diff
@@ -0,0 +1,402 @@
...
+    policy_document TEXT NOT NULL,
+    PRIMARY KEY (account_id, principal_type, principal_name)
+);
+
+-- Idempotency tokens for TransactWriteItems.
+CREATE TABLE IF NOT EXISTS idempotency_tokens (
+    token       TEXT PRIMARY KEY,
+    fingerprint TEXT NOT NULL,
+    created_at  TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ','now'))
+);
```
yesyayen, line 265: this table is keyed on `token` alone, but TWI idempotency tokens are unique per account, not globally. Key it on `(account_id, token)` so two accounts reusing the same token don't clash (currently they do: silent replay or false IdempotentParameterMismatch across accounts).

crates/storage-sqlite/src/data/transactions.rs
```diff
@@ -0,0 +1,464 @@
...
+        let mut table_indexes: HashMap<String, Vec<IndexMeta>> = HashMap::new();
+        for op in ops {
+            let name = op_table_name(op);
+            if !table_indexes.contains_key(name) {
+                let indexes = fetch_indexes_for_table(op_table_id(op), &self.pool).await?;
+                table_indexes.insert(name.to_owned(), indexes);
+            }
+        }
+
+        let _writer = self.write_lock.lock().await;
```
yesyayen: there is a TOCTOU race condition here, index metadata is read before taking write_lock, so a GSI added by a concurrent UpdateTable in this gap can be missed, and this write won't maintain it, leaving those items permanently missing from the new GSI. I believe we should read the index set after acquiring the lock.

this also exists on other write APIs.

LeeroyHannigan:
Thanks @yesyayen — really appreciate the thorough pass. All five are addressed; mapping each to the change:

**1. GSI create/backfill (`update_table.rs`) [HIGH]** — split into the three sub-issues you raised:
- *ACTIVE advertised before data exists* — fixed. The index is now inserted as `CREATING` and flipped to `ACTIVE` only after the data table is created and backfilled, so `DescribeTable` reports `CREATING` for the whole build window and a `Query`/`GetItem` never hits a not-yet-populated index.
- *Non-atomic create+backfill (crash leaves an orphan)* — fixed. Added a startup reconciler (`reconcile_incomplete_gsis`) that finds any GSI left in `CREATING`, drops any partial data table, rebuilds + backfills it, then flips it `ACTIVE`. Idempotent.
- *Backfill runs under the global write lock* — acknowledged and **deliberately deferred** to a follow-up. Making it non-blocking (chunked backfill + online catch-up) is a real redesign against SQLite's single-writer model; I didn't want to bundle that into this backend-landing PR. Called out explicitly in the commit message.

**2. TOCTOU on the index read (`transactions.rs` + others)** — fixed. The index-metadata read is now taken *after* acquiring `write_lock` on all four write paths (Put/Update/Delete/TransactWriteItems), so a GSI added by a concurrent `UpdateTable` (which holds the same lock) can't be missed.

**3. Idempotency token keyed globally (`schema.rs`)** — fixed. `idempotency_tokens` is now keyed on `(account_id, token)`, and the account is threaded through the check. Note: the PostgreSQL backend has the identical bug — tracked separately, out of scope for this SQLite PR.

**4. GSI worker poison-row stall (`workers.rs`) [not a blocker]** — fixed anyway, since it's a real liveness bug. Each pending row is now applied under a `SAVEPOINT`; an unprocessable row is logged-and-dropped (matching the PostgreSQL backend) instead of rolling back the batch and re-claiming it forever.

**5. Hot-path metadata read (`put_item.rs`) [not a blocker]** — agreed it belongs in the shared layer (same pattern in Postgres), so I've left it as a follow-up rather than a SQLite-local patch here.

Also added a data-plane parity fix: ops on a `CREATING`/`DELETING` table now return `ResourceNotFoundException` (was `ResourceInUseException`), matching DynamoDB and the Postgres backend.

Tests: idempotency (replay/mismatch/cross-account), poison-row drop, and reconciler crash-recovery added as in-crate tests. Full verification on the current tip — Python comprehensive suite 1106 passed / 0 failed, Rust integration 372 passed / 0 failed, unit 16 passed, fmt + clippy `-D warnings` clean.

The branch was also linearized (it carried internal merge commits that the merge queue rejects) and rebased onto current `main`.

crates/bin/src/cmd_serve.rs
jcshepherd: Neato.

crates/storage-sqlite/src/data/tx_helpers.rs
jcshepherd: Not blocking, but it wouldn't surprise me if some day some clever customer notices that there does seem to be some ordering across shards (if I'm understanding this correctly) and takes a dependency on it, not realizing that other backends may behave differently. (Having said that, I haven't confirmed that other backends _do_ behave differently ...)

jcshepherd (approved):
I think I left a minor comment below about stream seqnums, which is non-blocking, but otherwise I don't have strong reasons not to approve at this point. It'd be best right now to get it out and iterate on smaller improvements if needed.

PR #183, opened 2026-06-29
Fix/number validation and txn key errors

LeeroyHannigan:
## What

Three parity fixes to error reporting in number-value validation and transactional key validation:

1. **`N` attribute validation** (`crates/core/src/validation/number.rs`)
   - No longer trims input: leading/trailing whitespace (`" 5"`, `"5 "`) is
     now rejected, matching DynamoDB (previously silently accepted).
   - A bare `"."` (and `"+."`, `".e5"`) is now rejected instead of being
     accepted as `0`.
   - Malformed-but-non-empty numbers now return
     `The parameter cannot be converted to a numeric value: <input>` instead of
     reusing the empty-value message ("Supplied AttributeValue is empty, ...").
     The empty-value message is retained only for a genuinely empty string.

2. **Key-only transaction validation** (`crates/storage-postgres/src/data/transactions.rs`)
   - `TransactWriteItems` Delete/Update/ConditionCheck now validate their `Key`
     via `validate_batch_key_only`, so a key type mismatch surfaces as
     `The provided key element does not match the schema` (the batch/transaction
     form) instead of the single-item "Type mismatch for key attribute ..."
     string. The Put item-key path is intentionally left on `validate_item_keys`.

3. **Up-front empty/oversize key validation** (`crates/engine/src/transact_write_items.rs`)
   - Delete/Update/ConditionCheck now run `validate_key_sizes` during the
     prepare pass (as Put already did), so an empty or oversize key is reported
     as a top-level `ValidationException` rather than a per-item cancellation
     reason — matching DynamoDB's up-front input validation.

## Why

ExtendDB diverged from real DynamoDB on the error type/message/placement for
several invalid-input cases. Emulator consumers rely on these being faithful
(error name, exact message, and whether a fault is top-level vs a transaction
cancellation reason), so the divergences cause incorrect client behaviour in
code paths that branch on DynamoDB errors.

Closes # 

## Testing done

- Added Rust unit tests in `number.rs` covering whitespace rejection, bare-dot
  rejection, the malformed-vs-empty message split, and the empty-value message.
- `cargo test --workspace` — 562 passed, 0 failed.
- Manually validated against a local ExtendDB instance backed by PostgreSQL,
  exercising the affected operations and confirming the exact error name,
  message text, and top-level-vs-cancellation placement now match real DynamoDB
  (`us-east-1`):
  - `PutItem` with malformed `N` values (whitespace, bare-dot, `+e2`, `1+2`,
    `NaN`, etc.).
  - `TransactWriteItems` Update/Delete/ConditionCheck with wrong-typed,
    non-scalar, and empty-string keys.

## Checklist

- [x] I have updated documentation if behavior changed (n/a — no documented
      behaviour changed; error strings now match DynamoDB)
- [x] Breaking changes are noted below (none)

ADR / RFC: n/a — error-message and input-validation parity fixes only; no
change to the wire protocol, `Storage` trait, auth model, on-disk format, or
CLI surface.

## Breaking changes

None. Error names and HTTP status codes are unchanged; only error *message
text* and the top-level-vs-cancellation placement of certain key-validation
failures change, bringing them in line with real DynamoDB.

---

### DIFF
--- a/crates/core/src/validation/number.rs
+++ b/crates/core/src/validation/number.rs
@@ -9,7 +9,10 @@ const MIN_EXPONENT: i32 = -130;
 
 /// Validate and normalize a `DynamoDB` number string.
 pub fn validate_and_normalize_number(s: &str) -> Result<String, DynamoDbError> {
-    let s = s.trim();
+    // DynamoDB does not trim: leading or trailing whitespace makes the number
+    // malformed rather than acceptable (e.g. " 5" and "5 " are rejected). A
+    // genuinely empty string is reported with a distinct message from a
+    // malformed-but-non-empty one.
     if s.is_empty() {
         return Err(number_err());
     }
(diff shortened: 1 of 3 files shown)


pdf-amzn (commented):
Do we want some/all of these cases tested in an integration test (either in the Rust or the Java suite) not just as a Rust build-time test, so we get to compare DynamoDB and ExtendDB behavior? (I suspect the Rust integration tests have some quirks still, we should fix those and add them to our GitHub workflows.)

Issue #185, opened 2026-06-29
[RFC] Cloud Bigtable Storage Backend


annguy3n:
## Summary

Defines the design and implementation of a production-grade Cloud Bigtable storage backend for ExtendDB. It specifies the mapping of the DynamoDB data model, query engine, transaction controls, and streams onto Google Cloud Bigtable. The design focuses on ensuring that workloads utilizing the DynamoDB API can run on GCP with high fidelity, guaranteeing transaction isolation, serializability, and scalable Time-to-Live (TTL) eviction.

## Motivation

ExtendDB's goal is to allow developers to use the DynamoDB programming model in environments where the managed DynamoDB service is unavailable. Implementing a production-ready Cloud Bigtable connector supports this mission by:

1. Enabling Multi-Cloud Portability
2. Supporting Hybrid Deployments
3. Ensuring API Fidelity

## Proposed design

Describe the design in enough detail that someone familiar with ExtendDB could implement it. Include:

* API surface (new operations, changed behavior, wire format)
  * Implements the full mandatory subset of the StorageEngine trait.
  * Enforces strict GSI projection checks: queries requesting non-projected attributes on a GSI return 
  * ValidationException (matching DynamoDB behavior) instead of performing slow base-table fetches.
* Storage implications (schema changes, new tables, migration path)
  * Catalog Metadata: System metadata, accounts, and policies are stored in a single table named `__extenddb_catalog__`.
  * Data Tables: DynamoDB tables map to Bigtable tables named `t<table_id_hex>` (where `table_id_hex` is a 32-character hex UUID). Data attributes are stored in column family `d`.
  * GSIs: Shadow tables are named `t<table_id_hex>_g<idx_hash>` (where `idx_hash` is an 8-character hash of the index name) to respect Bigtable's 50-character limit.
  * Transaction Intents: Column family `m` in data tables stores locks.
  * TTL Index Table: Introduces a dedicated, sharded shadow table `__extenddb_ttl_index__` with key encoding `[shard_id:1] [expiry_timestamp_be:8] [account_id_len:1] [account_id] [table_name_len:1] [table_name] [encoded_base_row_key]`. This allows targeted range scans for expired items rather than scanning entire data tables.
* Configuration (new `extenddb.toml` sections or flags)
  * Adds the following [storage.bigtable] block to extenddb.toml:
```
[storage]
backend = "bigtable"

[storage.bigtable]
project_id = "my-project"
instance_id = "my-instance"
[...]

PR #186, opened 2026-06-29
docs: add RFC draft for Cloud Bigtable storage backend

annguy3n:
## What

This PR introduces the draft design for the Cloud Bigtable storage backend (RFC-0003).

Key design highlights include:

* Lock-then-Read 2PC Flow
* Single-Row Write Guarding
* Transactional Streams
* Sharded TTL Index shadow table
* GSI Validation and Reconciler
* Secure Credentials Path

## Why

Cloud Bigtable is GCP's managed wide-column store and serves as the natural target for DynamoDB workloads migrating to GCP. A robust connector allows AWS-to-GCP migration for high-throughput, low-latency applications with zero code changes.

## Checklist

ADR / RFC: https://github.com/ExtendDB/extenddb/issues/185

---

### DIFF
--- a/docs/rfcs/0000-bigtable-backend.md
+++ b/docs/rfcs/0000-bigtable-backend.md
@@ -7,17 +7,18 @@
 
 ## Summary
 
-This RFC defines the design and implementation of a production-grade Cloud Bigtable storage backend for ExtendDB. It addresses critical architectural requirements for this new backend: transaction isolation, stream emissions in transactions, Global Secondary Index (GSI) projections, GSI write consistency, Time-to-Live (TTL) eviction scalability, and secure Google Cloud Platform (GCP) connectivity.
+This RFC defines the design and implementation of a production-grade Cloud Bigtable storage backend for ExtendDB. It addresses critical architectural requirements for this backend: transaction isolation, stream emissions in transactions, Global Secondary Index (GSI) projections, GSI write consistency, Time-to-Live (TTL) eviction scalability, Optimistic Concurrency Control (OCC) row versioning, fenced two-phase commit (2PC) apply, application profile single-cluster routing, and secure Google Cloud Platform (GCP) connectivity.
 
 ## Motivation
 
 A naive mapping of DynamoDB structures to Bigtable introduces several architectural challenges that must be addressed to ensure production-readiness:
-1.  **Transaction Isolation Leak:** Non-transactional (single-row) writes bypass 2-Phase Commit (2PC) locks (intents), enabling dirty writes and lost updates.
-2.  **2PC Race Condition (Read-then-Lock):** The 2PC coordinator reads row data before placing locks, exposing the system to Time-of-Check to Time-of-Use (TOCTOU) races.
+1.  **Transaction Isolation Leak & TOCTOU Races:** Non-transactional (single-row) conditional writes read an item and evaluate conditions without atomic concurrency control, leading to Time-of-Check to Time-of-Use (TOCTOU) races where concurrent writes can overwrite uncommitted state or bypass 2-Phase Commit (2PC) locks.
+2.  **2PC Race Condition (Read-then-Lock & Unfenced Apply):** Reading row data before placing locks exposes 2PC to TOCTOU races. Furthermore, applying 2PC Phase 2 mutations unconditionally without fencing against expired/aborted locks allows slow coordinators to overwrite newer state.
 3.  **Missing Transaction Streams:** Multi-row transactions (`TransactWriteItems`) do not emit DynamoDB Stream records.
 4.  **Inefficient TTL Eviction:** The TTL worker performs full-table scans to identify expired items, which degrades performance on large datasets.
 5.  **GSI Consistency Gap:** GSI shadow table write failures are logged without retry or repair, causing permanent base-to-shadow divergence.
-6.  **Auth Configuration Constraints:** Lack of support for explicit credentials file configuration.
+6.  **Multi-Cluster Replication & Conditional Mutations:** Cloud Bigtable rejects conditional mutations (`CheckAndMutateRow`) under multi-cluster routing with eventual consistency.
+7.  **Auth Configuration Constraints:** Lack of support for explicit credentials file configuration.
 
 This RFC provides the technical specification to resolve these challenges.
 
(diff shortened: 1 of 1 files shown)


LeeroyHannigan:
Thank you for this RFC. At a quick glance, this looks solid. Let me do a thorough review and circle back to you early next week (week commencing 6 July).

LeeroyHannigan:
Thanks for the detailed RFC, @annguy3n this is a strong design and we're happy to accept a Cloud Bigtable backend on these lines. You've clearly thought hard about the DynamoDB semantics: the lock-then-read 2PC, single-row write guarding, stream replay via the txn log, and the sharded TTL index are exactly the right calls given Bigtable's single-row atomicity.

A few DynamoDB-parity points to fold in, split into what must match vs. what's fine to differ if documented:

### Must match DynamoDB

- TTL + Streams: the TTL sweeper's deletes must emit DynamoDB-compatible `REMOVE` stream records (with the TTL principal) when streams are enabled, please make that explicit in the design, since silently dropping them is a breaking behavior gap.

- Key encoding: DynamoDB sorts Number keys numerically (exact to 38 significant digits) and String/Binary keys by byte order. In a byte-lexicographic store that means numeric keys need an order-preserving byte encoding, and numbers must round-trip as exact decimals (not floats). This currently only shows up under Testing but it's a core design detail and should be promoted into the row-key format and GSI index-key sections.
    
- Transaction rules: `TransactWriteItems` needs `ClientRequestToken` idempotency (10-minute window) and must reject two operations on the same item, worth noting alongside the 2PC flow.

- Read isolation: DynamoDB gives read-committed for GetItem/Query/Scan/BatchGetItem and serializable for TransactGetItems. Two things fall out of the 2PC apply window: (1) a row's new value must not be visible until the transaction's commit point, applying mutations before the commit decision is durable would expose a write that can still roll back, which is a dirty read and breaks read-committed; and (2) TransactGetItems must observe the committed set atomically (read through the coordinator / at the commit boundary), otherwise it can see partial state and violate serializable. Please spell out how both are guaranteed.

### Fine to differ, just call it out

- GSI consistency: shadow tables make GSIs eventually consistent, which matches DynamoDB, good. Please still reject `ConsistentRead=true` on a GSI (API parity), and make sure the reconciler keeps GSI reads from returning orphaned/phantom rows during the divergence window.

- Cross-row visibility for standard reads: during the apply window a Query/Scan may see some rows of a committed transaction updated and others not. That's within read-committed (each row still returns a committed value), so it's fine, just document it, and confirm the read-committed and TransactGetItems guarantees above hold.

- Backup / PITR / export: not covered here, fine to defer to a follow-up RFC, but let's note that the effective PITR/export window will differ from DynamoDB's 35 days.

Overall: accepted in principle, let's tighten the TTL-stream path, key-encoding, and the read-isolation guarantees in the doc, then move to implementation behind the opt-in feature. Really nice write-up.

annguy3n:
@LeeroyHannigan Thanks for the feedback, I've updated the RFC to address them. Let me know if you have any other questions/comments.

LeeroyHannigan:
Hi @annguy3n , 

Thanks for the update to the RFC, please find below my findings

## Blocking

**B1. Internal phase-numbering is self-contradictory.** The 2PC mermaid in 2.B defines exactly three labeled phases (Phase 1 Open, Phase 2 Commit, Phase 3 Clean up), with intent placement and data apply as unlabeled steps. But the prose references non-existent/contradictory phases: 2.C says intents are written in "Phase 3" (the diagram's Phase 3 is *Clean up*) and data is applied "only in Phase 7"; 3 says log enrichment in "Phase 2" and stream records written in "Phase 5"; the recovery sweeper (2.E) and streams (3.4) roll forward against these. There is no Phase 4–7 anywhere. This must be reconciled to a single consistent phase list before implementation, because the commit/visibility ordering (see B2) hinges on which phase is the durable commit point.

**B2. `TransactGetItems` serializable guarantee is asserted but the mechanism is missing.** 2.C claims TransactGetItems "coordinates with `__extenddb_txn_log__` to observe the snapshot at the transaction commit boundary." No snapshot/MVCC mechanism is specified. The design keeps a single committed value per row in family `d` (mutations applied post-commit), and Bigtable cell timestamps are per-row, not a global read timestamp, so there is no described way to obtain a cross-row atomic snapshot at a commit boundary. As written this reads as read-committed, not serializable. Either specify the mechanism (e.g. read-through-log to reconstruct pre/post images, or a commit-timestamp read) or downgrade the claimed guarantee. 

## Non-blocking

**N1. Commit-point vs client-ack visibility gap (needs to be pinned down).** 2.C states data in `d` is applied only after `COMMITTED` is durable, and the sweeper rolls forward on crash. It never states *when success is returned to the client*. If the coordinator acks at the durable-COMMITTED point but applies to `d` afterward, a strongly-consistent `GetItem`/`Query` issued between ack and apply returns the pre-transaction (still-committed) value. DynamoDB reflects a succeeded `TransactWriteItems` in a subsequent strongly-consistent read. Read-committed per-row is preserved, but read-your-writes after a 200 is not, unless the client ack is deferred until apply completes. State the ack point explicitly.

**N2. GSI row-key delimiter contradicts the string-key encoding rule.** 1.B first says String/Binary keys use "null-byte / length-prefixed terminators to preserve lexicographical sort order," then defines the GSI shadow row key as `[pk_val]\0[sk_val]\0[base_pk_val]\0[base_sk_val]` — bare `\0` separators. DynamoDB String/Binary values can contain `0x00` bytes; bare-`\0` delimiting breaks both uniqueness and sort parity when a value embeds a null, and is inconsistent with the length-prefixed scheme stated one paragraph earlier. Use the same length-prefixed/order-preserving composite encoding for the GSI key.

**N3. Trait name is stale vs `main`.** 5.B says the reconciler is "spawned via `RuntimeHooks`." The actual trait is `ServerRuntimeHooks` with `async fn spawn_workers(&self, ctx: &WorkerContext)` at `crates/storage/src/hooks.rs:47,58`; postgres/sqlite/mongodb all implement it (`crates/storage-postgres/src/lib.rs:376`, `crates/storage-sqlite/src/hooks.rs`, `crates/storage-mongodb/src/lib.rs:50`). Align the name so the RFC points at the real extension surface (TTL/GSI/recovery workers should all register through `spawn_workers`).

## Notes (positive / low-risk)

- Config shape is consistent with repo convention: `[storage] backend = "..."` selects the backend and backend tables are `[storage.<name>]` (`crates/config/src/lib.rs:186`), so `[storage.bigtable]` with `pool_size = 20` matches the existing postgres default. Good.
- The design is docs-only and adds no code; it does not touch or break any current subsystem (backends on main: postgres, sqlite, mongodb; no bigtable dir).
- Core-level operations the RFC leans on already exist, so the Bigtable work is a backend implementation of established contracts, not new surface: `TransactWriteItemsInput`/`ClientRequestToken` (`crates/core/src/types/transaction.rs:83-89`), `IdempotentParameterMismatch` (`crates/core/src/error/mod.rs:37`), `transact_write_helpers.rs`.

Thanks again, looking forward to seeing your implementation of Bigtable here.

annguy3n:
@LeeroyHannigan, would you mind to take another pass at it?

LeeroyHannigan:
Hi @annguy3n,

Third pass done. B1 (phase numbering), N2 (bare-`\0` delimiters), and N3 (`ServerRuntimeHooks::spawn_workers`) are all cleanly resolved, and the new material is strong - the fenced apply, OCC versioning, GC rules, and the app-profile constraint are all accurate against Bigtable's actual behavior. One new blocking item, unfortunately introduced by the N2 fix itself, plus two carry-overs.

## Blocking

**B3. The GSI composite key has a non-terminal raw sort-key segment (1.B, GSI Index Row Keys).** `gsi_sk_bytes` is raw and unterminated, but the base key follows it, so the byte after a short sort key is `base_pk_tag` (0x53), which participates in comparisons. Two proven consequences (simulated the byte layout directly): (1) sort inversion - for the same `gsi_pk`, sort key `"A"` orders *after* `"AB"` because 0x53 > 0x42, breaking GSI Query ordering, `begins_with`, and `between`; and (2) the segment boundary is undecodable - I constructed two distinct `(gsi_sk, base_pk)` pairs with byte-identical row keys, which breaks uniqueness and the reconciler's parsing. Raw-to-end-of-key is only safe for the *terminal* segment (the base composite key gets this right). Suggested fix: escape-encode the non-terminal `gsi_sk` segment order-preservingly (`0x00` -> `0x00 0xFF`, terminate with `0x00 0x00` - the FDB tuple pattern); `base_sk` stays raw since it is terminal. This is the same inversion class your Alternatives section rejects for length prefixes, one level down.

## Should-fix

**B2 (carry-over, much improved).** The mutation payloads stored in the coordinator row at `COMMITTED` (section 3) now make the serializable `TransactGetItems` claim mechanically achievable - but 2.C still says "coordinates with the txn log to observe the snapshot" without stating the rule. One paragraph closes it: intent present + txn `COMMITTED` -> serve the logged post-image; intent present + txn `PENDING` -> serve the current `d` value (apply cannot have started).

**N1 (carry-over).** The success ack point is still unstated - failure paths return `TransactionConflict`/`Canceled`, but nothing says whether the client's 200 comes after `COMMITTED` is durable or after fenced apply completes. Read-your-writes on a subsequent strongly-consistent read hinges on it; one sentence in 2.B or 2.C.

## Nit

The OCC existing-row predicate is written as `(no active intent) AND (m:v == v)` with `true_mutations`; Bigtable filters have no NOT, so that literal form isn't expressible. The equivalent De Morgan form (intent-exists OR `m:v != v` as an interleave, mutations in `false_mutations`) is - and your insert case already uses that trick. Worth stating so the implementer doesn't hit the dead end.

Close on B3 and the two paragraphs and this is ready to land as an accepted RFC from my side. The design has gotten genuinely tight across the three rounds.

annguy3n:
Hi @LeeroyHannigan,

Thanks again for the thorough review! We have updated both the RFC (`docs/rfcs/0000-bigtable-backend.md`) and the reference implementation to close on all 4 items:

### 1. B3: FoundationDB Tuple Escape-Encoding for Non-Terminal GSI Sort Keys (§1.B)
* Encoded non-terminal `gsi_sk` segments using FDB tuple order-preserving escape encoding: each `0x00` is escaped as `0x00 0xFF`, and the segment is terminated with `0x00 0x00` (terminal `base_sk` remains raw bytes to end-of-key).
* Completely eliminates sort inversion (e.g. `"A"` vs `"AB"`) and ensures unambiguous, reversible segment boundaries for GSI shadow key decoders and the background reconciler.

### 2. B2: Explicit `TransactGetItems` Read-Through-Log Snapshot Rule (§2.C)
* Codified the exact snapshot resolution rule when an active intent marker (`m:intent:<txn_id>`) is encountered:
  * **Txn `COMMITTED`:** Serves the logged post-image from `__extenddb_txn_log__` coordinator mutation payloads.
  * **Txn `PENDING`:** Serves the pre-transaction value from family `d` (apply has not started).
  * **No active intent / `CLEANED`:** Serves family `d` directly.

### 3. N1: Success Acknowledgment Boundary (§2.B & §2.C)
* Explicitly specified in the text and Mermaid sequence diagram that HTTP 200 OK is returned to the client **after** all Phase 2 fenced mutations are durably applied to column family `d` and GSI shadow tables, guaranteeing monotonic Read-Your-Writes consistency.

### 4. Nit: De Morgan Form for Bigtable OCC CAS Predicates (§2.D)
* Documented that because Bigtable row filters lack a `NOT` operator, OCC CAS predicates use the equivalent De Morgan conflict form: `(active intent in m) OR (m:v != read_version)` via `RowFilter::Interleave` with mutations placed in `false_mutations`.

LeeroyHannigan:
Hi @annguy3n,

Thanks for the quick turnaround on pass three, all previous points are cleanly resolved.

Before anything else, an apology: some of the items below existed in the first draft and should have been caught in my first pass rather than trickling out over four rounds. To close this out fairly, I did a full front-to-back review of the RFC against the Bigtable documentation, and I also checked out your feat/bigtable-backend branch, built it, and ran the emulator integration suite against a live Bigtable emulator. 

The good news first: formatting is clean, clippy has zero errors, the feature-gating is exactly right (the default binary builds without the crate), all 44 unit tests pass, and 5 of 7 emulator integration tests pass. The 2PC design held up well under a hard look.

Three items block acceptance, all confirmed in the branch:

1. Cell timestamps vs MaxVersions(1). Every SetCell uses timestamp_micros: -1, so when two writers pass their filters in a narrow window, GC keeps the cell with the later wall clock, not the higher OCC version, and because timestamps are per-cell the row can end up torn (one writer's data, the other's m:v). The emulator can't surface this (single node, single clock). Deriving the timestamp from the OCC version so all cells of one write share it makes GC deterministic.

2. Single-item stream emission is unfenced. put/update/delete_item and the TTL sweeper write the base row, then streams::emit as a trailing write; a crash between them loses the stream record. Your 2PC path already solves this correctly by persisting stream_records in the coordinator row before commit, so the pattern exists in your own code, it just needs to cover the single-item paths.

3. test_transact_get_items_read_through_log_rule fails against a live emulator: TransactGetItems returns the pre-image (initial1) instead of the committed post-image. Worth a look, this one is your own test.

Handled by the rebase you'll need anyway (the branch is 84 commits behind and Cargo.lock, the workspace Cargo.toml, and extenddb.sample.toml will conflict): main now has a StorageError::Unsupported variant, which is what the BackupEngine stubs should return instead of Internal (Internal maps to HTTP 500 plus error logs; Unsupported maps to ValidationException). Please also update the RFC text to match, and keep the title at RFC-0000 until a number is assigned at acceptance, since 0003 is already referenced in shipped code.

Everything else I found (deliverables matrix, clippy warnings under -D warnings, transient-error classification, a line on why the table ledger beats change streams) is non-blocking and fine for the implementation PR. With the three items above addressed and the rebase done, we'd love to see your code in a PR. 

Thanks for your patience across the rounds, the quality of this contribution is excellent and we're really looking forward to supporting it as a backend.

PR #200, opened 2026-07-02
fix: accept a table ARN as TableName in data-plane operations

yesyayen:
## What

Resolve a table ARN supplied in place of a bare table name to its bare name, matching Amazon DynamoDB and DynamoDB Local. Normalizes the wire TableName, batch RequestItems keys, and transact sub-op table names in the request path so authorization, validation, throttling, and echoed names all see the bare name. Index, non-table, and malformed ARNs are rejected as ValidationException. Account and region are ignored (name resolves within the caller's account), matching DynamoDB Local.

## Behavior: before vs Amazon DynamoDB

Captured from Amazon DynamoDB and DynamoDB Local via the AWS CLI:

| Scenario | ExtendDB (before) | Amazon DynamoDB / DynamoDB Local |
|---|---|---|
| `GetItem`/`PutItem`/`Query` with a same-account table ARN as `TableName` | `ResourceNotFoundException` | Resolves and serves the request |
| `BatchGetItem`/`BatchWriteItem` with an ARN `RequestItems` key | `ResourceNotFoundException` | Resolves and serves the request |
| `TransactGetItems`/`TransactWriteItems` with an ARN sub-op `TableName` | `ResourceNotFoundException` | Resolves and serves the request |
| Index ARN (`.../table/T/index/i`) as `TableName` | `ResourceNotFoundException` | `ValidationException` |
| Non-table resource type (`.../stream/T`) as `TableName` | `ResourceNotFoundException` | `ValidationException` |
| Malformed ARN as `TableName` | `ResourceNotFoundException` | `ValidationException` |
| Echoed table name (`ConsumedCapacity.TableName`, batch response keys) | bare name | the supplied ARN, verbatim |
| Valid-format ARN, missing table (control) | `ResourceNotFoundException` | `ResourceNotFoundException` |
| Bare (non-ARN) name, missing table (control) | `ResourceNotFoundException` | `ResourceNotFoundException` |

- Cross-account is not implemented. A foreign-account ARN resolves within the caller's account rather than returning `AccessDeniedException`.
- The ARN region is ignored (no "Invalid AWS region" rejection).
- Control-plane operations (e.g. `DescribeTable`) do not accept an ARN as name.

## Why

Conformance gaps

Closes #

## Testing done

- New `tests/test_table_arn_as_name.py`: 16 dual-target cases 

## Checklist

ADR / RFC: 

## Breaking changes

N/A
---

### DIFF
--- a/crates/core/src/validation/table_arn.rs
+++ b/crates/core/src/validation/table_arn.rs
@@ -1,0 +1,132 @@
+// Copyright 2026 ExtendDB contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Resolve a table ARN supplied in place of a bare `TableName`.
+
+use crate::error::DynamoDbError;
+
+/// Resolve a `TableName` that may be supplied as a table ARN to its bare name.
+///
+/// A non-ARN value is returned unchanged. A table ARN
+/// (`arn:<partition>:<vendor>:<region>:<account>:table/<name>`) resolves to
+/// `<name>`; the account and region are ignored, so the name resolves within
+/// the caller's account. This matches Amazon DynamoDB and DynamoDB Local, which
+/// both accept a table ARN wherever a table name is expected. Index, non-table,
+/// or malformed ARNs are rejected as a validation error.
+///
+/// # Errors
+///
+/// Returns `ValidationException` when `name` begins with `arn:` but is not a
+/// well-formed `table/<name>` ARN.
+pub fn resolve_table_arn(name: &str) -> Result<&str, DynamoDbError> {
+    if !name.starts_with("arn:") {
+        return Ok(name);
+    }
+    // arn:<partition>:<vendor>:<region>:<account>:<resource>. The 6th field
+    // keeps everything after the 5th colon, including any slashes.
+    let resource = match name.splitn(6, ':').nth(5) {
+        Some(resource) if !resource.is_empty() => resource,
+        _ => return Err(invalid_arn_format(name)),
+    };
+    let table = match resource.split_once('/') {
+        Some(("table", table)) => table,
+        Some(_) => return Err(constraint_error(name, "Invalid resource type")),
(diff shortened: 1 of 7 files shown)


pdf-amzn:
This is an interesting one. The intent of ARNs is that they are a globally unique reference to a resource (in some contexts, at that point in time, in other contexts, over all time). See https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html for more.

With that in mind, maybe we should only accept ARNs with a partition of "extenddb" and when we initialize a extenddb instance, we generate a UUID to represent the "region"? But we'd want to reconcile that with our notion of sigv4 support and ARNs we already return to the user. I think current behavior is that the server's own region (used for ARNs, stream records, etc.) defaults to us-east-1 and is configurable via server.region in extenddb.toml, and ARNs returned do use "aws". Sigh. Seems like this needs a little more discussion.

crates/core/src/validation/table_arn.rs
```diff
@@ -0,0 +1,132 @@
...
+/// the caller's account. This matches Amazon DynamoDB and DynamoDB Local, which
+/// both accept a table ARN wherever a table name is expected. Index, non-table,
+/// or malformed ARNs are rejected as a validation error.
+///
+/// # Errors
+///
+/// Returns `ValidationException` when `name` begins with `arn:` but is not a
+/// well-formed `table/<name>` ARN.
+pub fn resolve_table_arn(name: &str) -> Result<&str, DynamoDbError> {
+    if !name.starts_with("arn:") {
```
jcshepherd, line 22: This method brings to mind "parse, don't validate". Is it possible that eventually some of the other ARN elements will be useful? If so, is it worth considering now parsing a table arn into its constituent parts? If it fails to parse correctly, it's invalid. Otherwise, callers can access the specific parts they need. I'd be okay with deferring this in order to resolve the immediate problem, but wanted to raise the suggestion.

PR #201, opened 2026-07-03
fix: data plane request validation

LeeroyHannigan:
## What

Data-plane request-validation fixes to align error classes, messages, and validation ordering with DynamoDB:

- Validate item size **before** the table-existence check in `PutItem` and `BatchWriteItem`, so a malformed request against a non-existent table returns `ValidationException` instead of `ResourceNotFoundException`.
- Return `ValidationException` (not `SerializationException`) for invalid `ExpressionAttributeValues` keys and values: validate all keys before parsing any value, and wrap value-validation errors as `"<field> contains invalid value: <message> for key <key>"`.
- Validate `ExpressionAttributeValues` number content (empty/non-numeric, >38 significant digits, overflow, underflow) and number-set duplicates, with the number error taking precedence over the more-than-one-datatype error. Report a non-true `NULL`, an unrecognized datatype, and multiple datatypes with the correct messages.
- Report an empty **binary** value on a secondary-index key as "empty binary value" (was incorrectly "empty string value").
- Route `PutItem` request deserialization through the shared error classifier so malformed-expression errors are consistently `ValidationException`.
- Make GSI/LSI secondary-index pagination tests poll for eventual consistency (they assumed synchronous index propagation), and guard the zero-delay synchronous-index test to skip when the server is not booted in that mode rather than failing spuriously.
- Add regression coverage for transactional updates that set empty string and empty binary values on non-key attributes.

## Why

These behaviors diverged from DynamoDB (wrong error class, wrong message, or wrong validation ordering). Each fix was verified against the real DynamoDB service so the error type and message match.

Closes # 

## Testing done

- `cargo test --workspace` (unit) — all pass.
- Rust integration tests — 371 pass.
- Python integration tests — 780 pass.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` clean.
- Every changed behavior was compared against the real DynamoDB service for exact error-class and message parity.

## Checklist

ADR / RFC: n/a — no change to wire protocol, `Storage` trait, auth model,
on-disk format, or public CLI surface. Error-class/message parity only.

## Breaking changes

None. Only error responses for already-invalid requests change (from the wrong error class/message to the DynamoDB-correct one).

---


tests/conftest.py
```diff
@@ -90,6 +90,34 @@ def wait_for_active(client, table_name: str, timeout: float = 120.0) -> None:
...
         time.sleep(interval)
     raise TimeoutError(f"Table {table_name} did not become ACTIVE within {timeout}s")
+
+
+def wait_for_gsi_items(paginate, expected: int, timeout: float = 15.0):
+    """Poll a paginated GSI query/scan until it yields at least ``expected`` items.
+
+    GSIs are eventually consistent: an item written to the base table is not
+    guaranteed to be visible through a secondary index immediately — ExtendDB
+    applies the configured ``gsi_propagation_delay_ms`` (like real DynamoDB), so
```
jcshepherd, line 100: I'd remove "(like real DynamoDB)" ... DynamoDB doesn't have a configurable GSI propagation delay.

tests/conftest.py
```diff
@@ -90,6 +90,34 @@ def wait_for_active(client, table_name: str, timeout: float = 120.0) -> None:
...
+    a read-back through a GSI right after the write can legitimately be short.
```
jcshepherd: short?

tests/conftest.py
LeeroyHannigan: Done in ceb769a — dropped the "(like real DynamoDB)" parenthetical; the configurable delay is ExtendDB-specific.

tests/conftest.py
LeeroyHannigan: Done in ceb769a — reworded to "can legitimately return fewer items than expected" (a paginated read-back may not have all items yet).

crates/core/src/serde_helpers.rs
```diff
@@ -75,11 +75,36 @@ where
...
+    // Pass 2: per-value check and conversion.
+    for (key, value) in raw {
         check_value(&key, &value).map_err(de::Error::custom)?;
-        let converted = convert(value).map_err(de::Error::custom)?;
+        let converted = convert(value).map_err(|e| {
+            let msg = e.to_string();
+            // Semantic value-validation errors are wrapped with the field name
+            // and the offending key (DynamoDB parity). Wire/type errors (wrong
+            // JSON shape for a datatype) pass through as-is.
+            let is_value_validation = msg.starts_with("One or more parameter values were invalid:")
```
robinnsc, line 92: This substring list and the one in `deserialize_error` (crates/engine/src/lib.rs) now carry the same set of validation-message fragments and must move in lockstep — a new message added to one but not the other silently produces the wrong error class, which is the exact bug the PutItem commit in this PR fixed (a private stale copy of the classifier drifting). A shared predicate in core, or at minimum cross-referencing comments in both places, would prevent the recurrence.

robinnsc (commented):
One note that doesn't anchor to a line: the What section predates #275 being merged into this branch and doesn't mention the UpdateItem secondary-index-key validation — arguably the biggest behavioral change in the set, turning a silently-corrupting write into a rejection. Worth adding to the description before merge since it's what the record shows.

crates/storage-postgres/src/data/update_item.rs
```diff
@@ -158,6 +158,22 @@ impl PostgresEngine {
             validation::validate_item_size(&item, self.max_item_size_bytes)
                 .map_err(|e| StorageError::Validation(e.to_string()))?;
 
+            // Secondary-index key validation on the post-update item, matching
+            // the transactional Update path: an update expression must not set
+            // an index key to a mismatched type, nor to an empty string or
+            // binary value. Validating the evaluated image rather than the
+            // expression is what covers if_not_exists and list_append, whose
+            // result is not knowable from the expression alone.
+            let idx_refs = super::transactions::index_key_refs(&indexes);
```
robinnsc, line 167: SQLite coverage question for this half of the PR: the #275 fix adds this post-update index-key validation to Postgres and fixes MongoDB's message variant, but nothing touches the SQLite update path, and the fix commit's verification names only mongodb and postgres. Does SQLite already validate the evaluated post-update image somewhere, and is `update_item_rejects_setting_an_index_key_to_an_empty_value` green against a SQLite server? If it passes, a sentence saying where SQLite's check lives closes the question; if it doesn't run there, the same silent-corruption hole remains on the third backend.

crates/core/src/types/attribute_value.rs
```diff
@@ -76,21 +76,50 @@ impl<'de> Visitor<'de> for AttributeValueVisitor {
...
+        while let Some(entry) = map.next_entry::<String, serde_json::Value>()? {
+            entries.push(entry);
+        }
+        if entries.is_empty() {
+            return Err(de::Error::custom(
+                "Supplied AttributeValue is empty, must contain exactly one of the supported datatypes",
+            ));
+        }
+
+        // DynamoDB validates the content of a number field before it reports the
```
robinnsc, line 89: Number validation now runs here for every N/NS at parse time, and the pre-existing validation layer presumably still checks numbers later (the old test comment said "rejected later by the validation layer"). If the later check is now unreachable for wire input, a note on which one is authoritative — or removing the dead layer — would keep the two from diverging on edge cases.

PR #211, opened 2026-07-13
fix: transaction key size validation - return types and values

LeeroyHannigan:
## What

Four DynamoDB request-validation parity fixes:

1. **Transaction-API primary-key size limit.** An oversized primary key (hash > 2048 bytes, range > 1024 bytes) in any transaction sub-op — `TransactGetItems` (`Get`) and `TransactWriteItems` `Put`/`Delete`/`Update`/`ConditionCheck` now cancels the transaction with a per-item `TransactionCanceledException` carrying a `ValidationError` cancellation reason for the offending item. Previously the write sub-ops raised a top-level `ValidationException` (wrong class for transactions) and `TransactGetItems` did no size check at all. An empty key value remains a top-level `ValidationException`, matching the distinct error class DynamoDB uses for that case.

2. **PutItem/DeleteItem `ReturnValues`.** A value that is a valid `ReturnValues` enum member but is not allowed for these operations (e.g. `UPDATED_OLD`) now returns `ReturnValues can only be ALL_OLD or NONE`; a non-enum value returns the generic constraint error listing the full set `[ALL_NEW, UPDATED_OLD, ALL_OLD, NONE, UPDATED_NEW]`.

3. **Query/Scan `Select` + `ProjectionExpression`.** The rejection now carries the `1 validation error detected: ` prefix for **Query**, but not for **Scan**  matching DynamoDB, which prefixes one and not the other.

4. **Query/Scan `Select=ALL_ATTRIBUTES` on a non-ALL GSI.** Now rejected with `Select type ALL_ATTRIBUTES is not supported for global secondary index <name> because its projection type is not ALL` (previously silently accepted).

## Why

ExtendDB diverged from the AWS DynamoDB service on these validation paths: some invalid inputs were accepted, and some rejections used the wrong error class or message. Each behavior and exact message was captured from the live
AWS DynamoDB service (us-east-1) before implementing, and verified to match verbatim.

Closes # 

## Testing done

Each case was captured from the live AWS DynamoDB service first, implemented to match, then verified end-to-end against a local server.

- **Unit** (`cargo test -p extenddb-core --lib`): 345 passed — new key-size, ReturnValues, and Select-projection (Query vs Scan prefix) cases.
- **Rust integration** (against a live server): transaction key-size 6/6, wording/GSI 5/5.
- **Python integration** (against a live server): transaction key-size 7/7, wording/GSI 7/7.
- Live spot-checks confirmed verbatim parity with AWS DynamoDB, e.g. `TransactWriteItems` oversized key → `TransactionCanceledException [ValidationError]`; `PutItem` `ReturnValues=UPDATED_OLD` → `ReturnValues can only be ALL_OLD or NONE`; Query `Select=COUNT` + `ProjectionExpression` → `1 validation error detected: ...` while Scan returns the same message without the prefix; `Select=ALL_ATTRIBUTES` on a `KEYS_ONLY` GSI → rejected.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Checklist

- [x] I have updated documentation if behavior changed — n/a (validation/error
      parity with DynamoDB; no user-facing docs affected)
- [x] If this changes the wire protocol, `Storage` trait, auth model, on-disk format, or public CLI surface, an RFC has been accepted or is linked below. Otherwise, an ADR captures the decision (link below). — none of these change; request-validation only.

ADR / RFC: n/a — request-validation/error-parity fixes. No change to the wire protocol, `Storage` trait, auth model, on-disk format, or CLI surface.

## Breaking changes

None to the wire protocol or response shapes for valid requests. Validation is stricter: inputs that AWS DynamoDB rejects — oversized transaction keys, disallowed `ReturnValues` on Put/Delete, and `Select=ALL_ATTRIBUTES` on a non-ALL GSI  are now rejected (some were previously accepted, or returned a different error class/message). A client relying on ExtendDB's previous lenient behavior will now receive the same error AWS DynamoDB returns.

---

### DIFF
--- a/crates/core/src/validation/mod.rs
+++ b/crates/core/src/validation/mod.rs
@@ -1026,6 +1026,7 @@ pub fn validate_select_projection(
     has_projection: bool,
     has_attributes_to_get: bool,
     has_index_name: bool,
+    is_query: bool,
 ) -> Result<(), DynamoDbError> {
     if has_projection {
         let incompatible = match select {
@@ -1035,10 +1036,16 @@ pub fn validate_select_projection(
             _ => None,
         };
         if let Some(what) = incompatible {
-            return Err(DynamoDbError::ValidationException(format!(
-                "1 validation error detected: \
-                 Cannot specify the ProjectionExpression when choosing to get {what}"
-            )));
+            // Real DynamoDB prepends "1 validation error detected: " to this
+            // rejection for Query, but NOT for Scan.
+            let body =
+                format!("Cannot specify the ProjectionExpression when choosing to get {what}");
+            let msg = if is_query {
+                format!("1 validation error detected: {body}")
+            } else {
+                body
+            };
+            return Err(DynamoDbError::ValidationException(msg));
         }
     }
     if matches!(select, Some(Select::SpecificAttributes))
@@ -2151,14 +2158,17 @@ mod tests {
             (Select::Count, "only the Count"),
         ];
         for (select, what) in cases {
-            let err = validate_select_projection(Some(select), true, false, true).unwrap_err();
+            let body =
+                format!("Cannot specify the ProjectionExpression when choosing to get {what}");
+            // Query prepends the "1 validation error detected: " prefix.
+            let q = validate_select_projection(Some(select), true, false, true, true).unwrap_err();
             assert_eq!(
-                err.to_string(),
-                format!(
-                    "1 validation error detected: \
-                     Cannot specify the ProjectionExpression when choosing to get {what}"
-                )
+                q.to_string(),
+                format!("1 validation error detected: {body}")
             );
+            // Scan does NOT prepend the prefix (matches real DynamoDB).
+            let s = validate_select_projection(Some(select), true, false, true, false).unwrap_err();
+            assert_eq!(s.to_string(), body);
         }
     }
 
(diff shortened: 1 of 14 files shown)


pdf-amzn (dismissed):

crates/core/src/validation/mod.rs
```diff
@@ -1838,45 +1965,62 @@ mod tests {
...
+            assert_eq!(s.to_string(), body);
         }
     }
 
     #[test]
     fn select_specific_attributes_requires_projection() {
         // No projection and no AttributesToGet -> rejected.
         assert!(
-            validate_select_projection(Some(Select::SpecificAttributes), false, false, false)
+            validate_select_projection(Some(Select::SpecificAttributes), false, false, false, true)
```
jcshepherd: My kingdom for some IS_QUERY = true; IS_SCAN = false constants ...

crates/engine/src/query.rs
```diff
@@ -69,6 +70,20 @@ pub async fn handle_query(
         ));
     }
 
+    // Select=ALL_ATTRIBUTES requires an ALL-projection GSI (a GSI that does not
+    // project all attributes cannot serve ALL_ATTRIBUTES). Matches real DynamoDB.
+    if matches!(input.select, Some(Select::AllAttributes))
```
jcshepherd: Any reason not to pull this up to 'validation' and use it from both query and scan?

jcshepherd (dismissed):
Couple minor comments below but nothing blocking.

crates/core/src/validation/mod.rs
LeeroyHannigan: Done in 071629e — added `IS_QUERY`/`IS_SCAN` constants in the validation module and used them at the Query/Scan call sites (and in these tests) instead of bare `true`/`false`.

crates/engine/src/query.rs
LeeroyHannigan: Done in 071629e — pulled it up into `validation::validate_all_attributes_index_support`, now called from both `query.rs` and `scan.rs` (the block was duplicated verbatim in both).

pdf-amzn (dismissed):

jcshepherd (approved):

PR #221, opened 2026-07-22
feat: adopt sqlx::migrate for PostgreSQL catalog and data migrations (ADR-0003)

yesyayen:
## What

Replace the homegrown PostgreSQL migration runner with sqlx's built-in migrator
(`sqlx::migrate!`) for both the catalog and data databases. sqlx checksums each
migration file, applies it atomically with its ledger row, and refuses to run
when an applied file changed. Deletes the `schema_history` table and the
filename-tracked runner. Bumps `CATALOG_VERSION` 0.0.2 -> 0.1.0.

Rebased onto current main and integrated with what landed since:
- The advisory migration lock is kept; sqlx now closes the apply-vs-ledger
  atomicity gap that main's `TODO(#221)` comments deferred to this PR.
- Programmatic code migrations (`003_gsi_base_key_index`) are kept: they cannot
  be static sqlx files (dynamic `_ddb_*` table names, `CREATE INDEX
  CONCURRENTLY` cannot run in a transaction). They now track in a small
  `code_migrations` ledger; the migrate `--yes` gating and pending report cover
  them.
- `init` and `migrate` refuse a pre-sqlx catalog with a `destroy` + `init`
  directive instead of failing on a non-idempotent DDL re-run.
- A stateless test pins each shipped migration's checksum, so editing an
  already-shipped (immutable) migration `.sql` fails the CI test job at PR
  time, no database required.

## Why

The old runner tracked migrations by filename with no checksum, so editing an
already-shipped migration was a silent no-op that caused schema drift (the
incident behind ADR-0003). sqlx's `_sqlx_migrations` table closes that gap.

Closes # n/a (tracked by ADR-0003)

## Testing done

- `cargo build`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`,
  `cargo test --workspace` all green. Tripwires pin migration counts (SQL and
  code), `CATALOG_VERSION`, and per-file checksums.
- Live PostgreSQL: `tests/test_cli_lifecycle.py` + `tests/test_cli_migrate_concurrency.py`
  -> 16 passed, 1 skipped. The concurrency suite's mid-barrier assertion now
  proves DDL and ledger row commit atomically (the gap the old runner pinned as
  known). Full lifecycle: init, serve (version gate), idempotent migrate
  (including the code migration), pre-sqlx refusal, checksum-tamper refusal
  ("migration 2 was previously applied but has been modified"), destroy, re-init.

## Checklist

ADR / RFC: docs/adr/0003-catalog-migration-mechanism.md (status flipped to Accepted here)

## Breaking changes

Upgrading from a pre-sqlx catalog requires `destroy` + `init`, which drops both
databases (all items wiped, not just schema). `init` and `migrate` detect a
pre-sqlx catalog and refuse with that directive. Acceptable at v0.1 with no
dependent catalogs. `CATALOG_VERSION` 0.0.2 -> 0.1.0. Documented in the upgrade
manual.

### DIFF
--- a/crates/app/src/cmd_init.rs
+++ b/crates/app/src/cmd_init.rs
@@ -322,8 +322,27 @@ async fn run_init_migrations(
         .map_err(|e| anyhow::anyhow!("{e:?}"))?;
 
     if initialized {
+        // ADR-0003: a pre-sqlx deployment cannot be adopted in place. Without
+        // this guard, init would skip the catalog migrations below but still
+        // run the data migrator against the live pre-sqlx data database,
+        // re-executing 001's setval('stream_seq', ...) and 003's DROP TABLE
+        // idempotency_tokens on live data.
+        if bootstrapper
+            .catalog_predates_sqlx()
+            .await
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?
+        {
+            anyhow::bail!(
+                "This catalog predates the sqlx migration system (ADR-0003). In-place \
+                 upgrade is not supported. Run 'extenddb destroy' then 'extenddb init' \
+                 to recreate both databases (this drops all data). See \
+                 docs/manuals/07-upgrade-manual.md."
+            );
+        }
         println!("--- Catalog already initialized. Use 'extenddb migrate' for pending migrations.");
     } else {
+        // run_catalog_migrations applies the schema and writes catalog_version
+        // (the version write moved here from the migration SQL; ADR-0003).
         bootstrapper
             .run_catalog_migrations()
             .await
--- a/crates/app/src/cmd_migrate.rs
+++ b/crates/app/src/cmd_migrate.rs
@@ -3,7 +3,8 @@
 
 //! `extenddb migrate` — apply catalog schema migrations (REQ-CAT-014).
 //!
-//! Reads current catalog version, runs pending migrations, and reports the result.
+//! Reads the current catalog version, runs both migrators (which validate the
+//! checksums of already-applied migrations), and reports the result.
 
 use clap::Args;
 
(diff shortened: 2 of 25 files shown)


crates/app/src/cmd_migrate.rs
```diff
@@ -170,16 +170,15 @@ async fn apply_migrations(
...
-        bootstrap
-            .run_catalog_migrations()
-            .await
-            .map_err(|e| anyhow::anyhow!("{e:?}"))?;
-    }
-
-    // Always run data migrations: the data-database ledger is idempotent
-    // (already-applied migrations are skipped) and independent of the catalog
-    // version, so an upgrade that only touched the data schema is still applied.
+    // Run both migrators unconditionally, even when nothing is pending. sqlx
```
LeeroyHannigan, line 173: **Should-fix:** this now runs unconditionally and write_catalog_version upserts the compiled-in version every time, so an older binary with --yes against a newer catalog that shipped no new migration files rewrites catalog_version backward, and the newer server then refuses to start. Refuse when the stored version parses newer than CATALOG_VERSION, and label the prompt, which currently prints 0.1.1 to 0.1.0 as if it were an upgrade.

crates/app/src/cmd_init.rs
```diff
@@ -322,8 +322,27 @@ async fn run_init_migrations(
         .map_err(|e| anyhow::anyhow!("{e:?}"))?;
 
     if initialized {
+        // ADR-0003: a pre-sqlx deployment cannot be adopted in place. Without
```
LeeroyHannigan, line 325: **Should-fix:** this guard only fires while is_catalog_initialized probes the settings table; a refactor to a different probe (say the sqlx ledger) would make init adopt a pre-sqlx catalog in place and re-run setval and migration 003's drop and recreate of idempotency_tokens against live data, exactly what the comment above warns about. The migrate-path refusal is tested; add the sibling test pinning init refusal on a legacy-shaped catalog.

LeeroyHannigan (commented):
This is careful work. I verified the baseline migration is byte-equivalent to the old fresh-install schema, the pre-sqlx refusal chain across serve, migrate, and init with the destroy plus init stance per ADR-0003, the checksum pins against the file bytes, and the SIGKILL atomicity tests. One Blocking before merge:

**Blocking:** the branch is stale against main and the rebase is semantic, not textual. Main's 002_vector_indexes.sql and 004_vector_index_state.sql were written for the old runner: they carry their own BEGIN and COMMIT, and under sqlx the file's COMMIT ends the per-migration transaction early, so the ledger INSERT runs outside any transaction. That silently reintroduces the applied-but-unrecorded crash gap this PR exists to close, and Postgres only warns on the nested BEGIN, so nothing fails visibly; 002 also writes catalog_version in-file against your new runner-only invariant. On rebase, strip BEGIN/COMMIT and the in-file version write from both files, update their header comments, and re-pin the tripwire checksums. I am calling it out here because a mechanical conflict resolution passes CI silently.

**Should-fix:** add a crates/storage-postgres/build.rs with cargo:rerun-if-changed for migrations and data_migrations; sqlx::migrate! does not track the directory listing, and the tripwire-edit discipline should not be the only thing forcing a recompile. Two smaller Should-fixes inline.

PR #226, opened 2026-07-27
fix: scope backup ARN operations to the caller's account

LeeroyHannigan:
## What

Scope the ARN-addressed backup operations to the caller's account. `DescribeBackup`,
`DeleteBackup` and `RestoreTableFromBackup` previously resolved a caller-supplied
`BackupArn` without reference to the caller's account, so an ARN issued for one
account resolved for any caller. `ListBackups` was already account-scoped; this
brings the ARN-addressed operations in line with it.

- `BackupEngine::describe_backup` and `delete_backup` now take `account_id`, and the
  restore source lookup reuses the caller's `account_id`, so the scoping rule lives in
  the trait rather than in each backend.
- The engine resolves the `BackupArn` field and treats an ARN naming another account
  as absent (`ResourceNotFoundException`) — the same response as an ARN that was never
  issued.
- The Postgres backend filters on `account_id` in the describe, delete and restore
  queries.
- Backup ids now carry the `<epoch_millis>-<8 hex>` suffix DynamoDB uses, instead of a
  bare timestamp — matching the documented ARN format and making ids unique within a
  millisecond.

## Why

The ARN-addressed backup operations were the only backup path not scoped to the
caller's account, which is inconsistent with `ListBackups` and with DynamoDB's
resource model. Bringing them in line closes that inconsistency.

Closes #

## Testing done

Verified live against a Postgres-backed server built from this branch:

- `cargo build --workspace` — clean
- `cargo clippy --workspace --all-targets` — 0 warnings
- `cargo fmt --all -- --check` — exit 0
- `cargo test -p extenddb-engine backup` — 5/5 ARN-resolution unit tests pass
- Rust integration suite — **391 passed, 0 failed** (386 baseline + 5 new)
- Backup tests specifically — 16/16 (5 new `backup_arn_scoping` + 11 existing
  `backup_restore`, unchanged)

The 5 new integration tests drive the ARN directly (no second account provisioned):
another account's ARN returns `ResourceNotFoundException` on Describe, Delete and
Restore; Restore additionally asserts the target table is not created as a side
effect; a positive control confirms the caller's own ARN still resolves; and one test
pins the `<epoch_millis>-<8 hex>` id format.

## Checklist

ADR / RFC: needed — this changes the `BackupEngine` trait signature (adds
`account_id` to `describe_backup`/`delete_backup`). An ADR should capture the
decision before merge.

## Breaking changes

The `BackupEngine` trait signatures for `describe_backup` and `delete_backup` gain an
`account_id` parameter. Any out-of-tree backend implementing `BackupEngine` must update
these signatures and filter on `account_id`.

---

### DIFF
--- a/crates/engine/src/backup.rs
+++ b/crates/engine/src/backup.rs
@@ -8,6 +8,41 @@ use serde_json::{Value, json};
 
 use crate::{OperationContext, serialize_output};
 
+/// Resolve the `BackupArn` field, rejecting ARNs that name a different account.
+///
+/// A backup ARN embeds the owning account:
+/// `arn:aws:dynamodb:<region>:<account>:table/<table>/backup/<id>`. Backups are
+/// account-scoped resources, so an ARN belonging to another account is treated
+/// as absent rather than as a permission error — the same response a caller gets
+/// for an ARN that was never issued, and consistent with `ListBackups`, which
+/// only ever returns the caller's own backups.
+///
+/// Backends also filter on `account_id`; this check keeps the rule in the engine
+/// so it holds for every backend rather than depending on each one to repeat it.
+fn backup_arn_field(body: &Value, account_id: &str) -> Result<String, DynamoDbError> {
+    let backup_arn = body
+        .get("BackupArn")
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| {
+            DynamoDbError::ValidationException(
+                "1 validation error detected: Value null at 'backupArn' \
+                 failed to satisfy constraint: Member must not be null"
+                    .to_owned(),
+            )
+        })?;
+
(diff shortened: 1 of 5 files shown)


robinnsc (approved):

pdf-amzn (approved):
I'm approving because this is a net win, but please follow up on two issues:

* DynamoDB returns AccessDeniedException for foreign-account backup ARNs, not ResourceNotFoundException. All three operations — DescribeBackup, DeleteBackup, RestoreTableFromBackup — return AccessDeniedException ("Access is denied") when given an ARN embedding a different account. The PR's implementation returns ResourceNotFoundException instead.
* Please add integration tests for ARNs that have an empty string for the account id, so it's easy to compare ExtendDB and DynamoDB behavior. Note that some AWS APIs treat the empty string as the caller's account id.

PR #228, opened 2026-07-27
docs: state the catalog trust boundary for the credential encryption key

LeeroyHannigan:
## What

State the catalog trust boundary for the credential encryption key in the security model. The doc noted that the key lives in the catalog database and that a catalog backup includes it, but left the consequence to inference and did not mention that no rotation path exists.

- Add the catalog database contents as an explicit trust boundary: the encryption key and the ciphertext it protects share one trust domain, so catalog read access yields every stored access key secret.
- Scope that boundary to direct database access — the management API and console redact `encryption_key` from settings responses.
- Record that there is no key rotation path; re-keying means re-creating the affected access keys.
- Replace the vague "use a KMS-backed encryption layer at the PostgreSQL level" guidance with what extenddb does and does not do, and what below-extenddb protection can and cannot achieve.
- State that a catalog dump is credential-equivalent and should not be copied to less-protected environments.

## Why

The security model understated a real trust boundary. Making it explicit lets operators
protect the catalog database and its backups to the standard the credentials require,
and sets up a follow-up RFC on key wrapping / rotation.

Closes #

## Testing done

Documentation only, no code change:

- Key storage location and unwrapped form: `bootstrapper.rs:261` (`INSERT INTO settings ... VALUES ('encryption_key', $1)`).
- `encryption_key` redacted from settings API/console responses: `settings.rs:19-25` (`REDACTED_KEYS`).
- No rotation path: grep of `crates/` for rotation/re-key logic returns nothing.

## Checklist

ADR / RFC: n/a — documentation only, no behavior change.

## Breaking changes

None.

---

### DIFF
--- a/docs/manuals/10-security-model.md
+++ b/docs/manuals/10-security-model.md
@@ -19,12 +19,13 @@ This document describes the security architecture of extenddb, including the thr
 1. **Client ↔ extenddb**: Untrusted. All input is validated. SigV4 signatures are verified. IAM policies are evaluated.
 2. **extenddb ↔ PostgreSQL**: Trusted network. The PostgreSQL connection string contains credentials. Use TLS for the PostgreSQL connection in production (`sslmode=require` in the connection string).
 3. **Admin ↔ Management API/Console**: Authenticated via admin credentials or IAM user credentials. CSRF tokens protect the web console.
+4. **Catalog database contents**: Credential-equivalent. The AES-256-GCM key that encrypts access key secrets lives in the catalog `settings` table alongside the ciphertext it protects, so read access to the catalog database yields every stored access key secret. Treat catalog read access, and any catalog backup, as equivalent to holding those credentials. See [Credential Storage](#credential-storage).
 
 ### Out of Scope
 
 - **PostgreSQL security**: extenddb relies on PostgreSQL access controls and network security. Securing the PostgreSQL instance (firewall rules, TLS, authentication) is the operator's responsibility.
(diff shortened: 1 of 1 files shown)


docs/manuals/10-security-model.md
```diff
@@ -19,12 +19,13 @@ This document describes the security architecture of extenddb, including the thr
...
 
 ### Out of Scope
 
 - **PostgreSQL security**: extenddb relies on PostgreSQL access controls and network security. Securing the PostgreSQL instance (firewall rules, TLS, authentication) is the operator's responsibility.
 - **Operating system security**: File permissions on `extenddb.toml`, TLS keys, and the PID file are the operator's responsibility.
-- **Key management**: Access key secrets are encrypted with a locally generated AES key stored in the catalog database. For HSM-grade key management, use a KMS-backed encryption layer at the PostgreSQL level.
+- **Key management**: extenddb generates the credential encryption key locally and stores it unwrapped in the catalog database. It does not wrap the key with an external KMS or HSM, derive it from an operator passphrase, or rotate it. Deployments that require any of those must add protection below extenddb — for example PostgreSQL transparent data encryption, an encrypted volume, or a managed database with a customer-managed KMS key — which protects the key at rest but not from a principal that can query the catalog.
```
jcshepherd, line 28: Maybe something for the backlog here: pluggable support for generating/managing credential encryption keys. Seems like this would be important for a lot of users.

jcshepherd (commented):
Thanks for documenting this. I think it'd be good to cut an issue to support other, more secure, mechanisms for handling the credential encryption key like the ones you mentioned: HSM, passcodes, etc. Status quo is probably not sufficient for many production uses.

PR #232, opened 2026-07-28
fix: authorize batch and transaction operations per-table

LeeroyHannigan:
## What

Authorize batch and transaction operations against each table they touch, instead of a single wildcard resource.

- `BatchGetItem`, `BatchWriteItem`, `TransactGetItems` and `TransactWriteItems` previously authorized once against the `table/*` wildcard, because table names were only extracted from a top-level `TableName`. These operations carry their
  tables in `RequestItems` / `TransactItems`, so a policy scoped to a specific table was not evaluated for them.
- `authorize_request` now extracts every `(action, table)` pair for these four operations, authorizes each against its real per-table ARN, and denies the whole request if any single table is denied (all-or-nothing).
- Actions follow the AWS IAM Service Authorization Reference: `BatchGetItem`/`BatchWriteItem` use their own action per table; `TransactGetItems` maps to `GetItem` per table; `TransactWriteItems` decomposes to `PutItem`/`DeleteItem`/`UpdateItem`/`ConditionCheckItem` per sub-item. Non-batch operations are unchanged.

## Why

The ARN-addressed batch and transaction operations were the only data-plane paths not authorized per table, inconsistent with the single-item operations and with DynamoDB's IAM model. A policy statement scoped to one table had no effect on these operations. This brings them in line.

Closes #

## Testing done

- Rust integration tests (new, `tests/rust/src/batch_transact_authz.rs`): provision a user with an explicit `Deny` on one table via the management API and assert each of the four operations is denied all-or-nothing when it touches that table, with positive controls (a request touching only allowed tables succeeds). Guard tests confirm the action decomposition: a `GetItem` deny does not block `BatchGetItem` but does block `TransactGetItems`; a `PutItem` deny does not block
  `BatchWriteItem` but does block `TransactWriteItems`. 8/8 pass against a live ExtendDB server.
- Full Rust integration suite: 399 passed, 0 failed (no regression on the shared authz path).
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` 0 warnings.

## Checklist

- [x] Clippy is clean -- CI gate `cargo clippy --workspace --all-targets` passes with 0 warnings 
- [x] I have updated documentation if behavior changed -- n/a (authorization correctness fix; no user-facing docs or config surface affected)
- [ ] Breaking changes are noted below (if any) -- none

ADR / RFC: n/a -- no trait, wire, on-disk, or CLI change. This is a correctness
fix that enforces the existing authorization model on the batch/transact paths
(no new auth model); an ADR can be added if a reviewer wants the decision recorded.

## Breaking changes

None. No public interface changes; behavior change is limited to correctly
denying batch/transaction requests that touch a table the caller is denied.

---

### DIFF
--- a/crates/server/src/request_helpers.rs
+++ b/crates/server/src/request_helpers.rs
@@ -62,6 +62,32 @@ pub(crate) async fn authorize_request(
     operation: &str,
     account_id: &str,
 ) -> Result<Option<extenddb_core::types::TableKeyInfo>, DynamoDbError> {
+    // Batch and transaction operations address multiple tables in nested request
+    // structures with no top-level TableName. DynamoDB authorizes each of them
+    // per table, against that table's specific ARN, using the IAM action the
+    // operation maps to (BatchGetItem/BatchWriteItem are their own actions;
+    // TransactGetItems decomposes to GetItem; TransactWriteItems decomposes to
+    // the per-sub-op action). Authorizing the whole request against a single
+    // `table/*` wildcard — as the generic path below does — lets an explicit
+    // Deny on one table be bypassed. Evaluate every (action, table) pair and
+    // reject the entire request if any is denied (all-or-nothing), matching
+    // DynamoDB. Verified against the AWS IAM Service Authorization Reference.
+    if let Some(targets) = batch_transact_authz_targets(operation, input) {
+        for (action_op, table) in targets {
+            let resource_arn = build_resource_arn(&state.region, account_id, Some(&table));
+            authorization::check_authorization(
+                state.authz_cache.as_ref(),
+                identity,
+                &action_op,
+                &resource_arn,
+                false,
+                extenddb_auth::policy::context::RequestParams::default(),
+            )
+            .await?;
+        }
+        return Ok(None);
+    }
+
     let table_name = extract_table_name(input);
     let resource_arn = build_resource_arn(&state.region, account_id, table_name.as_deref());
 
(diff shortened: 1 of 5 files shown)


crates/server/src/request_helpers.rs
```diff
@@ -177,6 +203,69 @@ pub(crate) fn extract_attributes(input: &Value) -> Option<Vec<String>> {
...
+            }
+        }
+        "TransactWriteItems" => {
+            let items = input.get("TransactItems")?.as_array()?;
+            for entry in items {
+                for (sub, action) in [
+                    ("Put", "PutItem"),
+                    ("Delete", "DeleteItem"),
+                    ("Update", "UpdateItem"),
+                    ("ConditionCheck", "ConditionCheckItem"),
```
pdf-amzn, line 252: The ConditionCheck sub-op in TransactWriteItems maps to IAM action dynamodb:ConditionCheck, not dynamodb:ConditionCheckItem. The AWS IAM Service Authorization Reference lists ConditionCheckItem as the resource type required by the dynamodb:ConditionCheck action — not as the action name itself. The action name is dynamodb:ConditionCheck.

This bug is not caught by the Rust tests in tests/rust/src/batch_transact_authz.rs because those tests deny/allow dynamodb:*, which matches any action name. A policy written as Allow dynamodb:ConditionCheck on table/X would fail to authorize with the ConditionCheckItem action string.

pdf-amzn (changes requested):

crates/server/src/request_helpers.rs
LeeroyHannigan: The `ConditionCheck` sub-op does authorize under **`dynamodb:ConditionCheckItem`**, which is what this code produces. Per the AWS DynamoDB developer guide, [Using IAM with DynamoDB transactions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis-iam.html):

> Permissions for Put, Update, Delete, and Get actions are governed by the permissions used for the underlying PutItem, UpdateItem, DeleteItem, and GetItem operations. **For the ConditionCheck action, you can use the `dynamodb:ConditionCheckItem` permission in IAM policies.**

There is no `dynamodb:ConditionCheck` IAM action, and `ConditionCheckItem` is an action name, not a resource type (DynamoDB's resource types are `table`, `index`, `stream`, `backup`, `global-table`).

You're right, though, that the existing `dynamodb:*` tests can't pin the exact action string. I've added two guard tests that do (`7dbf23b`):

- deny exactly `dynamodb:ConditionCheckItem` → the `TransactWriteItems` ConditionCheck is **denied**
- deny the (non-existent) `dynamodb:ConditionCheck` string → the operation is **not** denied

Both pass live against a running server, so the mapping is now pinned to the correct action rather than relying on the wildcard. Let me know if that resolves it.

pdf-amzn (approved):

PR #233, opened 2026-07-29
fix: percent-decode connection string components when parsing

### LINKED ISSUE #223: [Bug] migrate cannot use Unix socket connection strings generated by init
## Describe the bug

`extenddb init` accepts a PostgreSQL Unix socket path and generates a valid percent-encoded PostgreSQL connection string, but `extenddb migrate` cannot read that connection string back.

For example, initialization with:

```text
--pg-host /run/postgresql
```

generates a connection string containing:

```text
@%2Frun%2Fpostgresql:5432/
```

The migration path uses a custom connection-string parser that does not percent-decode the host. It then attempts to connect using `%2Frun%2Fpostgresql` as a hostname, causing a DNS lookup failure.

## To Reproduce

1. Run PostgreSQL using a Unix socket at `/run/postgresql`.

2. Initialize ExtendDB:

   ```bash
   extenddb init \
     --pg-host /run/postgresql \
     --pg-user extenddb \
     --extenddb-user extenddb_runtime \
     --config /var/lib/extenddb/extenddb.toml
   ```

3. Confirm the generated connection string contains:

   ```text
   @%2Frun%2Fpostgresql:5432/
   ```

4. Run:

   ```bash
   extenddb migrate --config /var/lib/extenddb/extenddb.toml
   ```

5. Observe that migration attempts to resolve the encoded socket path as a hostname.

## Expected behavior

`extenddb migrate` should correctly parse the connection string generated by `extenddb init` and connect through the PostgreSQL Unix socket.

## Actual behavior

Migration fails while trying to resolve `%2Frun%2Fpostgresql` as a network hostname.

## Environment

* ExtendDB version: `0.1.0 (using commit 40ae1e1292dc0e689a1bb02b7df89846c1ff504e so should be 0.1.1)`
* Operating system: NixOS
* Rust version (if building from source): 1.96.1
* Client SDK/driver and version: N/A
* Deployment method (binary, container, source): NixOS package managed by systemd

## Logs / stack trace

```text
=== extenddb migrate ===
Config: /var/lib/extenddb/extenddb.toml
INFO extenddb_storage::bootstrapper: Found registered backend: postgres
--- Loading defaults from /var/lib/extenddb/extenddb.toml
Error: Internal("Internal(\"Cannot connect as admin: error communicating with database: failed to lookup address information: ...\")")
```

## Additional context
DISCLAIMER: this issue was diagnosed by an llm agent. I've reviewed it.

## Checklist

* [x] I have searched existing issues for duplicates
* [x] I have included the ExtendDB version and environment details
* [x] I can reproduce this with the latest release

LeeroyHannigan:
## What

`parse_connection_string` now percent-decodes every component (user, password, host, database), mirroring the encoding `extenddb init` applies when it writes the connection string and the decoding libpq/sqlx apply when they read one.

## Why

Fixes #223. `init --pg-host /run/postgresql` writes the host as `%2Frun%2Fpostgresql`. The sqlx runtime path decodes it, so `serve` and `settings` work over the socket. But the custom parser used by `migrate` and `verify` did not decode, handing `%2Frun%2Fpostgresql` to DNS resolution. `init` was writing a config file that half the product could not read back.

Decoding is lenient: a literal `%` that does not form a valid escape passes through unchanged, matching libpq, so a hand-written config with a raw `%` in a password keeps working.

## Testing done

- Live repro of the issue: with a `%2Frun%2Fpostgresql` connection string, `migrate` previously failed with `failed to lookup address information` and now attempts the Unix socket connection.
- New unit tests: plain TCP string unchanged; encoded socket host decodes to `/run/postgresql`; a full round-trip through the same `urlencoding::encode` calls `init` uses (host, user with space, password with `@ : / %`); literal `%` non-escape passes through; wrong scheme rejected.
- `cargo test --workspace` green, fmt and clippy clean.

## Checklist

- [x] Clippy is clean
- [x] Documentation: no behaviour contradicts the docs after this change

## Breaking changes

None.

---

### DIFF
--- a/crates/storage-postgres/src/config.rs
+++ b/crates/storage-postgres/src/config.rs
@@ -59,13 +59,17 @@ pub struct ConnParts {
 /// Handles the standard `postgresql://user:pass@host:port/db` format.
 ///
 /// Percent-decodes every component, mirroring the encoding applied by
-/// `extenddb init` when it writes the connection string (and by libpq/sqlx when
-/// they read one). Without the decode, a Unix socket host such as
-/// `/run/postgresql` — written as `%2Frun%2Fpostgresql` — is treated as a DNS
-/// hostname and fails resolution, so a config that `serve` accepts is rejected
-/// by `migrate` and `verify` (issue #223). Decoding is lenient: a literal `%`
-/// that does not form a valid escape passes through unchanged, matching libpq,
-/// so hand-written configs are unaffected.
+/// `extenddb init` when it writes the connection string (and the decoding
+/// sqlx applies when it reads one on the `serve` path). Without the decode, a
+/// Unix socket host such as `/run/postgresql` — written as
+/// `%2Frun%2Fpostgresql` — is treated as a DNS hostname and fails resolution,
+/// so a config that `serve` accepts is rejected by `migrate` and `verify`
+/// (issue #223). Decoding is lenient: a literal `%` that does not form a
+/// valid escape passes through unchanged. This matches sqlx's parser (libpq
+/// is stricter and rejects malformed escapes); since sqlx is the parser the
+/// `serve` path uses, a strict parser here would recreate the #223 split in
+/// the other direction — a hand-written config with a raw `%` in a password
+/// would work under `serve` but fail under `migrate` and `verify`.
 ///
 /// # Errors
 ///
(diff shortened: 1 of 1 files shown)


crates/storage-postgres/src/config.rs
```diff
@@ -81,12 +91,18 @@ pub fn parse_connection_string(conn: &str) -> anyhow::Result<ConnParts> {
         .parse()
         .map_err(|_| anyhow::anyhow!("Invalid port: {port_str}"))?;
 
+    let decode = |component: &str, what: &str| -> anyhow::Result<String> {
+        Ok(urlencoding::decode(component)
```
pdf-amzn, line 108: note that this won't convert `+` to ` ` which is either good or bad. I think you've done the right thing here, but a unit test with a component containing a space, and one with a plus, would be good.

crates/storage-postgres/src/config.rs
```diff
@@ -113,3 +129,66 @@ impl extenddb_storage::config::StorageConfig for PostgresStorageConfig {
...
+            urlencoding::encode(host),
+        );
+        let parts = parse_connection_string(&conn).expect("encoded connection string must parse");
+        assert_eq!(parts.user, user);
+        assert_eq!(parts.password, pass);
+        assert_eq!(parts.host, host);
+    }
+
+    /// A hand-written config with a literal `%` not forming a valid escape
+    /// must keep working: the decoder is lenient and passes malformed escapes
```
pdf-amzn, line 224: huh. I'm not sure being lenient here is a good thing; being lenient is often something one regrets later. I can be convinced either way though.

pdf-amzn:
probably needs a rebase

crates/storage-postgres/src/config.rs
LeeroyHannigan: Right call to flag it, not converting `+` to space is deliberate and correct here: `+`-as-space is form-encoding (application/x-www-form-urlencoded), not URI percent-encoding. libpq does not treat `+` as space in connection URIs, and `urlencoding::encode` (what `init` writes with) never emits `+` for a space, it emits `%20`. So a literal `+` in a password must survive as `+`.

Added both tests: a component with a space (`%20` decodes to space) and a component with a literal `+` (stays `+`).

crates/storage-postgres/src/config.rs
LeeroyHannigan: You were right to push on this, I tested both authorities live and the PR text I wrong about one of them:

- libpq is STRICT: `psql "[REDACTED: credential]..."` fails with `invalid percent-encoded token: "se%ZZcret"`. The "matching libpq" claim in the description was incorrect; I've fixed the doc comment and PR body.
- sqlx (our 0.8, the parser `serve` actually uses) is LENIENT: the same string parses fine via `PgConnectOptions::from_str`.

That second point is the real argument for leniency: this custom parser exists alongside sqlx's, and #223 was precisely a config that one parser accepted and the other rejected. If we make this parser strict, a hand-written config with a raw `%` in the password works under `serve` (sqlx) but fails under `migrate`/`verify`,
recreating the split-brain this PR fixes, in the other direction. Consistency with the runtime parser wins over strictness, so lenient it stays, now documented with the correct rationale (sqlx-parity, not libpq-parity).

pdf-amzn (approved):

PR #236, opened 2026-08-05
RFC: vector search support across storage backends

LeeroyHannigan:
## Summary

DynamoDB reached GA with native vector search on 2026-08-05: a new index type created over an attribute holding an embedding, plus a `SearchVectors` operation returning nearest neighbours ranked by a distance function, with optional exact-match inline filtering.

An application that adopts it cannot use ExtendDB for that workload at all today. `CreateTable` rejects the unknown index type and `SearchVectors` is an unknown operation, so local development, CI, and offline or edge deployments lose the ability to exercise the code path.

This RFC proposes adding that surface.

## The central design question

The RFC separates what must be identical across backends from what cannot be.

Wire surface, validation, score sign convention, index lifecycle and the consistency model are defined once in the engine. **ANN execution is a declared per-backend capability** (`Ann`, `ExactScan`, `Unsupported`), because the stores diverge sharply:

- PostgreSQL has `pgvector`, but only if the operator installed it
- SQLite has `sqlite-vec`, likewise a loadable extension
- MongoDB has vector search in Atlas, not in community `mongod`

Mandating ANN everywhere would convert a feature gap into a backend gap. The recommendation is that vector search be **optional** for backend acceptance, provided a backend that cannot do it fails closed at `CreateTable` with a typed error rather than returning wrong results later.

## Two points that carry most of the risk

**Score semantics are asymmetric.** Lower is more similar for `Cosine` and `Euclidean` (0 is identical); higher is more similar for `DotProduct`. Normalising all three into one "similarity" number inverts the ranking for two of them. I have seen this exact defect in a first-party client library against the real service, so the RFC asks for the sign convention to be asserted per distance function rather than inferred from result sets.

**A vector index is eventually consistent, the same model as a GSI.** So the proposal reuses the existing pending-index queue and `gsi_propagation_delay_ms` rather than adding a second asynchronous mechanism. It also requires that an index must not report `ACTIVE` while its backfill is still draining, with the transition causally ordered after the drain rather than scheduled on a wall clock beside it.

## Testing

Recall is not assertable by equality, so `ExactScan` is proposed as the oracle (perfect recall) with ANN backends held to a recall threshold against it. Everything deterministic (sign convention, top K truncation, partition-key scoping, filter rejection, lifecycle) is asserted exactly.

## What I am asking reviewers for

1. Should vector search be optional for backend acceptance? I propose yes.
2. What recall threshold should ANN backends be held to, and is it a gate or a reported metric?
3. Should `ExactScan` be reachable in a production configuration at all, or gated?

Five service behaviours are listed as unresolved rather than guessed at, including the dimension-mismatch error on `PutItem` and whether a missing vector attribute gives sparse-index behaviour. I am happy to verify each against the real service before the corresponding code is written.

Note on numbering: the file is `0000-vector-search.md` per the process in `docs/rfcs/README.md`, which keeps `0000` until acceptance.

### DIFF
--- a/docs/rfcs/0000-vector-search.md
+++ b/docs/rfcs/0000-vector-search.md
@@ -1,0 +1,318 @@
+# RFC-NNNN: Vector search support across storage backends
+
+- Status: Draft
+- Author: @LeeroyHannigan
+- Created: 2026-08-05
+- FCP ends: (set when entering Final Comment Period)
+- Tracking issue: #NNN
+
+## Summary
+
+DynamoDB became generally available with native vector search on 2026-08-05. It
+adds a new index type created over an attribute holding an embedding, and a new
+`SearchVectors` operation that returns the nearest neighbours to a query vector,
+ranked by a distance function, with optional exact-match filtering. This RFC
+proposes adding that surface to ExtendDB.
+
+The proposal is deliberately split in two. The wire surface, validation,
+consistency model and index lifecycle are defined once in the engine and are
+identical for every backend. Approximate nearest neighbour (ANN) execution is
+declared as a per-backend capability, because the underlying stores differ
+sharply in what they can do, and requiring ANN everywhere would either block
+backends from being accepted or force us to ship something that silently returns
+wrong results.
+
+## Motivation
+
+ExtendDB's value is that an application written against DynamoDB runs unchanged
+against a backend the operator chooses. Vector search is not a peripheral
+addition to DynamoDB's surface: it changes what a table is for. The launch
+targets agentic memory, retrieval augmented generation, recommendations and
+anomaly detection, and the pitch is explicitly that you no longer copy
+operational data into a separate vector store.
+
+That has a direct consequence for us. An application that adopts vector search
+in DynamoDB and then points at ExtendDB does not degrade gracefully: it fails at
+`CreateTable` on an unknown index type, or at `SearchVectors` on an unknown
+operation. Local development, CI, and any offline or edge deployment that
+depends on ExtendDB stops being able to exercise the code path at all. The gap
+is not a missing optimisation, it is a missing feature that silently pushes
+users back onto a real endpoint for the one workload they most want to test
+cheaply and repeatedly.
+
+Evidence the surface is real and stable enough to target: it is GA in all
+commercial Regions plus GovCloud (US), it is documented in the DynamoDB
+Developer Guide, and it is exposed in the console, CLI, SDKs and
+CloudFormation. This is not a preview we would be chasing.
+
+## Detailed design
+
+### Data model: no new attribute type
+
+Embeddings are stored in the existing `L` (list) type, where each element is an
+`N` (number) holding one float. There is no new attribute type and no schema
+change to store a vector.
+
+This is the single most important fact for our implementation, and it is good
+news. The `AttributeValue` codec, the item serialisation path, the 400KB item
+limit check, and every existing validation rule stay exactly as they are. A
+vector is an ordinary attribute that an index happens to interpret numerically.
+Writes go through unchanged `PutItem` and `UpdateItem`; there is no vector write
+path to add.
+
+One interaction worth stating explicitly: a 4096 dimension embedding is 4096
+numbers inside one item, so vectors consume the existing 400KB item budget. The
+existing limit check applies unchanged and we should not special case it.
+
+### Index configuration
+
+A vector index is created on a table and carries:
+
+- the vector attribute name
+- the number of dimensions, up to 4096
+- a distance function: `Euclidean`, `Cosine`, or `DotProduct`
+- an optional partition key attribute
+- optional inline filter attributes
+- an attribute projection
+
+The partition key deserves attention because it changes query semantics rather
+than only physical layout. When a vector index declares a partition key, each
+search is scoped to a single value of it. A search is therefore not global over
+the index; it is nearest neighbours within one partition. Our implementation
+must treat the partition key as part of the query contract, not as a
+distribution hint we are free to ignore.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+targets agentic memory, retrieval augmented generation, recommendations and
+anomaly detection, and the pitch is explicitly that you no longer copy
+operational data into a separate vector store.
+
+That has a direct consequence for us. An application that adopts vector search
+in DynamoDB and then points at ExtendDB does not degrade gracefully: it fails at
+`CreateTable` on an unknown index type, or at `SearchVectors` on an unknown
+operation. Local development, CI, and any offline or edge deployment that
+depends on ExtendDB stops being able to exercise the code path at all. The gap
+is not a missing optimisation, it is a missing feature that silently pushes
```
jcshepherd, line 39: This feels like a bit of an overstatement, especially for a feature that's been out for a week. :-) I think this can be simplified to vector search being a new first-class feature of DynamoDB, so ExtendDB should support it to support its mission of letting applications written against DynamoDB run unchanged against ExtendDB. it's that simple.

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+users back onto a real endpoint for the one workload they most want to test
+cheaply and repeatedly.
+
+Evidence the surface is real and stable enough to target: it is GA in all
+commercial Regions plus GovCloud (US), it is documented in the DynamoDB
+Developer Guide, and it is exposed in the console, CLI, SDKs and
```
jcshepherd, line 45: It'd be good to provide links to documentation references.

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+The proposal is to extend that path to carry vector index maintenance rather
+than to introduce a second, parallel asynchronous mechanism. One queue, one
+recovery story, one setting operators already understand.
+
+Two lifecycle requirements follow, and the second is a bug we should prevent by
+design rather than fix later:
+
+1. Index creation is asynchronous and the index reports `CREATING` until the
+   backfill over existing items completes, then `ACTIVE`.
+2. The index must not report `ACTIVE` while its backfill is still running. A
```
jcshepherd, line 128: Is this specific to vector indexes?

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+- top K at most 100
+- inline filter conditions exact-match only
+- search scoped to exactly one partition key value when the index declares one
+
+### Implementation sketch
+
+1. `crates/core`: index configuration and `SearchVectors` request and response
+   types, dimension and top K validation, distance function enum.
+2. `crates/storage`: extend `TableEngine` for vector index create, describe and
+   delete; add the search entry point to `DataEngine`; add the capability
```
jcshepherd, line 203: It sounds like it's not just "capability declaration", but actually "detection". Is that right? In other words, since for at least some backends ANN capability is based on configuration or product version, there needs to be a runtime check: we can't compile the capability directly into the plugin because the plugin will run on different instances with different capabilities. Correct?

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+   declaration to `Backend`. This is the breaking trait change and should land
```
jcshepherd, line 204: Can it default to something reasonable so it's not breaking?

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+## Drawbacks
+
+- **Extension dependency.** The only credible ANN path for our reference backend
+  is `pgvector`, which the operator must install. That is a new external
+  dependency in a project that has so far kept its backend requirements to a
+  stock database, and it makes capability depend on deployment rather than on
+  our build.
+- **A correct-but-slow mode invites misreading.** `ExactScan` will be measured
+  by somebody as though it were ANN, and we will be compared unfavourably. The
+  mitigation is documentation and an explicit capability in `DescribeTable`, not
```
jcshepherd, line 229: The DynamoDB DescribeTable? How does that work?

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+   confirmed rather than assumed.
+3. Can a vector index be created on a table that is not `ACTIVE`, and can more
+   than one be created concurrently?
+4. Is there a per-table limit on vector indexes, as GSIs have?
+5. Exact error codes and messages for exceeding 4096 dimensions, exceeding top K
+   of 100, and supplying a range condition as an inline filter.
+
+Design questions for reviewers specifically:
+
+6. Should vector search be optional for backend acceptance? I propose yes.
```
jcshepherd, line 282: Yes. Maybe in 2-10 years we can revisit.

docs/rfcs/0000-vector-search.md
```diff
@@ -0,0 +1,318 @@
...
+7. What recall threshold should ANN backends be held to in conformance, and
+   should it be a hard gate or a reported metric?
+8. Should `ExactScan` be allowed in a non-development build at all, or should it
```
jcshepherd, line 285: Is the concern "cost"? Is the cost evident to the developer (kind of like 'scan' is its own operation in part to enforce understanding that its performance and costs is quite different from other operations).

jcshepherd (commented):
A few naive questions below. Also, I wonder if it'd be helpful to outline a PG or MySQL implementation: in particular, what the physical index schema is and what the backend query looks like. I don't have a good feel for either.

PR #239, opened 2026-08-06
fix(storage): tie-break GSI pagination on the full base primary key

### LINKED ISSUE #238: [Bug] GSI pagination with duplicate index sort keys returns no items on a composite base table
## Describe the bug

A Query on a GSI whose index sort key values are duplicated returns page one plus a `LastEvaluatedKey`, but resuming with that same key returns zero items. Nothing errors, so a paginating client reads page one and concludes it has everything.

The base table needs a range key. `TestGSIOnHashOnlyBaseTable::test_paginate_duplicate_gsi_sort_keys` already covers this index shape on a hash only base table and passes; adding a range key to that fixture reproduces the failure. Same class as #145.

Reproduced on both the `sqlite` and the default `postgres` backend at aefd1b6.

## To Reproduce

```bash
EP=https://localhost:18443   # or your endpoint

aws dynamodb create-table --table-name esk \
  --attribute-definitions \
      AttributeName=pk,AttributeType=S AttributeName=sk,AttributeType=S \
      AttributeName=gsipk,AttributeType=S AttributeName=gsisk,AttributeType=S \
  --key-schema AttributeName=pk,KeyType=HASH AttributeName=sk,KeyType=RANGE \
  --global-secondary-indexes '[{"IndexName":"gsi","KeySchema":[{"AttributeName":"gsipk","KeyType":"HASH"},{"AttributeName":"gsisk","KeyType":"RANGE"}],"Projection":{"ProjectionType":"ALL"}}]' \
  --billing-mode PAY_PER_REQUEST --endpoint-url $EP
aws dynamodb wait table-exists --table-name esk --endpoint-url $EP

# Five items sharing one index sort key value, differing only in the base keys.
for i in 1 2 3 4 5; do
  aws dynamodb put-item --table-name esk --endpoint-url $EP \
    --item "{\"pk\":{\"S\":\"item$i\"},\"sk\":{\"S\":\"S\"},\"gsipk\":{\"S\":\"P\"},\"gsisk\":{\"S\":\"DUP\"}}"
done

# Page one: Count 2, plus a LastEvaluatedKey.
aws dynamodb query --table-name esk --index-name gsi --limit 2 --endpoint-url $EP \
  --key-condition-expression "gsipk = :p" \
  --expression-attribute-values '{":p":{"S":"P"}}'

# Page two, resuming from that LastEvaluatedKey verbatim.
aws dynamodb query --table-name esk --index-name gsi --limit 2 --endpoint-url $EP \
  --key-condition-expression "gsipk = :p" \
  --expression-attribute-values '{":p":{"S":"P"}}' \
  --exclusive-start-key '{"gsipk":{"S":"P"},"gsisk":{"S":"DUP"},"pk":{"S":"item2"},"sk":{"S":"S"}}'
```

## Expected behavior

Page two returns the remaining items, as it does for the same index shape on a hash only base table.

## Actual behavior

Page two returns nothing, with no error:

```json
{"Count": 0, "ScannedCount": 0, "Items": []}
```

## Environment

- ExtendDB version: 0.1.2, commit aefd1b68
- Operating system: macOS 15 (Darwin 24.6.0)
- Rust version (if building from source): 1.97.1
- Client SDK/driver and version: AWS CLI v2

## Additional context

A regression test is the existing `gsi_on_hash_only_base_table` fixture with a range key added to its `key_schema`; the surrounding test body needs no change. Happy to open a PR.

DISCLAIMER: this issue was diagnosed with an LLM agent. I've reviewed and reproduced it.

## Checklist

- [x] I can reproduce this on current `main` (aefd1b68)

LeeroyHannigan:
## What

Fixes the tie-breaker used to resume a paginated Query on a global secondary index. Both backends built the `ExclusiveStartKey` predicate as

```
idx_sk > :sk OR (idx_sk = :sk AND base_sk > :base_sk)
```

treating the base table's **sort** key as the uniqueness tie-breaker. It is not one. A GSI entry is unique on `(index SK, base PK, base SK)`: many base partitions can project the same index sort key, and those rows can also share a base sort key. When they did, both disjuncts were false and page two came back empty. The predicate is now

```
idx_sk > :sk OR (idx_sk = :sk AND (base_pk > :base_pk
                 OR (base_pk = :base_pk AND base_sk > :base_sk)))
```

and the `ORDER BY` gained `base_pk` so the ordering and the predicate agree. They have to agree: ordering by base sort key alone leaves rows that share an index sort key and a base sort key in an arbitrary order, which no `ExclusiveStartKey` can resume from deterministically.

Changed in `crates/storage-sqlite/src/data/query_scan.rs` and `crates/storage-postgres/src/data/query_scan.rs`. The postgres `PaginationBinds` doc comments in `data/query.rs` are corrected, because `BaseSkOnly` is now LSI-only and its previous comment described the broken case as the intended one.

**LSI behaviour is deliberately unchanged.** An LSI query always constrains the partition key, so every row shares it and the base sort key alone does identify a row. It is also a user-visible sort dimension there, so it continues to follow `ScanIndexForward`. Only the GSI path changed.

This also removes a divergence inside the codebase: the index **scan** path already ordered by the full base key (`ORDER BY pk, idx_sk, base_pk, base_sk`), while the query path did not.

## Why

A Query on a GSI with duplicated index sort keys returned page one plus a `LastEvaluatedKey`, and resuming from that key returned zero items with no error. Nothing failed loudly, so a paginating client read page one and concluded it had
everything. Silent partial results are worse than an error, because the caller has no signal to retry on.

This is why a hash-only base table was unaffected and why `TestGSIOnHashOnlyBaseTable` passed throughout: there the base partition key is the whole base primary key, so the existing tie-breaker was already complete. Adding a range key to the base table is what exposed it.

Closes #238

## Testing done

Three new tests in `TestGSIOnCompositeBaseTable` (`tests/test_query_scan.py`), against a live server on both backends, with a negative control in each case.

| | postgres | sqlite |
|---|---|---|
| unmodified `main` | 2 failed, 1 passed | 2 failed, 1 passed |
| with this change | 3 passed | 3 passed |

The test that passes in both directions is `test_paginate_duplicate_gsi_sort_keys_distinct_base_sort_keys`, the case the old tie-breaker could already handle. It is included so a fix that only works when the base sort key happens to break the tie cannot pass. Reverse pagination is covered too, because the base key tie-breaker stays ascending while
the index sort key reverses, so the predicate and the `ORDER BY` have to agree on that asymmetry.

Regression coverage:

- `cargo test --workspace`: 668 passed, 0 failed, 0 filtered out.
- `cargo fmt --all -- --check`: clean, zero diffs.
- `cargo clippy --all-targets -- -D warnings`: clean.
- `cargo clippy --workspace --all-targets -- -W clippy::pedantic`: 421 warnings,
  identical to the count on `main`, so this change adds none. It does grow one
  pre-existing `too many lines` violation in the postgres query builder from
  208/100 to 228/100; that function was already twice over the limit and
  splitting it is out of scope for a bug fix.
- Full `tests/test_query_scan.py` on postgres with the change: 83 passed,
  2 failed. Both failures are `TestBaseKeySchemaFlow::test_index_pagination_uses_base_key_schema_for_tiebreaker`
  and `::test_index_scan_pagination_uses_base_key_schema`. They fail identically
  on unmodified `main` in my environment and **pass in CI on the same commit**
  (`d6afa1e`), so they are unrelated to this change. They are not the GSI
  propagation race either: they still fail with `gsi_propagation_delay_ms` set to
  0, and unlike their siblings they do not poll with `wait_for_gsi_items`. I have
  not root-caused that divergence and am reporting it separately rather than
  bundling it here.

## Checklist

- [x] All tests pass (`cargo test --workspace`) - 668 passed, 0 filtered out
- [x] Clippy is clean (`cargo clippy -- -W clippy::pedantic`) - no new warnings
      versus `main` (421 both sides); see the note above on the pre-existing
      function-length violation
- [x] I have updated documentation if behavior changed - the `PaginationBinds`
      doc comments, which previously described the broken tie-breaker as correct

ADR / RFC: n/a. No trait, wire, on-disk or CLI surface changes; this corrects
SQL generation inside two backends to match documented DynamoDB pagination
behaviour.

## Breaking changes

None. `LastEvaluatedKey` keeps the same shape and contents, so keys issued by an
older build remain usable.

One behaviour change worth stating explicitly, though it is not breaking: the
order in which rows with tied index sort keys are returned was previously
arbitrary and is now deterministic, ordered by the base primary key. Callers
could not have depended on the old order, since it was the absence of a total
order that made pagination lose rows.

---

### DIFF
--- a/crates/storage-postgres/src/data/query.rs
+++ b/crates/storage-postgres/src/data/query.rs
@@ -24,9 +24,13 @@ pub(crate) enum PaginationBinds {
     None,
     /// Index query where base table has no SK — only `base_pk` as tie-breaker.
     BasePkOnly { pk_text: String },
-    /// Index query where base table has a SK — `base_sk` as tie-breaker.
+    /// LSI query. `base_sk` alone is a complete tie-breaker, because every
+    /// row shares the queried partition key.
     BaseSkOnly { sk: SortKeyValue },
-    /// Hash-only index where base table has a SK — both `base_pk` and `base_sk`.
+    /// Full base primary key as tie-breaker. Required by a GSI query when the
+    /// base table has a SK: rows sharing an index SK can come from different
+    /// base partitions and can also share a base SK, so `base_sk` alone does
+    /// not identify a row. Also used by a hash-only index on such a table.
     BasePkAndSk { pk_text: String, sk: SortKeyValue },
 }
 
--- a/crates/storage-postgres/src/data/query_scan.rs
+++ b/crates/storage-postgres/src/data/query_scan.rs
@@ -46,17 +46,34 @@ fn build_pagination_where(
             } else {
                 ""
             };
-            // LSI: base SK follows ScanIndexForward because items share the
-            // same partition and the base SK is part of the composite sort order.
-            // GSI: base SK is always ">" (ascending) because it's only a
-            // uniqueness tie-breaker, not a user-visible sort dimension.
-            let base_cmp = if is_lsi { cmp } else { ">" };
-            format!(
-                " AND ({sk_col}{collate} {cmp} ${p1} OR \
-                 ({sk_col}{collate} = ${p1} AND {base_col}{base_collate} {base_cmp} ${p2}))",
-                p1 = param_idx,
-                p2 = param_idx + 1
-            )
+            if is_lsi {
+                // LSI: every row shares the queried partition key, so the base
+                // table's sort key alone identifies a row uniquely. It is also a
+                // user-visible sort dimension, so it follows ScanIndexForward.
+                format!(
+                    " AND ({sk_col}{collate} {cmp} ${p1} OR \
+                     ({sk_col}{collate} = ${p1} AND {base_col}{base_collate} {cmp} ${p2}))",
+                    p1 = param_idx,
+                    p2 = param_idx + 1
+                )
+            } else {
+                // GSI: the tie-breaker must be the FULL base primary key. Rows in
+                // a GSI partition are unique on (index SK, base PK, base SK), not
+                // on (index SK, base SK): many base partitions can project the
+                // same index SK and the same base SK. Comparing base SK alone
+                // made a page-two query return nothing whenever the rows sharing
+                // an index SK also shared a base SK, so a paginating client
+                // silently stopped after page one. The base key stays ascending
+                // because it is a uniqueness tie-breaker, not a sort dimension.
+                format!(
+                    " AND ({sk_col}{collate} {cmp} ${p1} OR \
+                     ({sk_col}{collate} = ${p1} AND (base_pk COLLATE \"C\" > ${p2} OR \
+                     (base_pk = ${p2} AND {base_col}{base_collate} > ${p3}))))",
+                    p1 = param_idx,
+                    p2 = param_idx + 1,
+                    p3 = param_idx + 2
+                )
+            }
         } else if is_index {
             // Index with no base SK — use base_pk as tie-breaker
             format!(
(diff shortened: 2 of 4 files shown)


amrith (approved):
approved

PR #245, opened 2026-08-07
fix(storage): read GSI propagation delay live so runtime changes apply

LeeroyHannigan:
## What

Makes both storage backends honour a runtime change of the GSI propagation delay
immediately. The delay getter now reads `gsi_propagation_delay_ms` from the catalog
on each write instead of serving a value cached with a 30-second refresh, in
`crates/storage-sqlite/src/store.rs` and `crates/storage-postgres/src/lib.rs`, with
the four write paths in each backend updated to await it.

PostgreSQL callers keep their `indexes.is_empty()` short-circuit, so a table with no
secondary indexes performs no extra query at all.

Three further changes, from reviewing the SQLite half:

- The delay is read **before** `write_lock.lock().await` at all four SQLite call
  sites. It is a runtime setting rather than an invariant of the write, so it does
  not need to be read under the lock, and that lock serialises every write in the
  process, so work inside it is the backend's throughput bottleneck.
- The error arm logs instead of swallowing. Both poll workers already log through
  `tracing::debug!`; now a live read that falls back to the cache does too, so a
  degraded catalog serves a stale delay loudly rather than silently.
- The default is a named constant, `DEFAULT_GSI_PROPAGATION_DELAY_MS`, per backend
  crate, shared by the live read and the poll worker. The literal was previously
  duplicated and the live read had added a third copy.

## Why

Setting `gsi_propagation_delay_ms` to 0 at runtime is supposed to make index updates
synchronously visible. With the cached value, writes kept using the stale non-zero
delay for up to 30 seconds, so the index update stayed async and queries against the
GSI returned 0 items immediately after a write. This is one of the red CI jobs on
#207, where the zero-delay GSI sync test sets the delay to 0 and immediately expects
synchronous visibility.

PostgreSQL had the identical defect. Both backends are fixed here rather than in two
PRs, following #248 where the same shared defect was corrected in both backends in
one change.

Closes # 

## Testing done

Every claim below was executed, on a fresh deployment per backend and on ports away
from any other local instance.

- **SQLite**: `test_gsi_sync_path_with_zero_delay` passes **20/20** consecutive runs.
  Full `TestGsiAsyncPropagation` class 5/5.
- **PostgreSQL**: the same test **failed twice before this change** with the same
  symptom as SQLite, `assert 0 == 1`, and now passes **20/20** consecutive runs.
- **Negative control**: reverting only the SQLite getter body to the cached read,
  keeping the `async` signature so it still compiles, reproduces the failure exactly.
  So the test discriminates and the fix is what makes it pass.
- Crate unit tests: 16/16 SQLite, 14/14 PostgreSQL, 0 filtered out.
- `cargo fmt --all -- --check` exit 0. `cargo clippy -D warnings` exit 0 on both the
  `sqlite` and `postgres` feature sets.

**One pre-existing failure found and deliberately not touched.**
`test_gsi_configured_delay_range` fails on PostgreSQL, observing about 1010 ms against
a 500 ms bound with the delay set to 50 ms. An A/B on a fresh server shows it failing
identically with and without this change, so it is not a regression here. It looks
like worker scheduling granularity dominating the configured delay, which deserves its
own investigation.

**Worth knowing about CI.** `run-integration` (PostgreSQL) was green on the previous
revision of this PR while PostgreSQL was still broken, so CI did not catch this class
of defect. Whether that is because the harness restarts the server, letting PostgreSQL
pick the setting up at construction, is unconfirmed.

## Checklist

- [x] Clippy is clean (`cargo clippy -- -D warnings` on both feature sets; the repo
      carries pre-existing `-W clippy::pedantic` warnings and this change adds none)
- [ ] I have added or updated tests for new functionality — no new test: the existing
      zero-delay test in the Python suite already covers the behaviour on both
      backends, and it is what the fix makes pass
- [ ] I have updated documentation if behavior changed — no doc change; the setting's
      documented meaning is unchanged, it is now actually honoured

ADR / RFC: n/a. Internal bug fix, no contract surface changed.

## Breaking changes

None.

---

### DIFF
--- a/crates/storage-postgres/src/data/delete_item.rs
+++ b/crates/storage-postgres/src/data/delete_item.rs
@@ -39,8 +39,7 @@ impl PostgresEngine {
         let sys_delay = if indexes.is_empty() {
             0
         } else {
-            self.gsi_default_delay_ms
-                .load(std::sync::atomic::Ordering::Relaxed)
+            self.gsi_default_delay().await
         };
 
         let needs_tx = condition.is_some() || return_old || !indexes.is_empty() || stream.is_some();
--- a/crates/storage-postgres/src/data/put_item.rs
+++ b/crates/storage-postgres/src/data/put_item.rs
@@ -59,8 +59,7 @@ impl PostgresEngine {
         let sys_delay = if indexes.is_empty() {
             0
         } else {
-            self.gsi_default_delay_ms
-                .load(std::sync::atomic::Ordering::Relaxed)
+            self.gsi_default_delay().await
         };
 
         // When there's a condition, return_old, indexes, or stream capture, we need a transaction
--- a/crates/storage-postgres/src/data/transactions.rs
+++ b/crates/storage-postgres/src/data/transactions.rs
@@ -83,10 +83,9 @@ impl PostgresEngine {
             }
         }
 
-        // D-4: Read system default delay from cache (P119).
-        let sys_delay = self
-            .gsi_default_delay_ms
-            .load(std::sync::atomic::Ordering::Relaxed);
+        // D-4: Read the system default delay live (P119), so a runtime change
+        // applies to this transaction rather than up to 30 s later.
+        let sys_delay = self.gsi_default_delay().await;
 
         let mut tx = self
             .data_pool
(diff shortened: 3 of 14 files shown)


amrith (approved):
approved

PR #246, opened 2026-08-07
fix(postgres): restore must not report ACTIVE before the data copy co…

LeeroyHannigan:
# PR body: fix/restore-active-implies-copy-complete

## What

Fixes a race in the Postgres restore path where a restored table could report `ACTIVE` in `DescribeTable` before its data copy finished. `create_table_impl` gains a `defer_active` flag: the restore path passes `true`, writing the table row `CREATING` with no scheduled transition, so the explicit `ACTIVE` update at the end of the row copy is now the only status flip. The public `TableEngine::create_table` path passes `false` and is unchanged.

## Why

DynamoDB's contract is that the first `ACTIVE` observation on a restore target implies the restored data is fully present. Our restore created the target through the normal create-table flow, which schedules the `CREATING` to `ACTIVE` transition on the creation-delay timer. Copying a large backup outlasts that timer, so a client that waits-for-ACTIVE can observe an empty or partial table. Reproduced live on baseline: `ACTIVE` at 515/40000 items. This is the same defect class the MongoDB backend in #207 fixed on its side, and it is what makes that PR's dual-target restore test red on the Postgres runner.

Closes # (no standing issue; surfaced by the restore-completeness test in #207's CI)

## Testing done

Live against Postgres 16, TLS, SigV4 auth:

- New integration test `restore_active_completeness::restored_table_has_all_items_when_first_active`: an observer task spawns before the restore, polls `DescribeTable`, and counts items via a paginated `Select=COUNT` scan at the first `ACTIVE` observation. Transient describe errors are retried; scan errors fail the test rather than truncating the count.
- Negative control on baseline (fix reverted): FAILED with `ACTIVE` at 515/40000 items (112s), confirming the test discriminates.
- With the fix: the test passes (129s), and the full Rust integration suite with `--test-threads=1` is 418/421 passed, 0 filtered out. The 3 failures are `capacity_throttling` tests that require `settings set throttling_enabled true`, which my ad-hoc harness omitted; CI sets it in `integration.yml`.
- `cargo fmt --check` exit 0, `cargo clippy` clean on the touched crate.

## Checklist

ADR / RFC: n/a (internal bug fix; `create_table_impl` is `pub(crate)`, no contract surface changed)

---

### DIFF
--- a/crates/storage-postgres/src/backup_engine.rs
+++ b/crates/storage-postgres/src/backup_engine.rs
@@ -8,7 +8,6 @@ use extenddb_core::types::{
     PointInTimeRecoveryDescription, SourceTableDetails, TableDescription,
 };
 use extenddb_storage::BackupEngine;
-use extenddb_storage::TableEngine;
 use extenddb_storage::error::StorageError;
 use futures::future::BoxFuture;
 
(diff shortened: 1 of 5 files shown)


amrith (approved):

amrith:
approved and merging when ready

PR #250, opened 2026-08-10
fix(postgres): concurrent UpdateItem upserts creating one item must all succeed

### LINKED ISSUE #249: Concurrent UpdateItem upserts to the same new item fail with ConditionalCheckFailedException
**Version:** built from source at v0.1.3 (main, `3961c63`), Postgres 16 backend, builtin auth, default settings.

## Summary

When two or more UpdateItem calls with no ConditionExpression race to create the same item (same table, same key, item does not exist yet), one call succeeds and the others fail with `ConditionalCheckFailedException`. DynamoDB serializes these writes and all of them succeed.

I found this while evaluating ExtendDB against an existing production DynamoDB workload. Our request flow issues two unconditional upserts to the same configuration item in parallel on first write, which works on DynamoDB but returns a 500 through ExtendDB every time a new item is first created.

## Repro

Node, any AWS SDK v3, against a fresh table:

```js
import {DynamoDBClient, CreateTableCommand, UpdateItemCommand} from "@aws-sdk/client-dynamodb";
import {marshall} from "@aws-sdk/util-dynamodb";

const c = new DynamoDBClient({endpoint: "https://localhost:18443", region: "us-east-1"});

await c.send(new CreateTableCommand({
  TableName: "Repro",
  BillingMode: "PAY_PER_REQUEST",
  AttributeDefinitions: [{AttributeName: "pk", AttributeType: "S"}],
  KeySchema: [{AttributeName: "pk", KeyType: "HASH"}],
}));
await new Promise((r) => setTimeout(r, 6000));

const upsert = (i) => c.send(new UpdateItemCommand({
  TableName: "Repro",
  Key: marshall({pk: "same-key"}),
  UpdateExpression: "SET #v = :v",
  ExpressionAttributeNames: {"#v": "value"},
  ExpressionAttributeValues: marshall({":v": `writer-${i}`}),
  ReturnValues: "ALL_NEW",
}));

const results = await Promise.allSettled([upsert(1), upsert(2), upsert(3), upsert(4)]);
results.forEach((r, i) => console.log(i, r.status, r.reason?.name ?? ""));
```

## Observed

```
0 rejected ConditionalCheckFailedException
1 rejected ConditionalCheckFailedException
2 fulfilled
3 rejected ConditionalCheckFailedException
```

Sequential create-then-update on the same key works fine (0 failures in 20 runs), so this is specific to concurrent creation of a not-yet-existing item.

## Expected

All four calls succeed, as they do on DynamoDB. UpdateItem without a ConditionExpression should never surface ConditionalCheckFailedException.

## Suspected cause

The losing writers appear to take the insert path, hit the Postgres unique constraint violation on the primary key, and map it to a conditional failure instead of retrying as an update. An `INSERT ... ON CONFLICT DO UPDATE` in the storage-postgres upsert path, or a retry of the update path on unique violation, would presumably resolve it.

Happy to provide more detail or test a fix against our workload.

LeeroyHannigan:
## What

Concurrent `UpdateItem` calls with no `ConditionExpression` racing to create the
same not-yet-existing item now all succeed, and each writer's update expression is
applied on top of whatever the previous writer committed.

## Why

Closes #249.

Reproduced on this backend before changing anything, at the exact commit the
reporter cited (`3961c63`):

- 10 of 20 calls failed with `ConditionalCheckFailedException` across five trials
- four writers each setting a **different** attribute left an item holding only
  one of them, so three writers' data was lost outright
- sequential create-then-update was unaffected, which is why this only appears
  under concurrency

Measured against real DynamoDB in us-east-1 for comparison: six concurrent writers
on one brand-new key succeeded 6/6 across three trials, and four writers setting
four different attributes produced an item holding all four. The service serialises
writes to one item, and an `UpdateItem` without a condition is an upsert, so every
writer must succeed and each expression applies on top of the previous commit.

## Root cause

Every read in this layer already uses `SELECT ... FOR UPDATE`, so writers to an item
that **exists** serialise correctly on the row lock. A row that does not exist yet
cannot be locked, so two writers could both take the insert path. The loser's
`ON CONFLICT DO NOTHING` affected no rows, and the code answered that with
`ConditionFailed` regardless of whether the caller ever supplied a condition.

The sibling `put_item` path already draws the distinction correctly, using
`DO UPDATE` for the unconditional case and `DO NOTHING` plus `ConditionFailed` only
for the conditional one. `update_item` had only the second shape.

## The fix

The read-modify-write now sits in a bounded loop with a single locking read at its
head. On the first pass that read is the original one; after a lost race it is the
re-read that resolves the conflict.

Re-reading `FOR UPDATE` is what makes this terminate. `ON CONFLICT DO NOTHING` does
not wait for a concurrent uncommitted inserter, so a plain re-read can see nothing
at all, whereas the locking read blocks until the winner commits (returning its row)
or aborts (returning none, so the insert can be retried). The loser then re-applies
its update expression on top of the winner's item.

Attempts are capped at five. A retry can only be provoked by another writer
committing or rolling back the row, so exhausting the cap means something
pathological and returns an internal error rather than looping.

### Deliberately not `ON CONFLICT DO UPDATE SET item_data = EXCLUDED.item_data`

That is the obvious one-line fix, it is what the issue suggests, and it is wrong
here. The losing writer computed its item from an **empty** base, so `EXCLUDED`
holds only the attributes its own expression set. Overwriting with it silently drops
whatever the winner set, which is exactly the data loss measured above, while
keeping every call successful. It is correct for `PutItem`, where blind overwrite is
the semantics, and it stays in place there and in the transactional helper.

## Testing done

```
cargo test                                  # 422 integration, 0 failed, 0 filtered out
cargo test --no-default-features --features postgres   # 671 workspace, 0 failed
cargo fmt --all --check                     # clean
cargo clippy --all-targets --no-default-features --features postgres -- -D warnings   # clean
```

Run against a live Postgres 16 backend over HTTP, not in process.

Two tests added to `tests/rust/src/concurrency.rs`:

1. **`concurrent_unconditional_upserts_to_one_new_item_all_succeed_and_merge`**
   asserts the merge, not merely the absence of the error. This is deliberate: a
   test that only counted failures would pass against the `EXCLUDED.item_data` fix
   while data was being dropped.
   - **Negative control:** fails on the unpatched build with exactly the reported
     `ConditionalCheckFailedException`.
   - **20 consecutive passes** on the fix, since one green run cannot distinguish
     deterministic from lucky.

2. **`only_one_racing_conditional_create_wins`** is a **regression guard, not a bug
   demonstration**, and the code comment says so because it is easy to misread: it
   passes both before and after. Re-evaluating a condition against the race winner
   produces the same answers either way, because `attribute_exists` fails against
   its empty base before reaching the insert and `attribute_not_exists` still fails
   when re-evaluated against the winner. I initially read this as a second defect
   and that was wrong; the test exists to prove the retry did not perturb either
   outcome.

One thing worth recording so nobody re-investigates it: three `capacity_throttling`
failures appeared mid-investigation. They were an artefact of running that suite
against a warm server immediately after the full run, and were proven unrelated by
an A/B on a fresh server, 7 passed both with and without the change.

## Scope

- **SQLite is unaffected.** Its `update_item` serialises on a process-wide write lock
  and has no lost-race branch.
- **`tx_helpers.rs` is unchanged.** It uses a blind upsert, which is defensible inside
  the serialised transact path, where DynamoDB itself rejects concurrent transactions
  on one item with `TransactionConflictException`.

## Checklist

ADR / RFC: n/a. This restores documented DynamoDB behaviour on an existing
operation; no contract, trait, format or CLI change.

## Breaking changes

None. Callers that previously received `ConditionalCheckFailedException` from an
unconditional `UpdateItem` were receiving an error the real service never returns,
so no correct client can depend on it.

---

### DIFF
--- a/crates/storage-postgres/src/data/update_item.rs
+++ b/crates/storage-postgres/src/data/update_item.rs
@@ -4,7 +4,7 @@
 //! `update_item` implementation for the `PostgreSQL` backend.
 
 use extenddb_core::expression::{self, Expr, ExpressionMaps, UpdateAction};
-use extenddb_core::types::{Item, KeyType, TableKeyInfo};
+use extenddb_core::types::{Item, TableKeyInfo};
 use extenddb_core::validation;
 use extenddb_storage::StreamCapture;
 use extenddb_storage::error::StorageError;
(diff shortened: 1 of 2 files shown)


crates/storage-postgres/src/data/update_item.rs
```diff
@@ -54,153 +54,178 @@ impl PostgresEngine {
...
+                            .execute(&mut *tx)
+                            .await
+                            .map_err(|e| StorageError::Internal(e.to_string()))?;
+                    }
+                }
+                break (old_item, new_item, item, pre_mutation_item);
+            }
+
+            // Row absent, so insert atomically; someone may beat us to it.
+            let inserted = match &sk_parts {
```
jcshepherd, line 191: General comment: Rust seems to allow a great deal of mashing together assignment statements, control statements and execution statements. In this change, I'm finding it's making it very hard to understand the control structures. It's probably not idiomatic Rust, but this would really benefit from breaking it up into some sub-functions that do the actual db access, so that the flow of control is more obvious in the main routine.

jcshepherd (approved):

PR #252, opened 2026-08-11
test: make CREATING-table PutItem rejection immune to the flip race

LeeroyHannigan:
## What

Makes `put_item::put_item_on_creating_table_returns_not_found` immune to the CREATING-to-ACTIVE flip race. The test now attempts against fresh tables (up to 10) and classifies each attempt: a failed PutItem must be `ResourceNotFoundException` (window caught); a successful PutItem is only a failure if `DescribeTable` afterwards still reports `CREATING`, which is proof the data-plane status gate is broken. A benign mid-flight flip retries, and 10 straight misses is a logged skip rather than a failure. Test-only change, no production code touched.

## Why

The test created a table and immediately issued a PutItem, expecting to land inside the CREATING window. CI sets `control_plane_delay_seconds` to 0.05, so that window is 50ms. When request latency exceeds it, the table is already ACTIVE, the put succeeds, and the test fails even though the backend behaved correctly. This is what broke main's MongoDB Integration Tests job after the MongoDB backend merge (run 31515546899, the only red: 425/426): mongo per-op latency under a parallel 426-test run misses the 50ms window. The same race exists on the other backends; they just usually win it. The rewrite keeps the regression coverage (a data plane that serves a CREATING table still fails the test deterministically) while removing the timing false-positive.

Closes # (no standing issue; main CI red on run 31515546899)

## Testing done

Against a mongo:7 single-node replica set with CI's `control_plane_delay_seconds = 0.05`:

- 20/20 consecutive passes of the rewritten test, with the CREATING window caught (put rejected with `ResourceNotFoundException`) on every run.
- Negative control: with the `require_active` gate in `crates/storage-mongodb/src/table_engine.rs` temporarily disabled, the test fails on the CREATING-proof assertion ("PutItem succeeded against a table still reporting CREATING"), confirming it still discriminates the real regression. Gate restored, test passes again.
- `cargo fmt --all -- --check` exit 0.

## Checklist

ADR / RFC: n/a (test-only change; no contract surface touched)

---

### DIFF
--- a/tests/rust/src/put_item.rs
+++ b/tests/rust/src/put_item.rs
@@ -463,42 +463,103 @@ async fn put_item_on_creating_table_returns_not_found() {
         AttributeDefinition, BillingMode, KeySchemaElement, KeyType, ScalarAttributeType,
     };
     let c = client();
-    let table = format!("CreatingInflux_{}", ts());
-    c.create_table()
-        .table_name(&table)
-        .billing_mode(BillingMode::PayPerRequest)
-        .key_schema(
-            KeySchemaElement::builder()
-                .attribute_name(HASH_KEY_S)
-                .key_type(KeyType::Hash)
-                .build()
-                .unwrap(),
-        )
-        .attribute_definitions(
-            AttributeDefinition::builder()
-                .attribute_name(HASH_KEY_S)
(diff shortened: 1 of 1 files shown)


amrith (approved):
approved, lgtm

PR #254, opened 2026-08-12
fix(deps): remediate Rust advisories and drop deprecated rustls-pemfile

robinnsc:
## What

Remediates the open Rust dependency advisories and removes one deprecated dependency,
in a single commit rebased onto current `main`:

- **`crossbeam-epoch` 0.9.18 → 0.9.20** — the fix for RUSTSEC-2026-0204.
- **`anyhow` 1.0.102 → 1.0.103** and **`event-listener` 5.4.1 → 5.4.2** — preferred
  direct-dependency updates. The `event-listener` bump also drops the transitive
  `concurrent-queue` crate entirely.
- **`rustls-pemfile` removed** (deprecated upstream). Both call sites
  (`crates/app/src/init_helpers.rs`, `crates/app/src/manage_http.rs`) migrate to the
  `rustls-pki-types` PEM API (`CertificateDer::pem_slice_iter`) already in the tree,
  with identical error handling and messages.
- **`sqlx` tightened** to `default-features = false` with the explicit feature list,
  plus `derive` (used by the storage backends' `FromRow` types). This trims unused
  default machinery from the dependency graph.

- **`SOFTWARE-LICENSE-NOTICES.html` regenerated** (second commit) with the pinned
  `cargo-about 0.9.0`, since the dependency changes above alter the license
  inventory: `rustls-pemfile` and `concurrent-queue` leave, three versions bump.

Net effect on `Cargo.lock`: zero packages added, `rustls-pemfile` and
`concurrent-queue` removed, three versions bumped. `tests/rust/Cargo.lock` picks up
the matching `anyhow` bump.

## Why

The release runbook blocks the first public `extenddb-postgres` container image on
RUSTSEC-2026-0204: no advisory-affected version may ship in a public image. The
`anyhow`/`event-listener` bumps are the recorded preferred updates from the same
review. Dropping `rustls-pemfile` now, rather than later, avoids shipping the first
public artifact with a dependency that is already deprecated and would force a
post-release lockfile churn to remove.

The branch predates the MongoDB backend landing on `main`; it has been rebased and
the workspace `Cargo.toml` conflict resolved keeping both this PR's hardened `sqlx`
line and `main`'s new `mongodb`/`bson`/`dashmap` dependencies.

## Testing done

- `cargo metadata --locked` — lockfile is consistent with the manifests.
- `cargo check --workspace --all-targets --locked` — clean, including the new
  `extenddb-storage-mongodb` crate from post-rebase `main`.
- `cargo test --workspace --locked` — **769 passed, 0 failed** across all suites.
- `cargo fmt --all --check` — clean.
- `cargo clippy --workspace --all-targets --locked` — clean, zero warnings.
- Lockfile inspected directly: resolves `crossbeam-epoch 0.9.20`, `anyhow 1.0.103`,
  `event-listener 5.4.2`; `comm` over package names confirms nothing was added.
- `devtools/generate-software-license-notices --check` — passes against the
  regenerated notices; spot-checked that `rustls-pemfile`/`concurrent-queue` no
  longer appear and the three bumped versions do.
- Full integration pass (`devtools/run-tests --extenddb --rust-integration --pytest
  --release` against a live server on the pinned PostgreSQL 16.10 image) is running
  at the time of writing — Python suite past 70% with zero failures so far. I will
  post the final counts as a comment when it completes.

Not verified:

- The container smoke test (`ci/smoke-test-container.sh`) has not been re-run for
  this change; it exercises the same binary the workspace and integration suites
  cover, and will run as part of the release process regardless.

## Checklist

- [ ] I have added or updated tests for new functionality — not applicable: no new
      functionality; existing tests cover the migrated PEM-parsing call sites
- [ ] I have updated documentation if behavior changed — not applicable: no behavior
      change

ADR / RFC: n/a — dependency maintenance only; no wire protocol, `Storage` trait,
auth model, on-disk format, or CLI surface change.

## Breaking changes

None. Public behavior is unchanged: the PEM-parsing migration is call-site-for-call-site
with the same error text, and the `sqlx` feature trim keeps every feature the
workspace actually uses (verified by the clean workspace check and test run,
including the SQLite and MongoDB backends).

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,9 +123,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "arc-swap"
@@ -631,15 +631,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cc14f565cf027a105f7a44ccf9e5b424348421a1d8952a8fc9d499d313107789"
 
 [[package]]
-name = "concurrent-queue"
-version = "2.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ca0197aee26d1ae37445ee532fefce43251d24cc7c166799f4d46817f1d3973"
-dependencies = [
- "crossbeam-utils",
-]
-
-[[package]]
 name = "config"
 version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


LeeroyHannigan (approved):
LGTM

PR #256, opened 2026-08-12
ci: add minimal GHCR candidate mirror job

robinnsc:
## What

Adds `.github/workflows/ghcr-mirror.yml`: a dispatch-only job that copies **one
already-verified, commit-addressed candidate** from Docker Hub to
`ghcr.io/extenddb/extenddb-postgres`, byte-for-byte, and does nothing else — no
builds, no Docker Hub publication, no ECR, no version tags, no `latest`, no
signing.

The operator supplies two strictly-validated inputs: the candidate tag
(`sha-<full 40-char commit>`) and the multi-arch index digest they verified.
The job then:

- resolves the Docker Hub tag and **refuses to run if it no longer matches the
  verified digest** (guards against the tag moving between verification and
  mirroring);
- **refuses to overwrite** an existing GHCR tag with a different digest
  (idempotent no-op if the exact digest is already mirrored);
- copies with `skopeo copy --all --preserve-digests`, which fails rather than
  rewrite any manifest, so the GHCR index digest is identical to the verified
  Docker Hub digest;
- re-inspects the published copy: digest equality plus both `linux/amd64` and
  `linux/arm64` present;
- writes a run summary for the release checklist.

Every run requires approval through the protected `ghcr` environment (required
reviewers, prevent self-review, `main` only), which **already exists with those
protections** — so merging this cannot auto-create an unprotected environment.

## Why

GHCR's only acceptable credential is the workflow-scoped `GITHUB_TOKEN`, which
exists only inside an Actions run: the GHCR copy cannot be a manual push, and a
PAT is not an option. This job is the smallest possible bridge for the initial,
otherwise-manual release — the mirroring equivalent of a single manual command,
wrapped in the environment's reviewer gate.

It deliberately duplicates none of #251: that workflow builds, tests, and
publishes the Docker Hub candidate; this one only copies a candidate that a
human has already verified per the release runbook. Folding it into #251's
workflow later (as a job inheriting the gate's validated SHA) is a reasonable
follow-up once the automated process matures.

## Testing done

- YAML parses (`yaml.safe_load`).
- The refusal paths, digest comparisons, and regexes were desk-checked against
  `skopeo` semantics; input validation matches the gate style introduced in
  #251 (untrusted inputs via environment, never interpolated).
- **The workflow itself has not run**: it is dispatch-only, needs the `ghcr`
  environment approval, and there is no published candidate yet. Its first
  real execution is the first candidate mirror — expected to be exercised
  during the release rehearsal.
- Verified via the environments API that `ghcr` exists with the required
  reviewers, prevent-self-review, `main`-only deployment, and zero secrets.

Not verified:

- `skopeo` presence on `ubuntu-latest` is asserted at runtime with an apt
  fallback rather than assumed.
- Organization package-creation settings are not readable without org admin;
  if they are wrong, the first push fails loudly with a package-creation error
  and nothing is published.

## Checklist

- [ ] I have added or updated tests for new functionality — see note below
- [ ] I have updated documentation if behavior changed — no repo docs describe registry mirroring; the release runbook (internal) is the operating document

On tests: there is no workflow test harness in this repo; the gates are the
assertions, and each failure path exits non-zero with a specific error.

ADR / RFC: n/a — release plumbing; no wire protocol, `Storage` trait, auth
model, on-disk format, or CLI surface change.

## Breaking changes

None. Adds one workflow; changes no existing behavior, workflow, or source.
Nothing runs without a manual dispatch plus an environment reviewer's approval.

---

### DIFF
--- a/.github/workflows/ghcr-mirror.yml
+++ b/.github/workflows/ghcr-mirror.yml
@@ -1,0 +1,144 @@
+# Copyright 2026 ExtendDB contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Minimal GHCR candidate mirror.
+#
+# Copies ONE already-verified, commit-addressed candidate from Docker Hub to
+# GHCR, byte-for-byte, and nothing else: no builds, no Docker Hub publication,
+# no ECR, no version tags, no `latest`, no signing. Promotion and signing are
+# manual, recorded runbook steps.
+#
+# Why this exists at all: GHCR's only acceptable credential is the
+# workflow-scoped GITHUB_TOKEN, which exists only inside an Actions run, so
+# the GHCR copy cannot be a manual push (and a PAT is not an option). Every
+# run is human-gated by the protected `ghcr` environment: reviewer approval,
+# main-only, no stored secrets.
+#
+# The digest input is the contract with the manual verification that precedes
+# this job: the operator supplies the digest they verified, and this job
+# refuses to mirror anything else, even if the Docker Hub tag has moved since.
+
+name: ghcr-mirror
+
+on:
+  workflow_dispatch:
+    inputs:
+      candidate_tag:
+        description: 'Verified Docker Hub candidate tag, e.g. sha-<full 40-char commit>'
+        required: true
+      expected_digest:
+        description: 'The verified multi-arch index digest, e.g. sha256:<64 hex>'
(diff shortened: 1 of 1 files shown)


jcshepherd (approved):
LGTM

PR #258, opened 2026-08-12
feat(sqllite): npm dev launcher

LeeroyHannigan:
## What

npm distribution for the ExtendDB dev server, as the first ecosystem of an
ecosystem-neutral launcher contract (pip etc.. follow the same document).

- `packaging/npm/`: the `extenddb` launcher package. `start()` spawns the slim
  dev-mode binary, health-gates it, and returns an endpoint plus credentials
  any AWS SDK accepts. Storage is **file-based by default**
  (`./.extenddb/data.db`, tables survive restarts); `--memory` /
  `{ memory: true }` opts into ephemeral; combining a persistence path with
  memory mode is refused loudly. Binaries resolve from
  `@extenddb/<os>-<arch>` platform packages via `optionalDependencies`, with
  `EXTENDDB_BINARY` and PATH as fallbacks and an actionable error naming all
  resolution steps (including the musl/Alpine trap) when none works.
- `packaging/npm/platform/`: templates for the five platform packages
  (linux x64/arm64, darwin x64/arm64, win32 x64).
- `.github/workflows/npm-publish.yml`: dispatch-only release workflow
  mirroring `release-image.yml` (PR #251) clause for clause: gated
  `npm-vX.Y.Z` tag validation (strict shape, existing tag, ancestor of main,
  version match at the tagged commit), five native runners that each smoke
  test and run the launcher suite against their own binary, and a single
  credentialed job in a reviewer-gated `npm` environment publishing with
  `--provenance` under the `candidate` dist-tag only. Promotion to `latest`
  is a manual dist-tag move. The platform list is derived from
  `optionalDependencies` with loud drift guards. LICENSE and NOTICE are
  staged into every tarball.
- `.github/workflows/launcher.yml`: PR gate running the launcher suite
  against a freshly built dev binary; triggers on `packaging/**` and
  `crates/**` so server changes that break the launcher contract fail at PR
  time.
- `packaging/README.md`: the launcher contract, release flow, one-time
  registry setup, and known limits (hard-kill orphan, musl, shared-file).

## Why

Non-Rust consumers currently need Docker or a manual build to run ExtendDB
locally. `npm install extenddb` gives Node projects a zero-config
DynamoDB-compatible dev server for local dev and CI, with persistent tables
by default and an in-memory mode for ephemeral test runs. The launcher
contract is deliberately ecosystem-neutral so PyPI and Maven wrappers
implement the same document rather than imitating npm code.

Closes # (no linked issue; scoped from the embedded-distribution plan)

## Testing done

- Launcher suite (4 scenarios) against a locally built dev binary:
  `EXTENDDB_BINARY=... node packaging/npm/test/launcher.test.js`. Persistence
  is proven with real data-plane calls (plain-node SigV4, no SDK dev
  dependency): an item written by server lifetime one is read back by
  lifetime two on the same file, and a restarted memory-mode server does not
  know the previous server's table. Negative controls discriminate: pointing
  run two at a different path fails the survival assertion; sabotaging memory
  mode into a shared file fails the loss assertion.
- End-to-end install simulation: ran the publish job's staging logic
  verbatim, `npm pack`ed the launcher and a platform package, installed both
  tarballs into a scratch project, and the launcher resolved the platform
  package and served with `EXTENDDB_BINARY` unset and `extenddb` absent from
  PATH. Control: removing the platform package fails with the actionable
  resolution error.
- Gate script simulated locally with adversarial inputs (command injection,
  path traversal, embedded newline, wrong namespace): all rejected; version
  match verified. Platform-list drift guard proven with a synthetic
  `linux-ppc64` entry.
- Two independent in-depth reviews (launcher internals / process lifecycle,
  and supply-chain / workflow security). All blocking and should-fix findings
  are closed in this branch; action SHA pins verified against the GitHub API;
  `--tag candidate` keeping `latest` unset on first publish verified against
  npm docs and npm/cli#7553.
- `npm pack --dry-run` on both package shapes confirms LICENSE + NOTICE ship
  in every tarball.
- No Rust code is touched (0 `.rs` files in the diff); `cargo fmt --check`
  exit 0 on the branch; the Rust suites are unchanged from main.

## Checklist

- [x] All tests pass (`cargo test --workspace`) — no Rust changes; suites identical to main
- [x] Clippy is clean (`cargo clippy -- -W clippy::pedantic`) — no Rust changes
- [x] Breaking changes are noted below (if any) — none

ADR / RFC: none yet — this adds a new public surface (the npm launcher CLI and
its ecosystem-neutral contract, documented in `packaging/README.md`). It does
not change the server's wire protocol, `Storage` trait, auth model, on-disk
format, or the `extenddb` binary's CLI. If maintainers want the launcher
contract captured as an ADR before merge, I'll extract `packaging/README.md`'s
contract section into one.

---


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,337 @@
...
+      # the finding real, and it would then need a trusted-ref guard.
+      - name: Check out the tagged commit
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4.4.0
+        with:
+          ref: ${{ needs.gate.outputs.sha }}
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.97.0
+
+      - name: Build the slim dev-mode binary
```
LeeroyHannigan, line 214: **CodeQL alert 14** (`actions/cache-poisoning/poisonable-step`, high) lands on this build step.

Two properties make this unexploitable, neither visible to the query:

1. `gate` has already proved this SHA is an ancestor of `origin/main` (`git merge-base --is-ancestor`, line 128), so the code executed here is reviewed code already on main. Poisoning it needs push access to main, which permits strictly worse things directly.
2. Nothing in this workflow reads or writes the Actions cache: no `actions/cache`, no `rust-cache`, no `cache:` input on `setup-node`. There is no cache entry to poison.

Both are recorded in the comment above the checkout, including what has to stay true if this job grows: adding a cache step here would make the finding real and would then need a trusted-ref guard.

Flagging rather than dismissing, since dismissing a security alert on my own PR is your call.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,337 @@
...
+        # `--locked` so the build is exactly the reviewed Cargo.lock.
+        run: cargo build --release --locked --no-default-features --features sqlite,dev-mode -p extenddb
+
+      - name: Smoke test - the native binary must run
+        shell: bash
+        run: |
+          set -euo pipefail
+          "target/release/${BIN_NAME}" --version
+
+      - name: Run the launcher test suite against the built binary
```
LeeroyHannigan, line 225: **CodeQL alert 15** (`actions/cache-poisoning/poisonable-step`, high) lands on this test step, same cause as alert 14 on the build step above.

Two properties make this unexploitable, neither visible to the query:

1. `gate` has already proved this SHA is an ancestor of `origin/main` (`git merge-base --is-ancestor`, line 128), so the code executed here is reviewed code already on main. Poisoning it needs push access to main, which permits strictly worse things directly.
2. Nothing in this workflow reads or writes the Actions cache: no `actions/cache`, no `rust-cache`, no `cache:` input on `setup-node`. There is no cache entry to poison.

Both are recorded in the comment above the checkout, including what has to stay true if this job grows: adding a cache step here would make the finding real and would then need a trusted-ref guard.

Flagging rather than dismissing, since dismissing a security alert on my own PR is your call.

LeeroyHannigan:
Three changes since this PR was last looked at, plus an explanation of the red CodeQL check.

**The launcher is now `@extenddb/dev`, not `extenddb`.** npm refuses the unscoped name outright:

```
403 Forbidden - PUT https://registry.npmjs.org/extenddb
Package name too similar to existing package extend
```

`extend` is the jQuery.extend port. The check is server-side on PUT, so it cannot be detected before attempting a publish, and the alternatives were a support appeal or a scoped name. The CLI command is unaffected and stays `extenddb` (`npx extenddb start`), because it comes from the `bin` key. The five `@extenddb/<os>-<arch>` packages publish fine, since the scope makes them unambiguous.

**Publishing authenticates with trusted publishing (OIDC), so no npm token exists in the repo.** Both `NODE_AUTH_TOKEN` lines are gone and npm is bumped to 11 before publishing (the OIDC exchange needs >= 11.5.1; the Node 22 image ships npm 10 and would fail with `ENEEDAUTH`). `environment: npm` now does more than hold a secret: npm matches the environment name as part of the trust relationship, so a run outside it cannot mint a publish credential. Configured per package on all six as `ExtendDB` / `extenddb` / `npm-publish.yml` / `npm`, publish only, not stage publish.

**The launcher now versions with the server**, off the same `vX.Y.Z` tag the container release uses, instead of its own `npm-v*` namespace. The platform packages ship the server binary, so a server release republishes all six anyway, and the launcher pins `optionalDependencies` to its own version. An independent number could only ever say which launcher you have, never which server is inside it. Same convention as `esbuild` and `@biomejs/biome`. The gate now validates the tag against the workspace `Cargo.toml` and also refuses a drifted `packaging/npm/package.json`. Accepted cost, written into `packaging/README.md`: a launcher-only fix rides the next server release.

**On the failing CodeQL check:** two high `actions/cache-poisoning/poisonable-step` alerts, both on this workflow, both present since the file was added rather than introduced by the above. I've commented inline at each. Short version: `workflow_dispatch` runs in the default branch's cache scope while checking out a dispatch-supplied SHA, but `gate` has already proved that SHA is an ancestor of `origin/main`, and nothing in the workflow touches the Actions cache at all, so there is no entry to poison. The reasoning is recorded at the flagged steps. I have deliberately not dismissed them: that call is yours.

Everything else is green, 17 checks including all three MongoDB jobs.

LeeroyHannigan:
Cleaned this up for review (`93d972f`):

**Manifest bumped to 0.1.7.** `npm-publish`'s gate refuses a drifted manifest outright (`packaging/npm/package.json at <sha> says 0.1.3, expected 0.1.7`), and the workspace is going to 0.1.7 in #296. `optionalDependencies` moved `0.1.0` -> `0.1.7` in the same edit: the publish step rewrites them anyway, so it isn't load-bearing, but a checked-in pin that no published platform package has ever carried just misleads whoever reads the manifest.

**Updated from main** (was 52 commits behind, predating the whole vector search merge and the dev image's 18080 port). Clean merge, no conflicts.

**The two CodeQL alerts are dismissed as false positives.** Verified rather than assumed: `grep -i cache` on `npm-publish.yml` returns only the comment block explaining that nothing in the workflow reads or writes the Actions cache — no `actions/cache`, no `rust-cache`, no `cache:` on `setup-node`. There is no cache entry to poison, and the gate proves the tag is an ancestor of `main` before any checkout runs. Same rule and same disposition as the pair on #281. The in-workflow comment already says "adding a cache step here would make the finding real", which stays true.

**Verified with a real dev-mode binary**, not just static review: launcher suite 4/4 scenarios passed, and the CLI serves correctly in memory mode (endpoint, seeded example credential, `:memory: (ephemeral)`).

Worth noting the 18080 dev-image port change does **not** affect the launcher: it sets `EXTENDDB__SERVER__PORT` to an ephemeral port explicitly (`index.js:136`), so there's no coupling.

Sequencing after this merges: #296 lands the 0.1.7 workspace bump -> tag `v0.1.7` on main -> dispatch `npm-publish` with that tag, which publishes all six packages under the **`candidate`** dist-tag with provenance. Promotion to `latest` stays manual: six `npm dist-tag add` commands from a maintainer machine. Until then `npm i @extenddb/dev` keeps resolving to the 0.0.0 placeholder.

packaging/npm/index.js
```diff
@@ -0,0 +1,201 @@
...
+        );
+        return;
+      }
+      reject(err);
+    });
+    child.on("exit", (code) =>
+      reject(new Error(`extenddb exited during startup (code ${code}): ${stderrTail}`))
+    );
+  });
+  await Promise.race([waitHealthy(endpoint, options.startupTimeoutMs ?? 15000), spawned]);
```
robinnsc, line 176: The timeout path here leaks the child. If `waitHealthy` rejects while the server is alive but slow, `start()` throws and the caller has no handle to the process, so it keeps running and holding its port and the SQLite lock until the caller's process exits. For a test runner that catches the error and carries on, the next `start()` against the same `dbPath` then fails on the lock error, which the README describes as a user mistake rather than the leak it actually is.

Killing the child before rethrowing on any post spawn startup failure closes it:

```js
try {
  await Promise.race([waitHealthy(endpoint, options.startupTimeoutMs ?? 15000), spawned]);
} catch (err) {
  child.kill("SIGKILL");
  throw err;
}
```

packaging/README.md
```diff
@@ -0,0 +1,138 @@
...
+   npm dist-tag add @extenddb/linux-x64@X.Y.Z latest        # and the other four
+   ```
+
+One-time registry and repo setup, in order: create the npm org `extenddb`.
+The launcher ships as the scoped `@extenddb/dev` rather than an unscoped
+`extenddb`, because npm's package-name similarity filter refuses the latter
+against the long-established `extend` package. Then create a granular
+automation token scoped to the `@extenddb` scope with publish permission only,
+and create the GitHub environment `npm` with deployment branch rule `main`,
+required reviewers, and the token as `NPM_TOKEN`.
```
robinnsc, line 125: This paragraph disagrees with the workflow it documents. `npm-publish.yml` says the environment has no secrets because authentication is npm trusted publishing over OIDC, and the publish job indeed mints an id token and never reads a stored credential. Following this text instead creates a long lived automation token that nothing uses, which is exactly the credential the workflow design went out of its way not to have. I think this predates the switch to trusted publishing and wants rewriting to describe the trusted publisher setup (the workflow header already has the exact field values).

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,337 @@
...
+      #
+      # Keep both true if this job grows: adding a cache step here would make
+      # the finding real, and it would then need a trusted-ref guard.
+      - name: Check out the tagged commit
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4.4.0
+        with:
+          ref: ${{ needs.gate.outputs.sha }}
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.97.0
```
robinnsc, line 212: This is the one action in the release path that is not pinned to a SHA. `1.97.0` here is a mutable tag, and this step runs in the five jobs whose output is the binary actually published to npm, so a retagged action lands directly in release artifacts. That is the highest value target in this workflow, and it sits oddly next to checkout, setup-node, and both artifact actions all being SHA pinned. Worth pinning like the others (same applies to `rust-cache@v2` over in launcher.yml, though that one runs in a credential free PR gate so the stakes are much lower).

robinnsc (approved):
Read the launcher, the CLI, the test suite, and both workflows. This is a well designed distribution: the candidate first flow, keeping registry credentials out of every job that executes a freshly built binary, the dispatch only trigger with the ancestry gate, the drift guards derived from optionalDependencies, and persistence tests that actually discriminate (the negative controls in the description are the kind of thing most packaging PRs skip). Approving with three findings inline, one of them a real behavior bug, all small fixes.

Two things I noticed but consider fine as they are:

1. `freePort()` has the usual bind, close, reuse race: between the probe closing and the child binding, another process can take the port, and in the worst case `waitHealthy` gets a 200 from an unrelated local server. Inherent to the ephemeral port launcher pattern and vanishingly unlikely in practice; a code comment acknowledging it would do.
2. A publish failure mid loop leaves a partial candidate release, and npm version immutability means the same version cannot be rerun. The candidate dist tag means users never see it and recovery is a version bump, so this is acceptable, but a line in the README release section would save the first maintainer who hits it some head scratching.

PR #260, opened 2026-08-13
fix: merge UpdateTable AttributeDefinitions instead of replacing them

### LINKED ISSUE #259: [Bug] Creating a GSI causes base-table GetItem to return the wrong item for composite keys
## Describe the bug

On a table with a composite primary key (`pk` HASH + `sk` RANGE), `GetItem` works correctly until a GSI is added to the already-populated table. After the GSI becomes `ACTIVE`, a strongly consistent `GetItem` against the **base table** can ignore the requested sort key and return a different item under the same partition key.

The request succeeds with HTTP 200 and returns a complete but incorrect item. There is no timeout, missing item, or API error.

In the minimal test:

- 100 items have the same `pk` and distinct `sk` values (`sk-00000` through `sk-00099`).
- Before adding the GSI, 100/100 serial strongly consistent reads return the exact requested item.
- One `ALL` projection GSI is added and allowed to reach `ACTIVE`.
- No writes occur after GSI creation.
- A request for `sk-00099` returns the full item for `sk-00000`.
- In a separate full verification run, only 1/100 reads was correct; the other 99 returned the item for `sk-00000`.

This test uses one process, one client, serial `PutItem`/`GetItem` calls, no threads, no multiprocessing, no `BatchWriteItem`, and no restart.

## To reproduce

The client credentials and endpoint are supplied through environment variables; no special client configuration is required.

```python
import os
import time

import boto3


endpoint = os.environ["EXTENDDB_ENDPOINT"]
table = f"gsi-getitem-repro-{int(time.time())}"

ddb = boto3.client(
    "dynamodb",
    endpoint_url=endpoint,
    region_name="us-east-1",
    aws_access_key_id=os.environ["AWS_ACCESS_KEY_ID"],
    aws_secret_access_key=os.environ["AWS_SECRET_ACCESS_KEY"],
    verify=False,  # local test certificate only
)

ddb.create_table(
    TableName=table,
    AttributeDefinitions=[
        {"AttributeName": "pk", "AttributeType": "S"},
        {"AttributeName": "sk", "AttributeType": "S"},
    ],
    KeySchema=[
        {"AttributeName": "pk", "KeyType": "HASH"},
        {"AttributeName": "sk", "KeyType": "RANGE"},
    ],
    BillingMode="PAY_PER_REQUEST",
)


def wait_active(index_name=None):
    while True:
        description = ddb.describe_table(TableName=table)["Table"]
        table_active = description["TableStatus"] == "ACTIVE"
        indexes = {
            index["IndexName"]: index
            for index in description.get("GlobalSecondaryIndexes", [])
        }
        index_active = (
            index_name is None
            or indexes.get(index_name, {}).get("IndexStatus") == "ACTIVE"
        )
        if table_active and index_active:
            return
        time.sleep(0.5)


wait_active()

expected = []
for i in range(100):
    item = {
        "pk": {"S": "shared-serial"},
        "sk": {"S": f"sk-{i:05d}"},
        "f01": {"S": f"gpk-{i % 10:02d}"},
        "f02": {"S": f"gsk-{i:05d}"},
    }
    # pk + sk + f01..f18 = 20 attributes, matching the observed case.
    for field in range(3, 19):
        item[f"f{field:02d}"] = {"S": f"payload-{i:05d}"}
    ddb.put_item(TableName=table, Item=item)
    expected.append(item)

# All reads are correct before creating the GSI.
for item in expected:
    response = ddb.get_item(
        TableName=table,
        Key={"pk": item["pk"], "sk": item["sk"]},
        ConsistentRead=True,
    )
    assert response.get("Item") == item
print("before GSI: 100/100 correct")

ddb.update_table(
    TableName=table,
    AttributeDefinitions=[
        {"AttributeName": "f01", "AttributeType": "S"},
        {"AttributeName": "f02", "AttributeType": "S"},
    ],
    GlobalSecondaryIndexUpdates=[{
        "Create": {
            "IndexName": "gsi01",
            "KeySchema": [
                {"AttributeName": "f01", "KeyType": "HASH"},
                {"AttributeName": "f02", "KeyType": "RANGE"},
            ],
            "Projection": {"ProjectionType": "ALL"},
        }
    }],
)
wait_active("gsi01")

# No writes are performed after creating the GSI.
requested = expected[99]
actual = ddb.get_item(
    TableName=table,
    Key={"pk": requested["pk"], "sk": requested["sk"]},
    ConsistentRead=True,
)["Item"]

print("requested:", requested["pk"], requested["sk"], requested["f18"])
print("returned: ", actual["pk"], actual["sk"], actual["f18"])
assert actual == requested
```

## Expected behavior

Adding a GSI must not change the result of a base-table `GetItem`. The request specifies the complete primary key and uses `ConsistentRead=True`, so it must return exactly the item identified by both `pk` and `sk`.

Expected output:

```text
before GSI: 100/100 correct
requested: {'S': 'shared-serial'} {'S': 'sk-00099'} {'S': 'payload-00099'}
returned:  {'S': 'shared-serial'} {'S': 'sk-00099'} {'S': 'payload-00099'}
```

## Actual behavior

```text
before GSI: 100/100 correct
requested: {'S': 'shared-serial'} {'S': 'sk-00099'} {'S': 'payload-00099'}
returned:  {'S': 'shared-serial'} {'S': 'sk-00000'} {'S': 'payload-00000'}
AssertionError
```

The same wrong item is returned by a fresh process and client.

Direct inspection of PostgreSQL showed that all 100 base-table rows have distinct, correct physical sort keys and matching `item_data`. The GSI relation also contains 100 distinct, correct index entries. This suggests the stored data and GSI backfill are correct, while the base-table `GetItem` path after the metadata change is selecting by `pk` without correctly applying `sk`.

## Impact

This is a silent correctness failure at the API boundary:

- A successful request returns another item's data rather than an error.
- `ConsistentRead=True` does not prevent it.
- Retries cannot correct a deterministic wrong result.
- Ordinary availability/error-rate monitoring will treat the request as successful.
- Applications can make decisions or mutations based on the wrong entity.
- Depending on the key design, it can disclose data belonging to another sort-key-scoped entity.

This is not normal GSI eventual consistency: the operation is `GetItem` on the base table, the GSI is already `ACTIVE`, no post-GSI writes occur, and the underlying stored rows are correct.

## Environment

- ExtendDB version: built from source at v0.1.3 
- Storage backend: PostgreSQL 16.10
- Deployment method: official/containerized PostgreSQL deployment
- Client: Python 3.12 / boto3
- Authentication: builtin
- Test mode: single process, serial requests

## Possibly related issues

- #238: GSI pagination with duplicate index sort keys on a composite base table
- #145: LSI/GSI pagination fails to use the base table sort key as a tie-breaker

Those issues also involve composite-key handling in index-related paths, but neither describes a base-table `GetItem` returning another item after GSI creation, so this appears to be a separate correctness bug.

## Checklist

- [x] I can reproduce this on two independently created tables

ivyxjc:
## Summary

Fixes #259 — after creating a GSI on a composite-key table, base-table `GetItem` (and `UpdateItem`/`DeleteItem`) ignored the sort key and silently targeted an arbitrary item under the same partition key.

## Root cause

`UpdateTable` persisted the request's `AttributeDefinitions` by **replacing** the stored definitions wholesale. Per DynamoDB semantics that list only needs to describe the attributes referenced by the new index, so the base table's own key attribute definitions (`pk`/`sk`) were dropped from the catalog. `sk_info()` could then no longer resolve the sort key type and returned `None`, causing key lookups to degrade to `WHERE pk = $1` — a partition-key-only query returning an arbitrary row.

## Fix

Merge the incoming definitions into the stored ones by attribute name (replace same-name entries, append new ones), in both the PostgreSQL and SQLite backends. The merged set is also used for GSI data-table creation and backfill.

The MongoDB backend does not have this bug (it never updates the stored definitions on UpdateTable), though it has a separate gap — new GSI key attribute definitions are not recorded at all — which is out of scope here.

## Verification

- The issue's reproduction script against PostgreSQL 16: **1/100** correct reads on unpatched v0.1.3 → **100/100** with this fix; catalog `attribute_definitions` now contains the full union (`pk, sk, f01, f02`).
- New regression test `tests/test_gsi_attribute_definitions.py` covering GetItem correctness, DeleteItem key targeting, and DescribeTable reporting the merged definitions.
- `cargo test --workspace`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` all pass.

Existing tables already corrupted by this bug can be repaired without rebuilding: the data rows and GSI entries are intact; only `tables.attribute_definitions` in the catalog needs the missing key definitions restored (e.g. via another UpdateTable carrying the full definitions after this fix, or direct SQL).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/storage-postgres/src/update_table.rs
+++ b/crates/storage-postgres/src/update_table.rs
@@ -265,6 +265,9 @@ impl PostgresEngine {
         }
 
         // Apply GSI updates (create/delete).
+        let mut merged_attr_defs: Vec<AttributeDefinition> =
+            serde_json::from_value(ad_json.clone())
+                .map_err(|e| StorageError::Internal(e.to_string()))?;
         let mut created_index_ids: Vec<String> = Vec::new();
         let mut deleted_index_ids: Vec<String> = Vec::new();
         if let Some(updates) = &input.global_secondary_index_updates {
@@ -344,12 +347,27 @@ impl PostgresEngine {
                 }
             }
 
-            // Update attribute_definitions on the table if new ones were provided.
+            // Merge new attribute definitions into the existing ones.
+            // UpdateTable's AttributeDefinitions only needs to describe the
+            // attributes referenced by the new indexes; replacing the stored
+            // definitions wholesale would drop the table's own key attribute
+            // definitions and break sort-key handling on the base table
+            // (issue #259).
             if let Some(new_attr_defs) = &input.attribute_definitions {
-                let ad_json = serde_json::to_value(new_attr_defs)
+                for new_def in new_attr_defs {
+                    if let Some(existing) = merged_attr_defs
+                        .iter_mut()
+                        .find(|d| d.attribute_name == new_def.attribute_name)
+                    {
+                        *existing = new_def.clone();
+                    } else {
+                        merged_attr_defs.push(new_def.clone());
+                    }
+                }
+                let merged_json = serde_json::to_value(&merged_attr_defs)
                     .map_err(|e| StorageError::Internal(e.to_string()))?;
                 sqlx::query("UPDATE tables SET attribute_definitions = $1 WHERE account_id = $2 AND table_name = $3")
-                    .bind(&ad_json)
+                    .bind(&merged_json)
                     .bind(account_id)
                     .bind(&input.table_name)
                     .execute(&mut *tx)
@@ -368,10 +386,7 @@ impl PostgresEngine {
                 .map_err(|e| StorageError::Internal(e.to_string()))?;
             let base_attr_defs: Vec<AttributeDefinition> = serde_json::from_value(ad_json.clone())
                 .map_err(|e| StorageError::Internal(e.to_string()))?;
-            let effective_attr_defs = input
-                .attribute_definitions
-                .as_deref()
-                .unwrap_or(&base_attr_defs);
+            let effective_attr_defs: &[AttributeDefinition] = &merged_attr_defs;
 
             let mut create_idx = 0usize;
             let mut delete_idx = 0usize;
--- a/crates/storage-sqlite/src/update_table.rs
+++ b/crates/storage-sqlite/src/update_table.rs
@@ -215,6 +215,8 @@ impl SqliteEngine {
         }
 
         // GSI create/delete.
+        let mut merged_attr_defs: Vec<AttributeDefinition> =
+            serde_json::from_str(&ad_json).map_err(|e| StorageError::Internal(e.to_string()))?;
         let mut created: Vec<String> = Vec::new();
         let mut deleted: Vec<String> = Vec::new();
         if let Some(updates) = &input.global_secondary_index_updates {
@@ -279,8 +281,24 @@ impl SqliteEngine {
                     deleted.push(del_id);
                 }
             }
+            // Merge new attribute definitions into the existing ones.
+            // UpdateTable's AttributeDefinitions only needs to describe the
+            // attributes referenced by the new indexes; replacing the stored
+            // definitions wholesale would drop the table's own key attribute
+            // definitions and break sort-key handling on the base table
+            // (issue #259).
             if let Some(new_attr_defs) = &input.attribute_definitions {
-                let j = serde_json::to_string(new_attr_defs)
+                for new_def in new_attr_defs {
+                    if let Some(existing) = merged_attr_defs
+                        .iter_mut()
+                        .find(|d| d.attribute_name == new_def.attribute_name)
+                    {
+                        *existing = new_def.clone();
+                    } else {
+                        merged_attr_defs.push(new_def.clone());
+                    }
+                }
+                let j = serde_json::to_string(&merged_attr_defs)
                     .map_err(|e| StorageError::Internal(e.to_string()))?;
                 update_col(
                     &mut tx,
@@ -303,7 +321,7 @@ impl SqliteEngine {
                 .map_err(|e| StorageError::Internal(e.to_string()))?;
             let base_ad: Vec<AttributeDefinition> = serde_json::from_str(&ad_json)
                 .map_err(|e| StorageError::Internal(e.to_string()))?;
-            let effective_ad = input.attribute_definitions.as_deref().unwrap_or(&base_ad);
+            let effective_ad: &[AttributeDefinition] = &merged_attr_defs;
 
             let mut ci = 0usize;
             let mut di = 0usize;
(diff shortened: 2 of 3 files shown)


LeeroyHannigan:
Thank you for this, and for the report in #259 that it came from. That report was a model of
what a bug report should be: a self-contained reproduction, the observation that the physical
rows and the GSI relation were both intact, and the conclusion that only the read path could
be at fault. You had the root cause before anyone here looked at it, and your fix reaches the
same place we did, including using the merged set for the index DDL, which is the part that is
easy to overlook.

I want to be straightforward about what we did with it, because you deserve that rather than
silence on an open PR.

We are trying to publish our first container image today, and this bug is on that path, so we
were working to a deadline that has nothing to do with you. You are under no obligation to
take review feedback or to keep iterating on a contribution, and it would not have been fair
to hold your PR open waiting for rounds of it. So instead of asking you to extend this, we
built on it: #261 carries your fix forward with a wider scope and a couple of corrections, and
your test file is in it unchanged, with you credited as co-author on the commit and in the PR
description.

What #261 adds beyond this PR: the MongoDB backend, which as you correctly noted does not have
#259, but never persisted the new index's key attributes at all, so a GSI created through
UpdateTable resolved its own sort key to nothing and behaved as hash-only; a `TableKeyInfo`
check so this class of metadata loss fails loudly instead of quietly returning another item;
and an `extenddb migrate` repair, because a table already damaged by the old code stays
damaged after upgrading and its stored sort key type has to be recovered from the data table's
PRIMARY KEY.

One correction worth passing back, since it is the same shape as the bug you found. On a
same-name redeclaration this PR does `*existing = new_def.clone()`, so the incoming type wins.
To be clear, this is not about supporting a key type change: UpdateTable cannot do that, and
nobody would send one deliberately. It is about what happens when a request carries a
definition it should not, say a client re-sending the full CreateTable attribute list from a
changed model, or a schema tool regenerating it. We measured that case: a table created with
`pk S` whose UpdateTable supplies `[f01 S, f02 S, pk N]` alongside a GSI create is accepted,
and the stored definitions afterwards still read `pk S`. The service neither applies the
conflicting entry nor rejects the request, so `AttributeDefinitions` is not a redefinition
mechanism. It matters because if such an entry names the *sort* key, overwriting it moves
every keyed read to a different physical column, and does so silently rather than raising a
ValidationException. #261 keeps the stored definition.

Genuinely, thank you. Finding and diagnosing a silent wrong-answer bug in someone else's
database is real work, and this one was well worth fixing. If you would like to look over
#261 you are very welcome to, with no obligation either way.

ivyxjc:
Thanks for the thoughtful writeup, and for carrying the fix forward rather than letting it sit — shipping it in #261 with the MongoDB gap, the loud-failure check, and the migrate repair is a strictly better outcome than this PR alone. The co-author credit is appreciated.

As someone who needs to run DynamoDB-compatible workloads in self-hosted environments, thank you for the work you're putting into this project.

I'll take a look at #261. Feel free to close this PR in favor of it.

PR #261, opened 2026-08-13
fix: update table must merge attribute definitions

### LINKED ISSUE #259: [Bug] Creating a GSI causes base-table GetItem to return the wrong item for composite keys

LeeroyHannigan:
> **Credit.** Reported by @ivyxjc in #259, with a reproduction script and the key
> observation that the stored rows and the GSI relation were both correct so only the read
> path was at fault. @ivyxjc also independently diagnosed the same root cause and proposed a
> fix in #260, which reaches the same conclusion for the two SQL backends. This PR supersedes
> it with wider scope (MongoDB, the invariant check, and the `migrate` repair) and different
> conflicting-redeclaration semantics, explained in #260.

# PR body: fix/update-table-must-merge-attribute-definitions

## What

Two commits.

**1. `UpdateTable` merges attribute definitions instead of replacing them**, on all three backends.

- `crates/storage/src/util/key.rs`: new `merge_attribute_definitions`, a union keyed on attribute name.
- `crates/storage-postgres/src/update_table.rs`, `crates/storage-sqlite/src/update_table.rs`: merge into the set already read under the write lock these methods hold, so the read-modify-write is atomic against a concurrent `UpdateTable` with no new locking. Both also now build the index DDL from the merged set rather than the request's subset.
- `crates/storage-mongodb/src/table_engine.rs`: persist the merged set, which this backend never wrote on this path at all.
- `crates/core/src/types/key_schema.rs`: `TableKeyInfo::validate_sort_key_definitions`, called where each backend loads key info, refuses catalog metadata whose base key schema names a RANGE attribute with no matching definition.

**2. `extenddb migrate` repairs tables already damaged by the pre-fix write path.** The merge stops new damage; a table damaged by an older build stays damaged and, with the invariant check in place, now fails its reads loudly instead of silently returning another item.

- `recover_sort_key_definitions` (`crates/storage/src/util/key.rs`) reads the lost scalar type back out of the physical column name: the data table's sort key columns were created from the original definition and are named by `sk_column_n`, so `sk_s` means the first sort key was `S` and `sk3_n` means the third was `N`. A recovery, not a guess.
- `Bootstrapper::repair_lost_sort_key_definitions` applies it, implemented for PostgreSQL (`information_schema.columns`) and SQLite (`PRAGMA table_info`). The default no-op is correct for MongoDB, which never wrote these definitions and so never destroyed any.
- `cmd_migrate` calls it under the migration lock the command already takes, deliberately before the up-to-date early return: the damage is in a row's contents rather than the schema, so a deployment whose catalog version is already current is exactly the case that needs checking.

Two conditions are reported rather than fixed. A sort key with no matching physical column is left alone, because a guessed type would resolve the key to the wrong column. The partition key definition, dropped by the same write, is not recoverable from the schema (the `pk` column is always `TEXT`) and is not needed for correctness, since partition key values are encoded from the key schema alone; only index key type validation is weaker without it. It is reported on every run until a human restores it.

## Why

Closes #259.

A GSI added to a populated composite-key table made a strongly consistent base-table `GetItem` return a different item under the same partition key, with HTTP 200 and no error.

An `UpdateTable` request carries only the attribute definitions it needs, which for a GSI creation is that index's key attributes. Both SQL backends wrote that set over the stored one, so a request supplying `[f01, f02]` deleted the base table's own `pk` and `sk` definitions. `sk_info()` derives the sort key's physical column from its attribute definition, so with the definition gone it returned `None` even though the key schema still named a RANGE element, and `get_item_impl` took its partition-only branch:

```
before: SELECT item_data FROM t WHERE pk = $1 AND sk_s = $2
after:  SELECT item_data FROM t WHERE pk = $1
```

The stored rows and the GSI backfill were both correct, as the reporter said; only the metadata describing them was wrong.

MongoDB never persisted attribute definitions here, so its base table kept working but a created index's key attributes were absent from the catalog and the index resolved its own sort key to nothing. Measured on the pre-fix binary, a GSI query qualified by both hash and sort key returned all 8 items in the partition instead of 1.

Merge semantics are measured against real DynamoDB (us-east-1, 2026-08-13) on a table created with `[pk S, sk S]`: supplying only `[f01, f02]` is accepted and the next `DescribeTable` reports all four; supplying the full set does not duplicate anything; and re-declaring `pk` as `N` while the table holds it as `S` is accepted with the stored type
left as `S`, so a conflicting redeclaration keeps the existing definition rather than raising an error the service does not raise.

## Testing done

All integration runs via `devtools/run-tests --extenddb --rust-integration`.

Three new integration tests, green on all three backends with 0 filtered out:

```
postgres  test result: ok. 3 passed; 0 failed; 0 filtered out
sqlite    test result: ok. 3 passed; 0 failed; 0 filtered out
mongodb   test result: ok. 3 passed; 0 failed; 0 filtered out
```

Negative controls on pre-fix binaries, per backend:

- postgres and sqlite fail the reported read, `GetItem returned a different sort key than the one requested (sk-00001)`, and fail the merge assertion;
- mongodb passes the base-table read, as expected since it was never affected, and fails the merge assertion (`["pk","sk"]` vs `["f01","f02","pk","sk"]`) and the index sort-key  assertion (8 rows returned where 1 was expected).

A second control keeps `validate_sort_key_definitions` while reverting only the merge: the read then fails with a 500 instead of returning another item's data, which is the point of the check.

Full suites with the fix: postgres 426/429, sqlite 425/429, 0 filtered out in both. The non-passing tests are pre-existing and unrelated:

- three `capacity_throttling` tests fail identically on the pre-fix binary; they need throttling settings this ad-hoc config does not enable;
- `binary_sort_key::query_binary_sort_key_begins_with_empty_prefix` on sqlite is the `vec![0xFF; prefix.len() + 1]` upper bound at `crates/storage-sqlite/src/data/query.rs:128`, untouched by this change.

Unit tests: 375 (`extenddb-core`) and 42 (`extenddb-storage`) passing, 0 filtered out. Covering the merge (base keys preserved when only new attributes are sent, idempotent on a full set, conflicting redeclaration keeps the stored type), the invariant check (rejects a RANGE key with no definition, accepts hash-only tables, ignores index key schemas), and the
recovery (S/N/B columns, multiple sort keys matched by position, intact metadata left alone, hash-only tables ignored, and no type invented when the column is absent).

The repair is verified end to end against a SQLite deployment holding one healthy table and one corrupted by a genuine pre-fix binary (`attribute_definitions` = `[f01, f02]`, both base key definitions gone, and the reported wrong-item read reproduced on it):

```
--- Checking table key metadata...
  Restored sort key definition: Gsi259_get_...: sk (S)
  NEEDS ATTENTION: Gsi259_get_...: partition key definition(s) [pk] are absent; ...
```

The healthy table was left untouched, a second run reported no further repair, and a fresh deployment reports `Table key metadata intact.` The partition key note keeps appearing on later runs, which is what an unresolved condition should do. The read-path consequence of correct metadata is covered by the integration tests above rather than re-proved on the
repaired table.

The GSI query test deliberately gives every item the same index hash key. Without that the assertion passed against the broken MongoDB path and proved nothing.

## Checklist

ADR / RFC: n/a. No wire, trait, auth, on-disk-format or CLI change. The `Storage` trait is untouched; `merge_attribute_definitions` is a new helper in the shared storage crate and `TableKeyInfo` gains one method. The catalog schema is unchanged, so `CATALOG_VERSION` is not bumped and no migration is needed: the fix corrects what
`UpdateTable` writes into an existing column.

Documentation is unchecked deliberately. The corrected behaviour is what `docs/differences-from-dynamodb.md` already implies by not listing a divergence here, so there is nothing to remove; one observed parity detail is left unimplemented and noted under Breaking changes rather than documented as intentional.

## Breaking changes

None for clients. `UpdateTable` requests that were previously destroying metadata now preserve it, and `DescribeTable` reports the union where it previously reported only the last request's attributes, which is what real DynamoDB reports.

Two notes for reviewers:

- A table already damaged by a pre-fix `UpdateTable` fails its keyed reads with a 500 until `extenddb migrate` is run, rather than silently returning the wrong item as before. That is the better failure, but it is a behaviour change for an already-damaged table, and it means an upgrade should run `migrate` before taking traffic. On MongoDB there is nothing to repair.
- Out of scope, noted for the record: on MongoDB, tables whose GSIs were created before this change still have no persisted definitions for those index key attributes, so those indexes continue to behave as hash-only until the index is recreated. That is the pre-existing MongoDB gap rather than #259, and repairing it needs the index key types rather than the base table's.
- Measured but not implemented: real DynamoDB does not persist attribute definitions that no key uses. Supplying `[f01, f02, zz9]` for a GSI keyed on `f01`/`f02` is accepted and `zz9` is absent from the next `DescribeTable`. We keep it. That is a cosmetic `DescribeTable` divergence with no effect on reads or writes, so it is out of scope here.

---

### DIFF
--- a/crates/app/src/cmd_migrate.rs
+++ b/crates/app/src/cmd_migrate.rs
@@ -107,9 +107,46 @@ async fn apply_migrations(
         println!("  Pending: {}", data_pending.join(", "));
     }
 
-    if !catalog_pending && data_pending.is_empty() {
+    // Detect table metadata damaged by the pre-fix UpdateTable (#259), which
+    // deleted a table's own key attribute definitions when a GSI was added.
+    //
+    // Detection is read-only and runs even when the catalog version is already
+    // current: the damage is in a row's contents, not in the schema, so an
+    // up-to-date deployment is exactly the case that needs checking. Nothing is
+    // written until --yes has been given, below.
+    println!("--- Checking table key metadata...");
+    let detected = bootstrap
+        .repair_lost_sort_key_definitions(false)
+        .await
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+    let repair_pending = !detected.repaired.is_empty();
+    if !repair_pending && detected.needs_attention.is_empty() {
+        println!("  Table key metadata intact.");
+    } else {
+        for line in &detected.repaired {
+            println!("  Repairable: {line}");
+        }
+        for line in &detected.needs_attention {
+            println!("  NEEDS ATTENTION: {line}");
+        }
+    }
+
+    if !catalog_pending && data_pending.is_empty() && !repair_pending {
         println!();
-        println!("Everything is up to date (catalog version {expected}). No migrations needed.");
+        if detected.needs_attention.is_empty() {
+            println!(
+                "Everything is up to date (catalog version {expected}). No migrations needed."
+            );
+        } else {
+            // Nothing is automatically applicable, but saying "everything is up
+            // to date" straight after a NEEDS ATTENTION line would be
+            // contradictory: those tables require a human.
+            println!(
+                "No migrations needed (catalog version {expected}), but {} item(s) above \
+                 need manual attention.",
+                detected.needs_attention.len()
+            );
+        }
         return Ok(());
     }
 
@@ -120,6 +157,12 @@ async fn apply_migrations(
         }
         if !data_pending.is_empty() {
             what.push(format!("data migrations [{}]", data_pending.join(", ")));
+        }
+        if repair_pending {
+            what.push(format!(
+                "table key metadata repairs [{}]",
+                detected.repaired.join("; ")
+            ));
         }
         anyhow::bail!(
             "--yes is required to apply migrations. Pending: {}.",
@@ -141,6 +184,18 @@ async fn apply_migrations(
         .run_data_migrations()
         .await
         .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+
+    // Apply the table key metadata repair after the schema migrations, so it never
+    // reads or writes a pre-migration catalog shape.
+    if repair_pending {
+        let applied = bootstrap
+            .repair_lost_sort_key_definitions(true)
+            .await
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+        for line in &applied.repaired {
+            println!("  Restored sort key definition: {line}");
+        }
+    }
 
     // Read new catalog version.
     let new = bootstrap
--- a/crates/storage-mongodb/src/table_engine.rs
+++ b/crates/storage-mongodb/src/table_engine.rs
@@ -17,7 +17,9 @@ use extenddb_core::types::{
 };
 use extenddb_storage::TableEngine;
 use extenddb_storage::error::StorageError;
-use extenddb_storage::util::{index_arn, sk_info, stream_arn, table_arn};
+use extenddb_storage::util::{
+    index_arn, merge_attribute_definitions, sk_info, stream_arn, table_arn,
+};
 
 use crate::MongoEngine;
 use crate::data::data_collection_name;
@@ -797,6 +799,33 @@ impl MongoEngine {
 
         // Handle GSI updates
         if let Some(gsi_updates) = &input.global_secondary_index_updates {
+            // Persist the request's attribute definitions, merged into the stored
+            // set. This backend previously never wrote them at all, so a created
+            // index's key attributes were missing from the catalog and `sk_info`
+            // resolved the index's own sort key to `None`, making it behave as
+            // hash-only. Merging (rather than replacing) is what keeps the base
+            // table's pk/sk definitions intact, which is issue #259 on the SQL
+            // backends. Read-modify-write without a transaction, matching the rest
+            // of this method; the catalog row is only written by UpdateTable.
+            let merged_attr_defs = if let Some(new_attr_defs) = &input.attribute_definitions {
+                let current = self
+                    .describe_table_impl(account_id, &input.table_name)
+                    .await?;
+                let merged =
+                    merge_attribute_definitions(&current.attribute_definitions, new_attr_defs);
+                let merged_bson =
+                    bson::to_bson(&merged).map_err(|e| StorageError::Internal(e.to_string()))?;
+                tables_coll
+                    .update_one(
+                        doc! { "_id": { "account_id": account_id, "table_name": &input.table_name } },
+                        doc! { "$set": { "attribute_definitions": merged_bson } },
+                    )
+                    .await
+                    .map_err(|e| StorageError::Internal(e.to_string()))?;
+                Some(merged)
+            } else {
+                None
+            };
             for update in gsi_updates {
                 if let Some(create) = &update.create {
                     // Fetch table_id
@@ -854,7 +883,9 @@ impl MongoEngine {
                         &index_id,
                         &create.key_schema,
                         &desc.key_schema,
-                        &desc.attribute_definitions,
+                        merged_attr_defs
+                            .as_deref()
+                            .unwrap_or(&desc.attribute_definitions),
                     )
                     .await?;
 
@@ -1000,7 +1031,7 @@ impl MongoEngine {
         }
         let has_lsi = !local_secondary_indexes.is_empty();
 
-        Ok(TableKeyInfo {
+        let key_info = TableKeyInfo {
             table_name,
             account_id,
             table_id,
@@ -1011,7 +1042,14 @@ impl MongoEngine {
             global_secondary_indexes,
             local_secondary_indexes,
             stream_specification,
-        })
+        };
+        // Catalog metadata that cannot describe its own sort key would make the
+        // keyed read paths fall back to a partition-only lookup and return the
+        // wrong item, so refuse it here rather than serve a wrong answer (#259).
+        key_info
+            .validate_sort_key_definitions()
+            .map_err(StorageError::Internal)?;
+        Ok(key_info)
     }
 
     async fn index_info_impl(
--- a/crates/storage-mongodb/src/table_engine.rs
+++ b/crates/storage-mongodb/src/table_engine.rs
@@ -808,11 +808,26 @@ impl MongoEngine {
             // backends. Read-modify-write without a transaction, matching the rest
             // of this method; the catalog row is only written by UpdateTable.
             let merged_attr_defs = if let Some(new_attr_defs) = &input.attribute_definitions {
-                let current = self
-                    .describe_table_impl(account_id, &input.table_name)
-                    .await?;
-                let merged =
-                    merge_attribute_definitions(&current.attribute_definitions, new_attr_defs);
+                // Read just the stored definitions rather than a full describe_table_impl,
+                // which would also load every index. Two concurrent UpdateTables on one
+                // table can still interleave here and lose one side's additions; that is
+                // inherent to this backend's UpdateTable, which is non-transactional
+                // throughout (the index inserts and collection creation below are not
+                // atomic with this write either), so the window is narrowed rather than
+                // closed.
+                let current_defs: Vec<extenddb_core::types::AttributeDefinition> = tables_coll
+                    .find_one(
+                        doc! { "_id": { "account_id": account_id, "table_name": &input.table_name } },
+                    )
+                    .await
+                    .map_err(|e| StorageError::Internal(e.to_string()))?
+                    .as_ref()
+                    .and_then(|d| d.get("attribute_definitions"))
+                    .map(|b| bson::from_bson(b.clone()))
+                    .transpose()
+                    .map_err(|e| StorageError::Internal(format!("attr_defs parse error: {e}")))?
+                    .unwrap_or_default();
+                let merged = merge_attribute_definitions(&current_defs, new_attr_defs);
                 let merged_bson =
                     bson::to_bson(&merged).map_err(|e| StorageError::Internal(e.to_string()))?;
                 tables_coll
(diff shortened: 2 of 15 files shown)


crates/app/src/cmd_migrate.rs
jcshepherd, line 1: Should we be baking detection and repair into the command? If I'm understanding the problem, this is something affecting builds up to version X: builds after that should not be affected. And version X is still considered a beta release anyway, right? Could we provide a simple separate tool for detecting and doing the repair, and advertise that, to reduce the risk of baked-in repair logic coming back to bite us in unanticipated ways, long after the original problem was forgotten about?

crates/storage-mongodb/src/table_engine.rs
```diff
@@ -797,6 +799,48 @@ impl MongoEngine {
...
+            // resolved the index's own sort key to `None`, making it behave as
+            // hash-only. Merging (rather than replacing) is what keeps the base
+            // table's pk/sk definitions intact, which is issue #259 on the SQL
+            // backends. Read-modify-write without a transaction, matching the rest
+            // of this method; the catalog row is only written by UpdateTable.
+            let merged_attr_defs = if let Some(new_attr_defs) = &input.attribute_definitions {
+                // Read just the stored definitions rather than a full describe_table_impl,
+                // which would also load every index. Two concurrent UpdateTables on one
+                // table can still interleave here and lose one side's additions; that is
+                // inherent to this backend's UpdateTable, which is non-transactional
```
jcshepherd, line 814: It'll be interesting to see if this becomes an issue. It's been one of the challenges with Cassandra as well, and I find myself wondering if we are eventually going to need/want more of a workflow-driven control plane than we have today.

crates/storage-postgres/src/bootstrapper.rs
```diff
@@ -522,6 +524,147 @@ impl Bootstrapper for PostgresBootstrapper {
...
+    /// Repair table metadata damaged by the pre-fix `UpdateTable` (#259).
+    ///
+    /// Reads the catalog's key schemas and attribute definitions, and for any base
+    /// sort key with no definition recovers the type from the data table's
+    /// physical sort key column (`sk_s`/`sk_n`/`sk_b`, or the numbered variants for
+    /// multiple sort keys). Runs under the migration lock the caller holds.
+    async fn repair_lost_sort_key_definitions(&self, apply: bool) -> OpResult<KeyDefinitionRepair> {
+        let mut report = KeyDefinitionRepair::default();
+        let Ok(catalog) = self.app_pool(&self.config.catalog_db).await else {
+            return Ok(report);
```
jcshepherd, line 536: Silent bailout if no catalog connection?

crates/storage-postgres/src/bootstrapper.rs
```diff
@@ -522,6 +524,147 @@ impl Bootstrapper for PostgresBootstrapper {
         migrations::table_exists(&pool, "settings").await
     }
 
+    /// Repair table metadata damaged by the pre-fix `UpdateTable` (#259).
+    ///
+    /// Reads the catalog's key schemas and attribute definitions, and for any base
+    /// sort key with no definition recovers the type from the data table's
+    /// physical sort key column (`sk_s`/`sk_n`/`sk_b`, or the numbered variants for
+    /// multiple sort keys). Runs under the migration lock the caller holds.
+    async fn repair_lost_sort_key_definitions(&self, apply: bool) -> OpResult<KeyDefinitionRepair> {
```
jcshepherd, line 533: Is this transactional? If not, why not?

jcshepherd (approved):
Comments/questions but nothing I'd consider blocking.

PR #264, opened 2026-08-13
ci: mirror Cosign signature artifacts through the GHCR mirror job

robinnsc:
## What

Extends the GHCR mirror job (#256) to also mirror **Cosign signature
artifacts**. The `candidate_tag` input now accepts two shapes:

- `sha-<40 hex>` — an image candidate (existing behavior, unchanged), or
- `sha256-<64 hex>.sig` — the Cosign signature artifact for an image digest.

For signature artifacts the job additionally derives the subject image digest
from the tag and **refuses to mirror a signature whose image is not already in
GHCR** — a dangling signature cannot exist through this path. The
architecture assertions (amd64/arm64 present) now apply only to image
indexes, since a signature artifact has no platforms. Everything else is
shared and unchanged: strict input validation via environment, the
source-must-still-match-verified-digest gate, overwrite refusal with
idempotent same-digest no-op, `skopeo copy --all --preserve-digests`, and
post-publish digest verification. Same protected `ghcr` environment, so every
signature mirror is reviewer-approved like every image mirror.

## Why

We are moving from detached signature bundles to **registry-attached** Cosign
signatures, so users get the standard verification UX
(`cosign verify --key …`) and policy engines (e.g. Kubernetes admission
controllers) can enforce signatures at deploy time.

Cosign stores a signature as an ordinary OCI artifact tagged
`sha256-<image digest>.sig` in the same repository. For Docker Hub and ECR we
attach signatures directly at signing time, but GHCR's only acceptable
credential is the workflow-scoped `GITHUB_TOKEN` — the same constraint that
motivated the mirror job for images applies to their signatures. Signing flow
becomes: sign the Docker Hub copy → dispatch this job once for the image,
once for its `.sig` artifact.

## Testing done

- YAML parses (`yaml.safe_load`).
- The tag→subject-digest derivation (`sha256-<hex>.sig` → `sha256:<hex>`) and
  both regexes were desk-checked; hex cannot contain the delimiter so the
  single-substitution rewrite is unambiguous.
- **The new path has not executed** (dispatch-only, reviewer-gated, and no
  signed candidate exists yet). It will be exercised in the signing rehearsal
  against the abandoned `sha-c34aac11…` candidate before the next release
  uses it in earnest.

Not verified: same runtime caveats as #256 (skopeo presence asserted with an
apt fallback; org package settings unverifiable without org admin).

## Checklist

- [ ] I have added or updated tests for new functionality — no workflow test harness; the gates are the assertions
- [ ] I have updated documentation if behavior changed — release runbook (internal) is the operating document and is being updated with the attached-signing procedure

ADR / RFC: n/a — release plumbing.

## Breaking changes

None. The existing image-candidate input shape and behavior are byte-for-byte
unchanged; the new shape is additive. Nothing runs without a manual dispatch
plus an environment reviewer's approval.

---

### DIFF
--- a/.github/workflows/ghcr-mirror.yml
+++ b/.github/workflows/ghcr-mirror.yml
@@ -17,6 +17,12 @@
 # The digest input is the contract with the manual verification that precedes
 # this job: the operator supplies the digest they verified, and this job
 # refuses to mirror anything else, even if the Docker Hub tag has moved since.
+#
+# Two artifact kinds are supported, distinguished by tag shape:
+#   sha-<40 hex>              an image candidate (multi-arch index)
+#   sha256-<64 hex>.sig       the Cosign signature artifact for an image digest
+# A signature may only be mirrored after the image it signs is already in
+# GHCR, so a dangling signature cannot exist through this path.
 
 name: ghcr-mirror
 
@@ -24,10 +30,10 @@ on:
   workflow_dispatch:
     inputs:
       candidate_tag:
-        description: 'Verified Docker Hub candidate tag, e.g. sha-<full 40-char commit>'
+        description: 'Verified Docker Hub tag: image candidate (sha-<full 40-char commit>) or Cosign signature artifact (sha256-<64 hex>.sig)'
         required: true
       expected_digest:
-        description: 'The verified multi-arch index digest, e.g. sha256:<64 hex>'
+        description: 'The verified manifest digest of that tag, e.g. sha256:<64 hex>'
         required: true
 
 # Copying between registries needs no repository content at all.
@@ -57,8 +63,17 @@ jobs:
           RAW_DIGEST: ${{ inputs.expected_digest }}
         run: |
           set -euo pipefail
-          if [[ ! "$RAW_TAG" =~ ^sha-[0-9a-f]{40}$ ]]; then
-            echo "::error::candidate_tag '$RAW_TAG' is not a commit-addressed candidate tag (sha-<40 hex>)"
+          KIND=''
+          SUBJECT_DIGEST=''
+          if [[ "$RAW_TAG" =~ ^sha-[0-9a-f]{40}$ ]]; then
+            KIND=image
+          elif [[ "$RAW_TAG" =~ ^sha256-[0-9a-f]{64}\.sig$ ]]; then
+            KIND=signature
+            # sha256-<hex>.sig  ->  sha256:<hex>  (the image digest it signs)
+            SUBJECT_DIGEST="${RAW_TAG%.sig}"
+            SUBJECT_DIGEST="${SUBJECT_DIGEST/-/:}"
+          else
+            echo "::error::candidate_tag '$RAW_TAG' is neither an image candidate (sha-<40 hex>) nor a signature artifact (sha256-<64 hex>.sig)"
             exit 1
           fi
           if [[ ! "$RAW_DIGEST" =~ ^sha256:[0-9a-f]{64}$ ]]; then
@@ -68,6 +83,8 @@ jobs:
           {
             echo "TAG=$RAW_TAG"
             echo "DIGEST=$RAW_DIGEST"
+            echo "KIND=$KIND"
+            echo "SUBJECT_DIGEST=$SUBJECT_DIGEST"
           } >> "$GITHUB_ENV"
 
       - name: Gate - skopeo present (preinstalled on ubuntu-latest)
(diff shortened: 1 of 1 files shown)


LeeroyHannigan:
Reviewed this two ways: a hostile static pass over the whole resulting workflow (not just the diff, and diffed three-dot against a freshly fetched `main`), plus a live pass that actually executed the path the PR says has never run, against a localhost registry with real skopeo and real cosign. **No blocking findings.** One should-fix, two nits, and one pre-existing issue that matters more than anything in this diff.

**The claims hold up, and I tested rather than eyeballed them.** Both input regexes are fully anchored, and the expression that classifies which branch runs *is* the validating expression, so they cannot drift apart. 24 adversarial `candidate_tag` values were rejected at the correct gate with rc=1: `$(id)`, backticks, slash, `..`, embedded and trailing newline, leading/trailing space, empty, wrong hex length and case, and `.sig.sig`. Notably bash `[[ =~ ^…$ ]]` does not leak a trailing newline, so `sha-<40hex>\nrm -rf /` is rejected, which is the classic hole in this pattern. The dangling-signature gate does inspect the destination (GHCR), not the source, so the property is real, and the login correctly precedes it.

**Your `.sig` shape assumption is empirically confirmed.** Real cosign v2.4.1 default-mode signing produced exactly `sha256-<64hex>.sig`, the gate accepted it, and the derived `SUBJECT_DIGEST` equalled the subject index digest exactly. `skopeo copy --all --preserve-digests` on the signature works with the digest preserved byte-for-byte, and the dangling refusal fires when the subject is absent from the destination and passes once present. Also worth recording: the `.sig` really is a single OCI manifest rather than an index, so skipping the amd64/arm64 assertions for signatures is not merely acceptable, it is necessary.

**Should-fix, the one real ask:** `KIND` is derived from the tag *name* only (lines 66-71) and never from the artifact's `mediaType`/`artifactType`, so a multi-arch image published under a `.sig`-shaped tag would take the signature branch and escape the architecture assertions entirely. `release-image.yml` verifies architecture from content, and that contrast is what makes this worth closing rather than a nit. Asserting the Cosign artifactType in the signature branch would do it. Bounded by operator trust today, since dispatch is reviewer-gated.

**Nit 1:** OCI-1.1 referrers-mode cosign produces a `sha256-<hex>` tag with no `.sig` suffix (reproduced), which this gate rejects, so such signatures are silently unmirrorable. It fails closed, which is the right direction, but the runbook should pin cosign's default tag-based mode explicitly.

**Nit 2:** the skopeo apt fallback omits `apt-get update` and fails rc=100 on a cold cache (reproduced). It is unreachable in practice, since ubuntu-24.04 ships Skopeo 1.13.3, so this is only about the fallback not working if it were ever needed. Related caveat on my own evidence: I proved the signature copy on skopeo 1.22.2, whereas the runner has 1.13.3.

**One ask that isn't about this diff, but which this diff needs.** The workflow declares only `permissions: contents: read` (line 40) with no `packages: write` at any level, yet the `mirror` job logs in with `GITHUB_TOKEN` and pushes to GHCR. Since your signature path pushes to GHCR the same way an image mirror does, it's blocked by the same thing, so it seemed better to raise it here than to send you a separate one-line PR that would only conflict with this file. It came in with #256, so it predates this change:

```yaml
  mirror:
    permissions:
      contents: read
      packages: write
```

Scoped to the job rather than the workflow, since `mirror` is the only thing that pushes. Worth knowing the org default won't rescue it: an explicit workflow-level `permissions:` block overrides the repo/org default `GITHUB_TOKEN` scope, so declaring `contents: read` pins the token read-only however permissive the default is. That's why this would fail deterministically rather than flakily. The reason nobody has hit it yet is that this workflow has run exactly once, on 13 Aug, and that run was cancelled before it pushed, so the push has never actually executed. Given mirroring is step 3 of the release runbook, better found now than mid-release.

Happy to re-verify both the artifactType assertion and the GHCR push against the same fixtures once you've pushed.

robinnsc:
Thank, addressed those comments:

1. **`packages: write` (the pre-existing #256 bug):** added, scoped to the `mirror` job only, with a comment explaining why the job-level grant is needed (the workflow-level block deliberately pins the default token read-only). You're right that this failed deterministically and simply hadn't been reached — better here than mid-release, and agreed it belonged in this PR rather than a conflicting one-liner.

2. **Should-fix, content-based classification:** added a gate on the signature branch that inspects the source manifest before anything else happens: `mediaType` must be a single OCI image manifest (your empirical point that a `.sig` is never an index, inverted into an assertion), and every layer must be a `application/vnd.dev.cosign.simplesigning.v1+json` payload. A multi-arch image under a `.sig`-shaped tag now fails this gate instead of escaping the architecture assertions.

3. **Nit 1 (referrers mode):** documented in the header — tag-based storage mode is the expectation, and the signing procedure in our release runbook pins cosign's default tag-based mode explicitly. The fail-closed behavior on bare `sha256-<hex>` tags is intentional and now stated.

4. **Nit 2 (cold apt cache):** `apt-get update` now precedes the fallback install. Noted your skopeo 1.22.2-vs-1.13.3 caveat; the operations used (`inspect --raw/--format`, `copy --all --preserve-digests`, `login/logout`) are all present well before 1.13, but the rehearsal against the abandoned candidate will run on the actual runner version and settle it empirically.

A re-verification of the artifactType assertion and the now-permitted GHCR push against your fixtures would be very welcome — the planned rehearsal (mirror the abandoned `sha-c34aac11…` candidate, dev-key sign, mirror the `.sig`) will exercise the full path on the real runner as well.

LeeroyHannigan (approved):
LGTM 🚀

PR #269, opened 2026-08-14
fix: declare --sqlite-path on init and make the documented path selection real

### LINKED ISSUE #267: [Bug] serve --sqlite-path is rejected as an unknown argument
## Describe the bug

`docs/getting-started.md` and `extenddb.sample.toml` both document
`--sqlite-path` as the way to set the SQLite database path at serve time, and
give it as the first entry in the resolution order. `serve` rejects the flag.

With the flag unavailable, `[storage.sqlite].path` in the config file is the
only way to set the path, including for `:memory:`.

## To Reproduce

1. Build with the SQLite backend:

   ```bash
   cargo build --release -p extenddb --no-default-features --features sqlite
   ```

2. Initialise:

   ```bash
   ./target/release/extenddb init --backend sqlite
   ```

3. Run the command from `docs/getting-started.md`:

   ```bash
   ./target/release/extenddb serve --config extenddb.toml --sqlite-path /data/extenddb.sqlite
   ```

## Expected behavior

`serve` accepts `--sqlite-path` and uses that path, ahead of
`[storage.sqlite].path` and the default, as documented:

> At serve time the path can be overridden with `--sqlite-path`. The resolution
> order is: the `--sqlite-path` flag, then `[storage.sqlite].path` in the
> config, then the default `extenddb.sqlite`
>
> - `docs/getting-started.md`

## Actual behavior

```
error: unexpected argument '--sqlite-path' found

Usage: extenddb serve --config <CONFIG>

For more information, try '--help'.
```

`extenddb serve --help` does not list `--sqlite-path`, and `ServeArgs` in
`crates/app/src/cmd_serve.rs` declares `config`, `port`, `foreground` and
`write_pid_file` only. The same happens with the flag before the subcommand
(`extenddb --sqlite-path <p> serve`), and `extenddb --help` shows no top-level
options besides `-V` and `-h`.

`crates/storage-sqlite/src/bootstrapper.rs` reads `--sqlite-path` from the
argument list via `helpers::extract_arg`.

Reproduced on `v0.1.3` (`3961c63`). `ServeArgs` on `main` is unchanged.

## Environment

- ExtendDB version: 0.1.3 (tag `v0.1.3`, `3961c63`)
- Operating system: macOS 26.6.1
- Rust version (if building from source): 1.95.0
- Client SDK/driver and version: n/a, reproduces on the CLI

## Logs / stack trace

Included under Actual behavior.

LeeroyHannigan:
## What

Makes the documented `--sqlite-path` real. It is declared on `init` (where the genuinely unworkaround-able gap was: no way to choose the SQLite path before a config file exists), the bootstrapper now creates the parent directory for new paths, and the docs are corrected to describe what actually works at serve time (`[storage.sqlite].path`, or the previously undocumented `EXTENDDB__STORAGE__SQLITE__PATH` override). `serve` stays backend-agnostic, matching the Postgres convention of no serve-time storage flags.

## Why

The flag was documented and half-implemented (the bootstrapper reads it via `extract_arg`) from the original SQLite backend commit, but never declared to clap, so it was rejected on every command and the extraction was unreachable dead code. Postgres's init flags work only because `InitArgs` declares them too.

Closes #267

## Testing done

- Clap-level regression test with a discriminating negative control: removing the declaration reproduces the pre-fix rejection; a genuinely unknown flag is still rejected.
- Live end-to-end on the built binary: `init --backend sqlite --sqlite-path <fresh-dir>/place.sqlite` creates the database there, writes the path into the generated `[storage.sqlite]` section, and `serve` opens it. The parent-directory fix was itself found by this live run (SQLite error 14 on a not-yet-existing directory).
- `cargo fmt --check` clean, `cargo clippy --workspace -D warnings` clean, workspace suite 787 passed / 0 failed / 0 filtered out.

## Checklist

- [x] Breaking changes are noted below (if any) - none

ADR / RFC: n/a - adds one already-documented flag to `init`'s existing per-backend flag set, mirroring the Postgres flags; no protocol, trait, auth, or format change.

---

### DIFF
--- a/crates/app/src/cmd_init.rs
+++ b/crates/app/src/cmd_init.rs
@@ -29,6 +29,12 @@ pub struct InitArgs {
     #[arg(long)]
     catalog_db: Option<String>,
 
+    /// SQLite database file path (default: extenddb.sqlite). The chosen path
+    /// is written to the generated config file, so `serve` finds it without
+    /// further flags. SQLite backend only.
+    #[arg(long)]
+    sqlite_path: Option<String>,
+
     /// PostgreSQL host (hostname, IP address, or absolute Unix socket directory path)
     #[arg(long)]
     pg_host: Option<String>,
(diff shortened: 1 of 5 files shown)


jcshepherd (approved):
LGTM

PR #273, opened 2026-08-16
fix: adding a GSI to a table with no sort key fails with a 500

LeeroyHannigan:
`UpdateTable` cannot add a GSI to **any partition-key-only table** on the PostgreSQL backend. The caller gets `InternalServerError` with nothing to go on.

## Root cause

The backfill that copies existing rows into a new index read the base table with a hardcoded ordering:

```sql
SELECT item_data FROM {base_table} ORDER BY pk, sk_s, sk_n, sk_b LIMIT $1 OFFSET $2
```

A hash-only table has no sort-key columns at all, so PostgreSQL rejects this with `column "sk_s" does not exist`. The backfill runs inside index creation, so the failure rolls the whole `UpdateTable` back and surfaces as a 500.

A partition-key-only table is the simplest table DynamoDB allows, so this makes GSIs unreachable for a large and entirely ordinary class of tables.

This is the same defect family as #259 and the sort-key handling in the restore path: key layout inferred from a fixed column list instead of the table's real schema.

## Fix

Order by only the key columns the base table actually has: `pk`, plus the three typed columns for each sort-key position that exists.

## Verification

`tests/rust/src/gsi_on_hash_only_table.rs` covers the add end to end on a hash-only table: `UpdateTable` succeeds, the index appears on `DescribeTable`, and a row written **before** the index existed is readable through it afterwards. That last assertion is the one that matters, since it distinguishes a real backfill from an index that copied nothing.

Proven discriminating rather than merely green:

| binary | result |
|---|---|
| hardcoded ordering restored | test **fails** on the `UpdateTable` call with `InternalServerError` |
| this fix | passes, and the pre-index row is returned through the index |

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --workspace` (798 passed, 0 failed, 0 filtered out) are all clean.

## Note for sequencing

A follow-up PR adds `AttributeDefinitions` pruning on `UpdateTable`. Its integ tests use hash-only tables, so they cannot pass until this lands. This one should go first.

### DIFF
--- a/crates/storage-postgres/src/table_helpers.rs
+++ b/crates/storage-postgres/src/table_helpers.rs
@@ -80,8 +80,22 @@ impl PostgresEngine {
         let idx_sks = data::all_sort_key_info(index_key_schema, attr_defs);
         let base_sks = data::all_sort_key_info(base_key_schema, base_attr_defs);
 
+        // Order by whatever key columns the base table actually has. A
+        // hash-only base table has no sort-key columns at all, so naming them
+        // unconditionally makes the backfill fail with "column sk_s does not
+        // exist" and the whole GSI add is rolled back.
+        let mut order_cols = vec!["pk".to_owned()];
+        for (i, _) in base_sks.iter().enumerate() {
+            if i == 0 {
+                order_cols.extend(["sk_s".to_owned(), "sk_n".to_owned(), "sk_b".to_owned()]);
+            } else {
+                let n = i + 1;
+                order_cols.extend([format!("sk{n}_s"), format!("sk{n}_n"), format!("sk{n}_b")]);
+            }
+        }
         let sql = format!(
-            "SELECT item_data FROM {base_table} ORDER BY pk, sk_s, sk_n, sk_b LIMIT $1 OFFSET $2"
+            "SELECT item_data FROM {base_table} ORDER BY {} LIMIT $1 OFFSET $2",
+            order_cols.join(", ")
         );
         let mut offset: i64 = 0;
         loop {
--- a/tests/rust/src/gsi_on_hash_only_table.rs
+++ b/tests/rust/src/gsi_on_hash_only_table.rs
@@ -1,0 +1,145 @@
+// Copyright 2026 ExtendDB contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Adding a GSI to a table that has no sort key.
+//!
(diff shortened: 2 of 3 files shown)


crates/storage-postgres/src/table_helpers.rs
```diff
@@ -80,8 +80,22 @@ impl PostgresEngine {
...
+        // unconditionally makes the backfill fail with "column sk_s does not
+        // exist" and the whole GSI add is rolled back.
+        let mut order_cols = vec!["pk".to_owned()];
+        for (i, _) in base_sks.iter().enumerate() {
+            if i == 0 {
+                order_cols.extend(["sk_s".to_owned(), "sk_n".to_owned(), "sk_b".to_owned()]);
+            } else {
+                let n = i + 1;
+                order_cols.extend([format!("sk{n}_s"), format!("sk{n}_n"), format!("sk{n}_b")]);
+            }
```
robinnsc, line 94: **Non-blocking / fine as a follow-up:** this inlines the position→column-name convention (`i == 0` → `sk_*`, else `sk{i+1}_*`) in a fourth place — `create_data_table` and `create_index_data_table` in `ddl.rs` already carry the same loop, and `sk_column_n` in `storage/src/util/key.rs` is the canonical encoding of the scheme.

Your own root-cause note says this defect family keeps recurring because key layout gets re-derived locally instead of coming from one source of truth — same argument applies to the naming convention itself. A small helper next to `sk_column_n`, e.g. `all_sk_columns_at(position: usize) -> [String; 3]`, used here (and opportunistically in `ddl.rs`, or in a follow-up PR) would keep the convention in one place.

Related micro-point, also non-blocking: the loop discards the type (`for (i, _)`), but ordering by exactly the PRIMARY KEY columns via `sk_column_n(i, sk_type)` would be equivalent for correctness and lets Postgres satisfy the sort from the PK index instead of sorting each batch — could matter on a large backfill. Not a regression either way.

robinnsc (commented):
Fix looks right — verified the generated column list against all three table shapes in `ddl.rs` (hash-only, single SK, multi SK), and it also quietly fixes the non-total ordering for multi-sort-key tables, where OFFSET pagination could previously skip or duplicate rows across batches. One non-blocking design suggestion inline.

robinnsc (approved):

PR #274, opened 2026-08-16
fix: prune unreferenced AttributeDefinitions on UpdateTable

LeeroyHannigan:
Stacked on #273. Based on `fix/gsi-add-on-hash-only-table` so the diff shows only this change; retarget to `main` once #273 merges.

`UpdateTable` merged the request's `AttributeDefinitions` into the stored set but never removed any, so the stored set only ever grew. Real DynamoDB keeps it equal to the attributes actually referenced by the table key schema plus the key schemas of the live indexes, so two cases diverged, both visible to any client calling `DescribeTable` after an `UpdateTable`:

* an unused `AttributeDefinition` supplied alongside a GSI add was stored, where DynamoDB drops it;
* deleting a GSI left behind the definitions only that index referenced.

## Fix

The effective set is recomputed as the stored definitions merged with the request's, then pruned to those still referenced.

Merging is still required and is not being undone here: the request carries only the attributes it needs, so replacing the set drops the base table's own `pk`/`sk` and degrades keyed reads to a partition-only lookup (#259, fixed in #261). **Pruning is the half that was missing.**

Two details worth a reviewer's attention:

* the recompute reuses `merge_attribute_definitions` rather than reimplementing the merge, which keeps the stored definition when both sides name the same attribute, so a conflicting redeclaration cannot silently retype a live index key;
* it runs whether or not the request carried `AttributeDefinitions`, because a GSI deletion prunes without the request naming anything. It reads the surviving index key schemas after the create/delete rows are applied in the same catalog transaction, under the `FOR UPDATE` lock already held on the table row.

## Verification

`tests/rust/src/update_table_attribute_definitions.rs`, 4 tests, all passing:

```
unused_attribute_definition_supplied_with_a_gsi_add_is_dropped ... ok
removing_a_gsi_drops_only_its_own_attribute_definitions ....... ok
conflicting_redeclaration_keeps_the_stored_attribute_type ..... ok
sequential_gsi_adds_accumulate_attribute_definitions ......... ok
```

These use hash-only tables, which is why this stacks on #273: against `main` they fail at table setup with the unrelated 500 that #273 fixes.

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --workspace` (798 passed, 0 failed, 0 filtered out) are clean.

### DIFF
--- a/crates/storage-mongodb/src/table_engine.rs
+++ b/crates/storage-mongodb/src/table_engine.rs
@@ -18,7 +18,8 @@ use extenddb_core::types::{
 use extenddb_storage::TableEngine;
 use extenddb_storage::error::StorageError;
 use extenddb_storage::util::{
-    index_arn, merge_attribute_definitions, sk_info, stream_arn, table_arn,
+    effective_attribute_definitions, index_arn, merge_attribute_definitions, sk_info, stream_arn,
+    table_arn,
 };
 
 use crate::MongoEngine;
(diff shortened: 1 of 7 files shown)


jcshepherd (approved):
LGTM

PR #275, opened 2026-08-16
fix: validate secondary-index keys on UpdateItem

LeeroyHannigan:
Stacked on #201. Based on `fix/data-plane-request-validation` so the diff shows only this change; retarget to `main` once #201 merges.

`UpdateItem` never validated the item it produced against the table's secondary index keys, so an update expression could set an index key to an empty string or binary value, or to a value of the wrong type, and the write silently succeeded. The transactional `Update` path already performed exactly these checks, so plain `UpdateItem` was the only way through.

## Fix

The validator was already correct, and #201 already made it name the offending value's type. Only the caller was missing. `UpdateItem` now runs the same two checks the transactional path runs, against the **evaluated post-update image** rather than the expression, which is what covers `if_not_exists` and `list_append` whose result is not knowable from the expression alone. `index_key_refs` becomes `pub(crate)` so both paths share one construction of the index key references.

Sixteen lines plus a visibility change.

## Verification

Live, against a running server with a `B`-typed GSI hash key:

```
UpdateExpression: SET bidx = :e   with :e = B""
before: succeeded silently
after:  ValidationException — "The update expression attempted to update a
        secondary index key to a value that is not supported. The AttributeValue
        for a key attribute cannot contain an empty binary value."
```

`tests/rust/src/index_key_validation.rs` gains the two `UpdateItem` cases: the rejection, and a **positive control** that a valid value is still accepted, so a change that rejected everything could not pass. All 6 tests in the file pass.

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --workspace` (808 passed, 0 failed, 0 filtered out).

### DIFF
--- a/crates/core/src/validation/mod.rs
+++ b/crates/core/src/validation/mod.rs
@@ -998,8 +998,27 @@ pub fn validate_index_keys(
     indexes: &[IndexKeyRef<'_>],
     attr_defs: &[AttributeDefinition],
 ) -> Result<(), DynamoDbError> {
+    validate_index_keys_in_context(item, indexes, attr_defs, SecondaryIndexEmptyContext::Item)
+}
+
+/// [`validate_index_keys`] with the caller choosing the empty-value message
+/// variant.
+///
+/// DynamoDB words the rejection differently depending on how the offending value
+/// arrived. An item supplied directly reports the attribute and index by name; a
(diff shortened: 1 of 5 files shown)


jcshepherd (approved):
LGTM

PR #277, opened 2026-08-17
fix(import-export): namespace files per account beneath each root

LeeroyHannigan:
## What this changes

Import and export file paths are now namespaced per account. A caller in account
`A` may only resolve paths beneath `<root>/A` for each configured `[import]` /
`[export]` root; anything else is refused. The account id comes from the
authenticated identity rather than the request.

Previously the allow-list was server-wide with no account dimension
(`ImportExportPathConfig` is a bare `Vec<String>`, resolved once at startup and
shared across request contexts), and `validate_path` / `validate_path_parent`
enforced containment in one of those roots and nothing else. On an instance
hosting more than one account, that meant a caller could name any path inside a
configured root regardless of who wrote it.

Four supporting changes on the same paths:

- **Containment is decided lexically, before any filesystem access.** Deciding it
  after canonicalization meant a path outside the caller's subtree answered
  "does not exist" when absent and the containment message when present, which
  distinguished the two. The post-canonicalization check is kept so a symlinked
  ancestor still cannot redirect out of the subtree.
- **Export refuses to overwrite.** The write is `create_new`, so an export cannot
  truncate a file it did not create, and an existing symlink at the destination
  is refused rather than followed. The final path component is symlink-checked
  explicitly, matching the import read path which already did this.
- **The import source is validated before the destination table is created**, so a
  refused import no longer leaves an empty orphaned table behind.
- **Both operations emit a record** carrying the acting account, the operation, the
  table and the resolved path. Neither was previously observable: they are
  synchronous and persist nothing.

## Behaviour changes for existing deployments

- A path directly under a configured root is no longer accepted.
- A file exported before this change cannot be imported after it without being
  moved into the account subtree.
- Re-exporting to an existing path now fails instead of truncating.

`import_export_root` continues to be honoured. It populates both lists from a
single value, which previously guaranteed the import and export roots
overlapped; with per-account namespacing a shared root is no longer a
cross-account concern, so startup warns rather than refusing. The shared example
path is dropped from the sample config.

## Tests

12 unit tests over the validators: own-subtree accept; foreign-subtree refuse
with the victim file present, so the refusal cannot come from absence;
bare-root refuse; identical answers for foreign-and-present versus
foreign-and-absent on both the import and export paths; symlink refusal;
relative paths in both directions; and the malformed-account-id guard.

4 integration tests drive the two primitives and the overwrite refusal over the
wire against a foreign account's path, following the approach
`backup_arn_scoping` takes for ARNs. They assert the exact containment message
rather than any `ValidationException`, the destruction test re-reads the victim's
bytes to prove they are intact, and the read-denial test asserts no table was
created.

Verified by building a server from the pre-change code and running the same four
integration tests against it: all four fail there (both primitives succeed, with
`DID NOT RAISE`), and all four pass after. Existing import/export tests move onto
the account subtree, and export destinations no longer pre-create the file.

Gates: `cargo fmt --all -- --check` exit 0, `cargo clippy --all-targets -- -D
warnings` exit 0, `cargo test --workspace` 810 passed with 0 filtered out, and
the 14 import/export integration tests green against a live server.

## Coverage note

This lives in `crates/engine` and touches no storage backend, so the behaviour is
identical on all three. Worth knowing that both the Postgres and SQLite
integration jobs run pytest with `--filter "not import_export"`, so
`test_import_export.py` executes only on the MongoDB job today. The unit tests run
everywhere via `cargo test`.

## Docs

`13-event-ticketing-demo.md` exported to a bare-root path, which this change
refuses, then `cat`ed a file that was never written; the demo was broken end to
end. `10-security-model.md` listed only `..`, canonicalization and symlinks under
path-traversal protection. `troubleshooting.md` explained the containment error
in terms that no longer cover a path inside a root but outside the account
subtree, and had no entry for the overwrite refusal.

### DIFF
--- a/crates/engine/src/import_export.rs
+++ b/crates/engine/src/import_export.rs
@@ -12,7 +12,9 @@ use serde_json::Value;
 
 use crate::OperationContext;
 use crate::create_table::storage_err_to_dynamo;
-use crate::import_export_io::{read_items, validate_path, validate_path_parent};
+use crate::import_export_io::{
+    ensure_account_dirs, read_items, validate_path, validate_path_parent,
+};
 use crate::serialize_output;
 use extenddb_core::error::DynamoDbError;
 use extenddb_core::types::{
@@ -76,8 +78,17 @@ pub async fn handle_import_table(
         .await
         .map_err(storage_err_to_dynamo)?;
 
-    // Validate and canonicalize the source path.
-    let source_path = validate_path(&input.file_source.path, &ctx.import_paths)?;
+    // Validate and canonicalize the source path. Files are namespaced per
+    // account, so a caller can only read beneath its own subtree of a root.
+    let source_path = validate_path(&input.file_source.path, &ctx.import_paths, &ctx.account_id)?;
+
+    tracing::info!(
+        account_id = %ctx.account_id,
+        operation = "ImportTable",
+        table_name = %tcp.table_name,
+        resolved_path = %source_path.display(),
+        "import/export file access"
+    );
 
     // Check file size against limit.
     let file_meta = std::fs::metadata(&source_path).map_err(|_| {
@@ -213,20 +224,50 @@ pub async fn handle_export_table(
         .await
         .map_err(storage_err_to_dynamo)?;
 
+    // Export files are namespaced per account. The account's subtree has to
+    // exist before the output path can be canonicalized and jailed, so create
+    // it first; this is idempotent and keeps the first export for a new account
+    // from failing on a missing parent directory.
+    ensure_account_dirs(&ctx.export_paths, &ctx.account_id)?;
+
     let output_path = validate_path_parent(
         input
             .resolve_file_path()
             .map_err(|e| DynamoDbError::ValidationException(e.to_owned()))?,
         &ctx.export_paths,
+        &ctx.account_id,
     )?;
     if let Some(parent) = output_path.parent() {
         tokio::fs::create_dir_all(parent).await.map_err(|_| {
             DynamoDbError::ValidationException("Cannot create output directory".to_owned())
         })?;
     }
-    let mut file = tokio::fs::File::create(&output_path)
+
+    tracing::info!(
+        account_id = %ctx.account_id,
+        operation = "ExportTableToPointInTime",
+        table_name = %table_name,
+        resolved_path = %output_path.display(),
+        "import/export file access"
+    );
+
+    // `create_new` refuses to write when anything already exists at the path,
+    // so an export cannot truncate a file it did not create. It also refuses an
+    // existing symlink, including a dangling one, rather than following it.
+    let mut file = tokio::fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(&output_path)
         .await
-        .map_err(|_| DynamoDbError::ValidationException("Cannot create export file".to_owned()))?;
+        .map_err(|e| {
+            if e.kind() == std::io::ErrorKind::AlreadyExists {
+                DynamoDbError::ValidationException(
+                    "Export file already exists; choose a different FilePath".to_owned(),
+                )
+            } else {
+                DynamoDbError::ValidationException("Cannot create export file".to_owned())
+            }
+        })?;
 
     let mut item_count: i64 = 0;
     let max_export_items = ctx.limits.max_export_item_count;
--- a/crates/engine/src/import_export.rs
+++ b/crates/engine/src/import_export.rs
@@ -51,6 +51,32 @@ pub async fn handle_import_table(
     let start_time = epoch_seconds();
     let tcp = &input.table_creation_parameters;
 
+    // Validate and canonicalize the source path before creating anything. Files
+    // are namespaced per account, so a caller can only read beneath its own
+    // subtree of a root. Validating first means a refused import leaves no
+    // orphaned destination table behind.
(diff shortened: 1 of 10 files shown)


crates/server/src/serve.rs
jcshepherd: I'm confused why we're doing this check now: by the time we've hit it we've already not-ignored import_export_root. I guess it's not logically incorrect, but it's unexpected.

crates/engine/src/import_export_io.rs
```diff
@@ -169,14 +169,120 @@ fn split_csv_line(line: &str, delim: u8) -> Vec<String> {
...
+/// account answers identically whether or not the file is there. Deciding the
+/// jail after canonicalization would turn the error into an oracle for other
+/// tenants' filenames. `..` is rejected by the caller, so a lexical prefix test
+/// is sound here; the post-canonicalization check still catches symlink escapes.
+fn reject_outside_account_subtree(
+    path: &Path,
+    jail_roots: &[Arc<PathBuf>],
+    account_id: &str,
+) -> Result<(), DynamoDbError> {
+    if jail_roots.is_empty() {
```
jcshepherd, line 266: If jail_roots is empty then any path is okay? It seems like an error if we get to this point.

crates/server/src/serve.rs
LeeroyHannigan: Good catch, and it was worse than unexpected: the advice was unactionable. When `import_export_root` supplies both lists the roots are the same directory by construction, so "consider separate directories" is not something the operator can act on. The remedy is migrating to `[import]`/`[export]`, which the warning never said.

Chasing it turned up a second defect in the same block. The deprecation warning only fired when the key was fully unused. A key that filled exactly one list, which is what happens when only one section is configured, was applied with no diagnostic at all, and "ignored" was wrong about a key in force for the other operation.

Fixed in 2d0aec9. Resolution now records where each list came from, so every state of the key produces one notice naming what happened and the remedy: supplies both, supplies one, or unused. The overlap notice is suppressed only for identical roots when both came from the key; nesting is still always reported since it cannot arise from a single root. Resolved paths are unchanged, only log lines differ.

It moved to `crates/server/src/import_export_paths.rs`, free of I/O, because it being inline in `serve` with no tests is why this reached you. 15 unit tests cover all eight section/key combinations, verified discriminating by sabotage.

Separately: the commented example in `extenddb.sample.toml` was unusable. It is a top-level key but the example sits after `[logging]`, so uncommenting it lands the key in that table and the server refuses to start with "unknown field import_export_root". Corrected in the same commit.

jcshepherd (commented):
General question: it seems like in serve,rs, we warn if import and export paths aren't configured, but in import_export.io we silently skip path validation when that happens, allowing any path to be used. Is that the desired behavior?

crates/engine/src/import_export.rs
```diff
@@ -49,6 +51,32 @@ pub async fn handle_import_table(
     let start_time = epoch_seconds();
     let tcp = &input.table_creation_parameters;
 
+    // Validate and canonicalize the source path before creating anything. Files
+    // are namespaced per account, so a caller can only read beneath its own
+    // subtree of a root. Validating first means a refused import leaves no
+    // orphaned destination table behind.
+    let source_path = validate_path(&input.file_source.path, &ctx.import_paths, &ctx.account_id)?;
+
+    tracing::info!(
```
robinnsc, line 60: Non-blocking: the description calls this "emitting a record", but it's an info-level log line — anyone running at `warn` (or with syslog filtering) loses the only visibility into these file accesses, which would be a shame given the whole point of adding it. Maybe give it a dedicated tracing target (e.g. `target: "audit"`) so operators can keep it regardless of level, or at least note the required log level in the security-model doc.

crates/engine/src/import_export.rs
```diff
@@ -213,20 +226,50 @@ pub async fn handle_export_table(
         .await
         .map_err(storage_err_to_dynamo)?;
 
+    // Export files are namespaced per account. The account's subtree has to
+    // exist before the output path can be canonicalized and jailed, so create
+    // it first; this is idempotent and keeps the first export for a new account
+    // from failing on a missing parent directory.
+    ensure_account_dirs(&ctx.export_paths, &ctx.account_id)?;
```
robinnsc, line 233: Non-blocking: this runs before the path is validated, so any export request — including one that's about to be refused — creates the caller's directory under every configured export root. It's harmless, but it is a filesystem write triggered by a request we haven't validated yet. Doing it after `validate_path_parent` succeeds (or only for the root the resolved path actually lives under) would be a touch tidier.

crates/engine/src/import_export_io.rs
```diff
@@ -169,14 +169,120 @@ fn split_csv_line(line: &str, delim: u8) -> Vec<String> {
...
+        return Ok(());
+    }
+    if is_under_scoped_root(&absolutize(path)?, jail_roots, account_id)? {
+        return Ok(());
+    }
+    Err(DynamoDbError::ValidationException(
+        "Path must resolve under one of the configured allowed paths".to_owned(),
```
robinnsc, line 273: Non-blocking: keeping this message identical across the three refusal cases is exactly right, but the wording predates the account dimension, and it's what every existing single-account user whose paths just broke will see. Since it's the same text for every caller, appending something like "; paths are namespaced per account" leaks nothing and would save people a trip to the troubleshooting doc. (Would need the same tweak in the tests and docs that assert the exact string.)

robinnsc (approved):
Left three small comments inline; none of them blocking.

PR #278, opened 2026-08-18
ci: check license notices freshness at PR time

robinnsc:
## What

A `Licenses` workflow that runs `devtools/generate-software-license-notices
--check` on pull requests, merge groups, and pushes to `main` — the exact
script and pinned `cargo-about 0.9.0` the `release-image` gate uses (the
script itself refuses any other version, so the PR-time and release-time
checks cannot drift apart). Path-filtered to the generator's inputs
(`Cargo.toml`/`Cargo.lock`, the notices file, about config, the script
itself); the binary is cached so the common case adds seconds, not a
compile. `contents: read` only.

## Why

Follow-up to the process gap called out in #272: the notices check ran only
inside `release-image`, so #271's version bump merged 32/32 green while
leaving `main` unable to produce a release candidate — discovered only when
the `v0.1.6` tag was dispatched. This moves the failure to the PR that
causes it, where the fix is one regeneration command instead of a burned
tag dispatch.

The release-gate check stays: it is the unconditional backstop for anything
the path filter cannot anticipate.

## Testing done

- `yaml.safe_load` passes on the workflow.
- The underlying script run locally on current `main` (post-#272):
  `SOFTWARE-LICENSE-NOTICES.html is current` — so this check goes green on
  merge, not red.
- Discriminating case: the same script at `75e1685` (pre-#272) fails with
  the exact CI message, per the verification table in #272 — this workflow
  would have blocked #271 at PR time.
- Path filter includes the workflow file itself, so future edits to the
  check re-trigger it.

## Checklist

- [x] I have added or updated tests for new functionality — the check is itself the test; discriminating pass/fail cases cited above
- [x] I have updated documentation if behavior changed — header documents rationale and the #271/#272 incident

## Breaking changes

None.

---

### DIFF
--- a/.github/workflows/licenses.yml
+++ b/.github/workflows/licenses.yml
@@ -1,0 +1,63 @@
+# Copyright 2026 ExtendDB contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# PR-time license notices freshness check.
+#
+# SOFTWARE-LICENSE-NOTICES.html embeds the version of every workspace crate,
+# so it goes stale on any version bump or dependency change. Until now the
+# only check lived in release-image's gate, so a PR could merge fully green
+# while leaving main unreleasable — exactly what happened with the 0.1.6
+# bump (#271, fixed by #272). This runs the SAME script with the SAME pinned
+# cargo-about (the script refuses any other version), so PR-time and
+# release-time can never disagree.
+#
+# Path-filtered to the generator's actual inputs; the release gate remains
+# the unconditional backstop.
+
+name: Licenses
+
+on:
+  pull_request:
+    paths:
+      - '**/Cargo.toml'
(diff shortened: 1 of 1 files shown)


LeeroyHannigan (approved):
Good change. Running the same script with the same pinned cargo-about at PR time removes any way for the PR gate and release gate to disagree, and the path filter matches the generator's real inputs. Approving.

- `dtolnay/rust-toolchain@master` is a floating branch ref. Everything else in CI is `@stable` or a pinned version, use `@stable` here too.
- If we make `notices-current` a required check, require it via the merge queue. It's path-filtered on pull_request, so PRs that don't touch those paths would wait forever on a check that never runs.
- Once the dev container lands there'll be a second notices file for the sqlite feature set that needs the same treatment.

PR #279, opened 2026-08-18
ci: CI signing via OIDC KMS, now with the Rekor transparency log

robinnsc:
## What

Revives the CI signing workflow from #268 (closed unmerged during
post-launch cleanup, no review findings) rebased onto current `main`, with
one substantive change: **the Rekor transparency log is ON**
(`--tlog-upload=true`), and the verify step checks the log entry instead of
ignoring it. Everything else is as reviewed before: `dockerhub` environment
(two-person rule enforced mechanically, registry credential from
environment secrets), OIDC-federated AWS access to
`ExtendDBContainerSigningCI` (live and verified since 2026-08-14),
checksum-pinned cosign, SHA-pinned actions, strict input validation
including the `release_tag` annotation input, and the post-sign gate that
fails the run unless the legacy `.sig` TAG exists.

## Why

Phase 1 of the release-pipeline consolidation (runbook §7): signing moves
off operator laptops into a reviewer-gated workflow. Rekor could not be
enabled for v0.1.5 because manual signing ran on corp machines, where
`rekor.sigstore.dev` is DNS-sinkholed; CI runners reach it fine. The Rekor
bundle rides in the `.sig` annotations, so the signature mirrored to GHCR
(and copied to ECR) verifies with the transparency log from any registry —
and from 0.1.6 on, user verify instructions drop the
`--insecure-ignore-tlog` flag.

v0.1.5's signatures remain valid and verifiable exactly as released; this
is additive.

## Testing done

- `yaml.safe_load` passes; rebase onto current `main` is conflict-free.
- The legacy tag-based flag combination was proven live against ECR on
  2026-08-14 (`.sig` tag, single OCI manifest, simplesigning layer). The
  only changed flag is `--tlog-upload=false` → `true`, which is the
  documented v2-era standard pairing with tag-based storage.
- Not executable pre-merge (main-only environment, as with #256/#264/#266).
  Proposed live verification: after merge, dispatch against the **v0.1.5**
  digest (`sha256:4890886b…`) — signing an already-signed digest just adds
  a second signature to the existing `.sig` tag, so it exercises OIDC, KMS,
  Rekor, and the tag gate end-to-end with zero risk to the released
  artifact, before 0.1.6 depends on this path.

## Checklist

- [x] I have added or updated tests for new functionality — post-sign `.sig` tag gate is the storage-mode regression test; live rehearsal plan above
- [x] I have updated documentation if behavior changed — header rewritten for the Rekor decision; runbook §7 tracks this as Phase 1

## Breaking changes

None. The manual signing path remains as break-glass.

---

### DIFF
--- a/.github/actions/sign-image/action.yml
+++ b/.github/actions/sign-image/action.yml
@@ -105,4 +105,4 @@ runs:
         KEY_ARN: ${{ inputs.key-arn }}
       run: |
         set -euo pipefail
-        cosign verify --key "awskms:///${KEY_ARN}" "${IMAGE_REPO}@${DIGEST}" | head -1
+        cosign verify --key "awskms:///${KEY_ARN}" "${IMAGE_REPO}@${DIGEST}" >/dev/null
--- a/.github/actions/sign-image/action.yml
+++ b/.github/actions/sign-image/action.yml
@@ -1,0 +1,108 @@
+# Copyright 2026 ExtendDB contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Composite action: sign one verified image digest with the production KMS
+# key. Attached Cosign signature, LEGACY TAG-BASED storage (sha256-<hex>.sig
+# — required by the mirror tooling and published verify instructions), Rekor
+# transparency log ON.
+#
+# Single source of truth for the signing flags. cosign v3's defaults write a
+# new-format OCI referrer and NO .sig tag (hit live on ECR, 2026-08-14); the
+# three-flag combination below is the proven tag-based mode, and the
+# post-sign gate fails if the .sig tag did not appear.
+#
+# The CALLER is responsible for: reviewer-gated environment, AWS credentials
+# (OIDC role with kms:Sign on the key), and registry login. This action only
+# signs and verifies.
+
+name: sign-image
+description: Sign a verified image digest with the production KMS key (tag-based .sig, Rekor on)
+
+inputs:
+  image-repo:
+    description: 'Registry repository, e.g. docker.io/extenddb/extenddb-postgres'
+    required: true
+  digest:
+    description: 'Verified image index digest (sha256:<64 hex>); caller must have validated it'
+    required: true
+  release-tag:
+    description: 'Release the digest belongs to (vX.Y.Z); caller must have validated it'
+    required: true
+  key-arn:
+    description: 'KMS key ARN'
+    required: true
+  cosign-version:
+    description: 'Pinned cosign version'
+    default: v3.1.3
+  cosign-sha256:
+    description: 'SHA-256 of the pinned cosign-linux-amd64 binary'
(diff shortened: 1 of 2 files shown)


.github/actions/sign-image/action.yml
```diff
@@ -0,0 +1,108 @@
...
+
+    - name: Verify the signature (includes the Rekor entry)
+      shell: bash
+      env:
+        IMAGE_REPO: ${{ inputs.image-repo }}
+        DIGEST: ${{ inputs.digest }}
+        KEY_ARN: ${{ inputs.key-arn }}
+      run: |
+        set -euo pipefail
+        cosign verify --key "awskms:///${KEY_ARN}" "${IMAGE_REPO}@${DIGEST}" | head -1
```
LeeroyHannigan:
```suggestion
        cosign verify --key "awskms:///${KEY_ARN}" "${IMAGE_REPO}@${DIGEST}" >/dev/null
```

Line 107 is set -euo pipefail, so head -1 closing the pipe can SIGPIPE cosign to exit 141 and fail the job after the signature is already pushed and the .sig tag exists. Fail-closed, so not dangerous, but it reads as a signing failure when signing worked.

.github/workflows/sign-image.yml
```diff
@@ -0,0 +1,113 @@
...
+          } >> "$GITHUB_ENV"
+
+      - name: Gate - the digest must exist on Docker Hub as a multi-arch index
+        run: |
+          set -euo pipefail
+          RAW=$(skopeo inspect --raw "docker://${IMAGE_REPO}@${DIGEST}")
+          grep -q '"architecture": *"amd64"' <<< "$RAW" || { echo "::error::amd64 missing from index"; exit 1; }
+          grep -q '"architecture": *"arm64"' <<< "$RAW" || { echo "::error::arm64 missing from index"; exit 1; }
+
+      - uses: actions/checkout@v4
```
LeeroyHannigan:
```suggestion
      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
```

@v4 is a moving tag: GitHub can repoint it at different code later, so what runs in this job can change without a PR. This job holds the signing credentials, so it should be pinned to a fixed commit like configure-aws-credentials below already is.

robinnsc:
Thanks, addressed those comments

LeeroyHannigan (approved):
Signing is digest-bound, inputs are validated and routed through env, actions are SHA-pinned, and dispatch-only plus the environment gate keeps forks away from the role. Approving.

Two follow-ups worth doing:
- The headers say routine releases sign inside release-image's publish stage, but release-image never references the composite action, so this wrapper is currently the only caller. Either wire release-image to it or fix the comments, otherwise the next person assumes pipeline signing already exists.
- Worth confirming the IAM trust policy pins `sub` to `repo:ExtendDB/extenddb:environment:dockerhub` rather than the whole repo. That condition is the real control and it lives outside this diff.

Minor: the arch check greps the raw index JSON for the architecture strings. `jq` on `manifests[].platform.architecture` would be tighter.

PR #280, opened 2026-08-18
fix: prune unreferenced AttributeDefinitions on UpdateTable (re-land #274)

LeeroyHannigan:
Re-lands #274, which merged into `fix/gsi-add-on-hash-only-table` rather than `main`.

#273 merged to `main` at 09:34:34 and #274 merged 30 seconds later, into #273's head
branch, before GitHub retargeted it. That branch had already served its purpose, so
#274's commits never reached `main`: `effective_attribute_definitions` is absent from
`main`, and `crates/storage-postgres/src/update_table.rs` still calls
`merge_attribute_definitions` with no prune step.

The two content commits from #274 (`a71bc7c`, `9d6aff9`) are cherry-picked onto current
`main`, which already carries #273 and #277. Both applied without conflict. No content
change from what was approved on #274.

## What this restores

UpdateTable prunes attribute definitions that are no longer referenced by the table key
schema or by a surviving index, instead of accumulating every definition ever supplied.
The merge-then-prune rule lives in `crates/storage/src/util/key.rs` as
`effective_attribute_definitions`, with all three backends wired to it: SQLite and
Postgres in `update_table.rs`, and MongoDB in `table_engine.rs`, where the prune runs
after the index create/delete loop because that loop consumes the merged set to resolve a
new index's key types.

## Verification

`cargo fmt --all -- --check` exit 0, `cargo clippy --all-targets -- -D warnings` exit 0,
`cargo test --workspace` 831 passed with 0 filtered out.

### DIFF
--- a/crates/storage-mongodb/src/table_engine.rs
+++ b/crates/storage-mongodb/src/table_engine.rs
@@ -18,7 +18,8 @@ use extenddb_core::types::{
 use extenddb_storage::TableEngine;
 use extenddb_storage::error::StorageError;
 use extenddb_storage::util::{
-    index_arn, merge_attribute_definitions, sk_info, stream_arn, table_arn,
+    effective_attribute_definitions, index_arn, merge_attribute_definitions, sk_info, stream_arn,
+    table_arn,
 };
 
 use crate::MongoEngine;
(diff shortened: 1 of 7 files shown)


LeeroyHannigan:
Merged under admin bypass, recorded here so it reads as deliberate rather than as a gap.

This PR is a re-land of #274, which merged into `fix/gsi-add-on-hash-only-table` rather than `main`: #273 merged to `main` at 09:34:34 and #274 merged 30 seconds later into #273's head branch, before GitHub retargeted it. Its commits never reached `main`.

The content here is a conflict-free cherry-pick of #274's two commits (`a71bc7c`, `9d6aff9`) onto current `main`, with no edits, so it is the same change @jcshepherd approved on #274. Gates re-run on the cherry-picked result: `cargo fmt --all -- --check` exit 0, `cargo clippy --all-targets -- -D warnings` exit 0, `cargo test --workspace` 831 passed with 0 filtered out.

Merged via a temporary admin bypass because the reviewers who could re-approve are in timezones that are hours from being online, and this blocks the v0.1.6 release that carries the import/export fix from #277. Branch protection and the `main` ruleset were both restored immediately after the merge and verified byte-identical to their prior state (`enforce_admins: true`, `bypass_actors: []`).

PR #283, opened 2026-08-19
ci(promote): fan out promotion to GHCR and ECR in one approved run

robinnsc:
## What

Extends `promote-image` so the one approved run promotes all three
registries: after the existing Docker Hub promotion and verification, the
same job applies the identical `X.Y.Z` (+ guarded `latest`) tags to GHCR
(`GITHUB_TOKEN`, job-scoped `packages: write`) and ECR Public (OIDC
publisher role, `id-token: write`) by digest with `--preserve-digests`.
Each leg has its own no-overwrite gate on the version tag and post-copy
digest verification; any mismatch fails the run. Existing gates (semver,
digest pin, immutability, latest-never-backwards) are untouched and now
govern all three registries from one evaluation.

## Why

Phase 1 of the pipeline consolidation (release runbook §7). In v0.1.6,
promotion took one workflow approval + a manual Isengard/crane ECR retag +
two more `ghcr-mirror` dispatches with two more approvals. This collapses
that to the single already-existing approval. The ECR credential is the
OIDC-federated `ExtendDBEcrPublicGitHubPublisher` role, trust-pinned to
this repository's `dockerhub` environment (audited least-privilege
2026-08-18): no human credential handling remains in promotion.

Ordering note: `latest` on GHCR/ECR is only written when `move_latest` is
true and after the version tag exists on that registry — the same
forward-only property `ghcr-mirror` enforces (#282), here guaranteed by
sequence rather than re-derivation.

## Testing done

- `yaml.safe_load` passes; the Docker Hub steps are byte-identical to the
  reviewed current version (pure additions after its verify step).
- Gate logic traced against live registry state (all three registries
  currently serve 0.1.6 + latest at the same digest): re-running for 0.1.6
  would no-op every leg; a divergent version tag fails the leg's gate.
- Not executable pre-merge (main-only environment). Live verification
  options, lowest-risk first: re-dispatch for 0.1.6 (full no-op path), or
  first use on 0.1.7.

## Checklist

- [x] I have added or updated tests for new functionality — per-leg gates + digest verification are the checks; no-op trace above

## Breaking changes

None. Dispatch inputs unchanged; Docker Hub behavior identical.

---


.github/workflows/promote-image.yml
```diff
@@ -247,6 +257,71 @@ jobs:
...
+              echo "::error::${GHCR_REPO}:${VERSION} already exists with different digest ${EXISTING}"
+              exit 1
+            fi
+          else
+            skopeo copy --all --preserve-digests \
+              "docker://${IMAGE_REPO}@${DIGEST}" \
+              "docker://${GHCR_REPO}:${VERSION}"
+          fi
+          # latest: forward-only, and only when this dispatch moves it.
+          if [[ "$MOVE_LATEST" == "true" ]]; then
```
LeeroyHannigan: The forward-only check only reads Docker Hub's tag list (`skopeo list-tags` on `IMAGE_REPO`, L187). Here and in the ECR leg, latest moves on `MOVE_LATEST` alone, so if this registry's tags ever diverge from Docker Hub (partial promotion, out-of-band push), re-promoting an older version passes the Docker Hub gate and still rewrites latest backwards here. The header comment at L38-40 claims each leg has the forward-only property, which isn't true yet.

Each leg should run the same highest-released-version comparison against its own registry's tag list before moving latest, or refuse to move latest when the destination's current latest resolves to a higher version tag than `$VERSION`.

LeeroyHannigan (changes requested):
The fan-out itself is in good shape: every copy is by digest with a post-copy digest assertion, inputs are validated and env-routed, and re-runs after a partial failure are safe no-ops. But the forward-only latest guard doesn't cover the new registries, and the header comment says it does. Inline comment on the GHCR leg, same applies to the ECR leg.

Two more before this merges:
- `skopeo copy` doesn't pick up the cosign `.sig` tag, so the GHCR/ECR copies land unsigned with nothing re-signing them. Needs a `cosign copy` per destination (or a re-sign step) plus a post-copy check that the signature exists.
- `skopeo login ghcr.io -u "${{ github.actor }}"` is the one place an expression reaches a run: block directly. The rest of the file routes everything through env, give this the same treatment.

Nit: the GHCR/ECR `skopeo logout` lines aren't `if: always()` like the Docker Hub one, a failed copy skips them.

.github/workflows/promote-image.yml
```diff
@@ -247,6 +257,71 @@ jobs:
             fi
           done
 
+      - name: Fan out - promote GHCR by digest (same tags, same gates)
+        env:
+          GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GHCR_REPO: ghcr.io/extenddb/extenddb-postgres
```
LeeroyHannigan: Hardcode 1 of 3, and this is what blocks `extenddb-dev` reusing the workflow.

Derive it from the image instead: one `IMAGE_NAME: extenddb-postgres` input, with the registries composing `ghcr.io/extenddb/${IMAGE_NAME}` and `public.ecr.aws/extenddb/${IMAGE_NAME}`. A second image then becomes an input rather than a fork of the file.

.github/workflows/promote-image.yml
```diff
@@ -247,6 +257,71 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          skopeo login ghcr.io -u "${{ github.actor }}" --password-stdin <<< "$GHCR_TOKEN"
```
LeeroyHannigan: `github.actor` is whoever clicked dispatch, so the login identity varies per operator even though the token is always `GITHUB_TOKEN`. GHCR ignores the username when the password is a GitHub token, so pin it to a constant (`${{ github.repository_owner }}`, or literally `x-access-token`) and take the operator out of the path.

.github/workflows/promote-image.yml
LeeroyHannigan: This is the blocking one from my first review, restated here because the fix and the refactor are the same work.

The Docker Hub leg gates `latest` on this being the highest released version. This leg doesn't, so a back-line promote with `move_latest` checked moves GHCR's `latest` backwards while Docker Hub correctly refuses, and the registries then disagree about what `latest` means. Same shape at line 314 for ECR.

Cheapest correct fix is one shared guard that checks the destination's current `latest` digest per registry, which falls out of the extraction.

.github/workflows/promote-image.yml
```diff
@@ -247,6 +257,71 @@ jobs:
...
+      - name: Fan out - assume the ECR publisher role via OIDC
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
+        with:
+          role-to-assume: arn:aws:iam::975306274793:role/ExtendDBEcrPublicGitHubPublisher
+          role-session-name: extenddb-promote-${{ github.run_id }}
+          aws-region: us-east-1
+
+      - name: Fan out - promote ECR Public by digest (same tags, same gates)
+        env:
+          ECR_REPO: public.ecr.aws/extenddb/extenddb-postgres
```
LeeroyHannigan: Hardcode 2 of 3, and this block is the GHCR block again with two names changed: same no-overwrite check, same `latest` copy, same verify loop.

That's the duplication worth extracting before a second image arrives, otherwise `extenddb-dev` turns three copies into six.

.github/workflows/promote-image.yml
```diff
@@ -265,7 +340,7 @@ jobs:
             echo ""
             echo "| field | value |"
             echo "|---|---|"
-            echo "| repository | \`${IMAGE_REPO}\` |"
+            echo "| registries | \`${IMAGE_REPO}\`, \`ghcr.io/extenddb/extenddb-postgres\`, \`public.ecr.aws/extenddb/extenddb-postgres\` |"
```
LeeroyHannigan: Hardcode 3 of 3, and this one lies rather than fails: the summary prints the Postgres GHCR/ECR paths regardless of what was promoted. If `IMAGE_REPO` ever becomes an input, this reports the wrong registries for the dev image on a green run. Compose it from the same variables the copy steps use.

LeeroyHannigan (commented):
We're shipping a second image, `extenddb/extenddb-dev` (#281, the SQLite dev container), and it should land in the same three registries as the Postgres one. As written this PR can't serve it: the repo is hardcoded in three separate places and the promote logic is copy-pasted per registry.

The direction I'd suggest is the pattern you set yourself in #279: pull the per-registry promote into a composite action taking `src_repo`, `dst_repo`, `version`, `digest`, `move_latest`. `promote-image.yml` then calls it once per registry (or over a matrix), and `release-dev-image.yml` calls the same thing for `extenddb-dev`. One no-overwrite rule, one forward-only rule, one verify loop, two images.

Worth knowing before that lands: the dev workflow has no promote stage. It's a single dispatch that publishes `:${version}` straight after the smoke test, so there's no candidate-to-release step for a promote-time fan-out to hook into. A composite action serves both shapes; extending `promote-image.yml` in place only serves the Postgres shape.

The blocking item from my first review is still open, and it's now the strongest argument for extracting: the forward-only `latest` guard runs only against Docker Hub, so both new legs can move `latest` backwards. Fixing that in one shared place is one fix. Fixing it in three copied blocks is where drift starts.

Two setup items that aren't code: `ghcr.io/extenddb/extenddb-dev` and `public.ecr.aws/extenddb/extenddb-dev` will need creating, and GHCR packages auto-create **private** and need a manual visibility flip plus repo linking, unlike Docker Hub which auto-creates public on first push.

robinnsc:
All of it addressed in 0198524, taking your extraction suggestion wholesale — you were right that the blocking fix and the refactor were the same work, and the composite came out cleaner than the three blocks it replaced:

**`.github/actions/promote-registry`** now holds the single implementation, called once per registry (and callable by `release-dev-image` for the dev image). Per destination it enforces:
- **no-overwrite** on version tags (idempotent on the identical digest — the re-entrancy property survives the extraction)
- **refuse-unsigned** (your second find, and the one that would have bitten hardest): the `.sig` tag must already exist on the destination or be copied from the source with digest assertion; no signature anywhere → the leg fails. Promotion can no longer strand an unsigned copy.
- **forward-only `latest`, judged against the destination's own tag list** (the blocking one): each registry defends itself, so a diverged registry refuses a backwards move even when Docker Hub's gate passes. Fixed once, applies three times.

The workflow slims to gates + login/logout + three action calls. Also:
- `image` is now an allowlisted **choice** input (`extenddb-postgres` | `extenddb-dev`) composing all three repo paths and the summary — hardcodes 1–3 gone, and the summary can no longer lie about what it promoted
- GHCR login uses the constant `x-access-token` — `github.actor` no longer appears anywhere in a run block
- registry logouts are consolidated under one `if: always()` step

On your setup notes: agreed — `ghcr.io/extenddb/extenddb-dev` + the ECR repo need creating before the dev image fans out, GHCR's auto-created-private + visibility-flip behavior is already documented in the runbook from the postgres launch, and the ECR role policies need `extenddb-dev` added to their resource lists (runbook carries that too; happy to run the role-policy edit whenever the dev multi-registry decision is made).

Happy for another pass — particularly on the refuse-unsigned ordering (signature copy happens after the version tag exists on the destination but before latest moves, so a failure between leaves a signed version tag and no latest move, which re-runs cleanly).

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,113 @@
...
+
+        # --- version tag: immutable, idempotent on the identical digest ---
+        if EXISTING=$(skopeo inspect --format '{{.Digest}}' "docker://${DST}:${VERSION}" 2>/dev/null); then
+          if [[ "$EXISTING" != "$DIGEST" ]]; then
+            echo "::error::${DST}:${VERSION} already exists with different digest ${EXISTING}; released versions are immutable"
+            exit 1
+          fi
+          echo "::notice::${DST}:${VERSION} already promoted with the exact digest"
+        else
+          skopeo copy --all --preserve-digests \
```
LeeroyHannigan, line 135: **Blocking:** ordering. For an unsigned artifact, :58 misses so this copy publishes `DST:X.Y.Z`, and only then does the sig gate at :70-82 fail the leg. The release tag is left live pointing at an unsigned image, and since it's immutable per :58-62 a corrected signed rebuild is then refused, so it needs a manual tag delete to recover.

Moving :70-82 above this block fixes it, and it's safe since the `.sig` is a separate tag you can check on SRC first.

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,113 @@
...
+            [[ "$DST_SIG" == "$SRC_SIG" ]] || { echo "::error::signature copy digest mismatch (${DST_SIG} != ${SRC_SIG})"; exit 1; }
+          else
+            echo "::error::no Cosign signature for ${DIGEST} on ${SRC} or ${DST}; refusing to promote an unsigned artifact"
+            exit 1
+          fi
+        fi
+
+        # --- latest: forward-only, judged against THIS destination's tags ---
+        if [[ "$MOVE_LATEST" == "true" ]]; then
+          HIGHEST=$(skopeo list-tags "docker://${DST}" \
```
LeeroyHannigan: **Blocking:** fails open. The `|| true` wraps the whole substitution, so a failed `list-tags` leaves `HIGHEST` empty, :92 is false and `latest` is copied unconditionally. Promoting a `1.2.4` hotfix while `2.0.0` is highest, one transient error downgrades every unpinned user.

Split the network call from the parse so a list failure is fatal, keeping `|| true` only on the parse where an empty match is legitimate.

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,113 @@
...
+        DST: ${{ inputs.dst-repo }}
+        VERSION: ${{ inputs.version }}
+        DIGEST: ${{ inputs.digest }}
+        MOVE_LATEST: ${{ inputs.move-latest }}
+      run: |
+        set -euo pipefail
+        SIG_TAG="sha256-${DIGEST#sha256:}.sig"
+
+        # --- version tag: immutable, idempotent on the identical digest ---
+        if EXISTING=$(skopeo inspect --format '{{.Digest}}' "docker://${DST}:${VERSION}" 2>/dev/null); then
```
LeeroyHannigan: **Should-fix:** same conflation as :86. This `if` sees only zero/nonzero with stderr discarded, so a transient inspect failure looks like "tag absent" and drops to the copy at :64-67, overwriting an already-released version. Only treat a definite `manifest unknown`/`not found` as absent.

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,113 @@
...
+          fi
+          echo "::notice::${DST}:${VERSION} already promoted with the exact digest"
+        else
+          skopeo copy --all --preserve-digests \
+            "docker://${SRC}@${DIGEST}" \
+            "docker://${DST}:${VERSION}"
+        fi
+
+        # --- refuse unsigned promotion: the signature must exist on DST ---
+        if ! DST_SIG=$(skopeo inspect --format '{{.Digest}}' "docker://${DST}:${SIG_TAG}" 2>/dev/null); then
```
LeeroyHannigan: **Should-fix:** this proves a tag exists, not that a signature verifies, and there's no `cosign verify` in the action. An empty or wrong-key manifest at that tag name passes. Either verify, or soften "refuses unsigned artifacts" at :11 to what's actually checked.

.github/workflows/promote-image.yml
```diff
@@ -63,7 +78,10 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  IMAGE_REPO: docker.io/extenddb/extenddb-postgres
+  # inputs.image is a closed choice list; no free text reaches these paths.
+  IMAGE_REPO: docker.io/extenddb/${{ inputs.image }}
+  GHCR_REPO: ghcr.io/extenddb/${{ inputs.image }}
+  ECR_REPO: public.ecr.aws/extenddb/${{ inputs.image }}
```
LeeroyHannigan, line 89: **Blocking:** `extenddb-dev` doesn't exist on ECR Public (404 `NAME_UNKNOWN`; postgres is 200) and ECR Public doesn't auto-create on push. This leg runs last, so a dev dispatch promotes Docker Hub and GHCR and then dies, leaving a half-rolled promotion with no rollback.

Also nothing upstream produces a signed dev candidate yet, so the Docker Hub leg would fail the unsigned check anyway. Either pre-flight all three destinations before any copy, or drop dev from the choice list for now.

LeeroyHannigan (commented):
Nice extraction, one shared implementation of the promotion rules is the right call. Two blockers inline, both fail-open.

One thing worth fixing once rather than four times: the same fail-open probe appears five times across this PR, #284 and #281. `skopeo inspect ... 2>/dev/null` and `list-tags ... || true` both read "probe failed" as "not there". yesyayen found it independently on #281's `release-dev-image.yml:277`, so it's a pattern, not a one-off.

Smaller stuff:
- **Nit:** `ALREADY_PROMOTED` at :179 is dead now, and the comment at :202-204 describes the removed path.
- **Should-fix:** `ghcr-mirror.yml` still moves GHCR `latest` (:152-172) under the `ghcr` gate, with a different concurrency group, so both can write it at once. Narrow it to candidate + sig, or retire it here.

Merge queue is fine, `promote-image.yml` wasn't touched on main in the 12 commits you're behind.

Also: I expected the cross-registry sig copy to break and it doesn't. cosign ignores `docker-reference` and binds on digest, so copying the `.sig` is sound.

.github/workflows/promote-image.yml
```diff
@@ -72,6 +90,10 @@ jobs:
     # Required reviewers + prevent-self-review + main-only branch rule: the
     # dispatcher cannot approve their own promotion.
     environment: dockerhub
+    permissions:
+      contents: read
+      packages: write   # GHCR promotion leg
```
LeeroyHannigan: **Should-fix:** a job can only hold one `environment`. This is `dockerhub` (:92), but `ghcr-mirror.yml:69` gates GHCR on `ghcr` with required reviewers, so these GHCR/ECR pushes bypass that gate and the design described at :38-44 isn't expressible in one job.

These permissions are also job-scoped, so every step carries them, and :234 still says the DH login is "the only step chain with the credential". Splitting each leg into its own job fixes both.

.github/workflows/promote-image.yml
```diff
@@ -217,39 +239,62 @@ jobs:
...
-              "docker://${IMAGE_REPO}:${VERSION}"
-          fi
-          if [[ "$MOVE_LATEST" == "true" ]]; then
-            skopeo copy --all --preserve-digests \
-              "docker://${IMAGE_REPO}@${DIGEST}" \
-              "docker://${IMAGE_REPO}:latest"
-          fi
+          skopeo login ghcr.io -u x-access-token --password-stdin <<< "$GHCR_TOKEN"
+
+      - name: Promote GHCR (same shared rules, destination-judged latest)
```
LeeroyHannigan: **Nit:** each leg re-judges "highest version" against its own tag list, so on diverged registries the same dispatch moves `latest` on some and refuses on others, exiting mid-fan-out. The DH pre-gate at :187-228 already decides this once, so passing an explicit allow/deny into each leg would be more predictable.

robinnsc:
Round two addressed across 242b740 + 6b3e161, with one structural change from the release owner on top of your findings:

**Blocking:**
- **Ordering** — the signature is now step 1 of every leg: probed, copied if absent, and **cryptographically verified** (this also resolves your :71 — the release public key is now committed as `extenddb-signing.pub.pem`, checked against the live 0.1.6 signatures before committing, so verification is credential-free). No tag of any kind is placed before the signature verifies; the unrecoverable released-but-unsigned state can no longer be constructed.
- **Fail-open latest guard** — the `list-tags` network call is now bare (fatal on failure); only the parse keeps `|| true`. More broadly, every existence probe in the action goes through one three-way helper: definite manifest-unknown → absent; success → digest; anything else → abort the leg. That's the systemic fix for the pattern you and yesyayen found — "probe failed" can no longer read as "not there", anywhere in this file.
- **extenddb-dev half-promotion** — dropped from the choice list (with the reason inline) until its registries exist and something upstream produces a signed dev candidate. Re-adding is two lines behind the same allowlist.

**Should-fixes:**
- **Environment bypass** — restructured into two waves per the release owner: wave 1 (dockerhub env) gates and promotes **Docker Hub alone as the canary**; wave 2 (ghcr env, runs only after wave 1 verifies) promotes GHCR + ECR together. GHCR writes are now approved by the ghcr environment's own reviewers, and your stale-comment point about "the only step chain with the credential" dissolves with the split. The ECR role's OIDC trust gained the ghcr environment sub (verified by read-back; both envs have identical protections).
- **Dual writer on GHCR latest** — `ghcr-mirror` is narrowed back to candidates + signatures in this PR; release tags are promote-image's job exclusively. The #266/#282 release-tag support served the two manual-era releases and retires with credit.
- **Dead `ALREADY_PROMOTED`** and its comment — gone (the composite action's idempotence replaces it).

On your last nit (per-destination judging vs. one decision): kept the destination-side guard but as an **error, never a skip** — wave 1's Docker Hub gate decides intent once, and a mirror that disagrees fails loudly rather than silently diverging. Since wave 2 only runs after the canary verified, the split-brain window you described is now a hard stop with one registry promoted, which re-runs cleanly.

And thanks for testing the cross-registry sig copy rather than assuming either way — digest-bound verification surviving the copy is what makes the whole propagate design sound, so having that empirically settled matters.

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,144 @@
...
+            "docker://${SRC}:${SIG_TAG}" \
+            "docker://${DST}:${SIG_TAG}"
+          DST_SIG=$(probe "${DST}:${SIG_TAG}") || { echo "::error::signature missing after copy"; exit 1; }
+          [[ "$DST_SIG" == "$SRC_SIG" ]] || { echo "::error::signature copy digest mismatch (${DST_SIG} != ${SRC_SIG})"; exit 1; }
+        fi
+        # The signature must actually VERIFY, not merely exist: an empty or
+        # wrong-key manifest at the .sig tag name must not pass. ignore-tlog
+        # because pre-Rekor releases (<= 0.1.6) have no transparency entries;
+        # key verification is unaffected. Tighten when all live releases do.
+        cosign verify --key "$PUBKEY" --insecure-ignore-tlog=true "${SRC}@${DIGEST}" >/dev/null
```
LeeroyHannigan: **Blocking:** cosign is not preinstalled on ubuntu-latest (checked the current runner image manifest - Skopeo 1.13.3 is there, cosign is not) and nothing on this branch installs it. Every leg dies here on first dispatch, after wave-1 approval is spent. CI can't catch it because the action only executes on dispatch.

Fix: `sigstore/cosign-installer` (SHA-pinned) in both jobs, or a version-pinned binary download with checksum in the action itself so callers can't forget it.

.github/actions/promote-registry/action.yml
```diff
@@ -0,0 +1,144 @@
...
+          elif grep -qiE 'manifest unknown|name unknown|not found|unauthorized: repository .* not found' <<< "$out"; then
+            return 9
+          else
+            echo "::error::probe of ${ref} failed (not a definite absence): ${out}" >&2
+            exit 1
+          fi
+        }
+
+        # --- 1. signature first: verified before ANY tag is placed ---
+        set +e; DST_SIG=$(probe "${DST}:${SIG_TAG}"); rc=$?; set -e
```
LeeroyHannigan, line 102: **Should-fix:** the verify at :97 always targets `${SRC}@${DIGEST}`, so when the destination already has a `.sig` (this probe hits, rc=0), that pre-existing destination signature is never verified against anything - a garbage or wrong-key manifest already sitting at the tag name passes the gate, which is the exact hole the verify was added to close. Copied sigs are fine (digest-asserted equal to source). Verifying against `${DST}@${DIGEST}` after this block covers all three paths.

LeeroyHannigan (commented):
Re-reviewed at 5a323a8. This is a big improvement, everything from the last round is addressed: signature-first ordering, the fail-closed `probe()` helper, real `cosign verify` instead of tag-existence, the committed public key (verified byte-identical to the v0.1.6 release asset), dev dropped from the allowlist, the two-wave split putting GHCR under its own `ghcr` gate, ghcr-mirror narrowed so GHCR `latest` has exactly one writer, `ALREADY_PROMOTED` gone, and the local `case` allowlist. Nice work.

One new blocker inline, and it is invisible to CI because nothing in CI executes this action - it only runs on a real dispatch.

- **Nit:** wave-2's checkout comment says `# releases-v4 pin` while wave-1's identical SHA says `# v4.4.0` (:106 vs :296). Same file, two labels.
- **Note:** the ECR role's OIDC trust including the `ghcr` environment is a server-side claim I can't verify from the repo; the comment documents it, which is the right thing, just confirm it before first dispatch.
- **Note:** the concurrency groups still differ from ghcr-mirror's, but since the narrowing means the only overlapping writes are byte-identical `.sig` copies asserted by digest, the race is now benign. No change needed.

robinnsc:
Confirmed — cosign is **not** in the ubuntu-24.04 runner manifest (skopeo 1.13.3 is), so the verify would have died command-not-found on the first leg right after approval. Fixed in the latest commit: the action now installs the same checksum-pinned cosign v3.1.3 the sign-image action uses, skipping when the pinned version is already present so the three calls in one job install once. Pinning rather than relying on preinstallation also insulates verify semantics from runner-image drift — the cosign v2→v3 flag changes were painful enough once. Good catch, thank you.

robinnsc:
Remaining two from your re-review, in 9b8f969 (the cosign install went out separately an hour ago):

- **:102 (verify target)** — right again: the rc=0 path sailed past the gate unverified. The verify now targets `${DST}@${DIGEST}` after the presence/copy block, covering all three paths (copied now, copied earlier, unknown junk at the tag name). Empirically checked: GHCR's mirrored 0.1.6 signature verifies at the destination as expected, so the cross-registry DST-verify works on real artifacts, not just in theory.
- **Nit** — pin comment converged to `# v4.4.0` in wave 2.

On the **trust note** — confirmed again just now rather than from memory; the role's OIDC sub condition reads:

```json
[
    "repo:ExtendDB/extenddb:environment:dockerhub",
    "repo:ExtendDB/extenddb:environment:ghcr"
]
```

(read back via `aws iam get-role` at 2026-08-20T10:18:31Z; the before/after policies from the change are archived in the release evidence.)

And agreed on the concurrency note — with the narrowing, the only overlapping writes are digest-asserted byte-identical `.sig` copies, so leaving the groups separate is deliberate laziness rather than an oversight. That should be the full set — four rounds down, and the action is meaningfully harder to misuse than the three copy-pasted blocks it replaced.

LeeroyHannigan (approved):

PR #284, opened 2026-08-19
ci(release): sign and propagate in the publish stage

robinnsc:
## What

Extends `release-image`'s
`publish-candidate` job so the one publish approval carries the release to
fully-signed-everywhere:

1. registry-side anonymous verification of the just-pushed index (digest +
   both platforms, no credential)
2. sign via the shared composite action from #279 (KMS via OIDC, Rekor ON,
   tag-based `.sig`, post-sign tag gate)
3. propagate image + signature to GHCR (`GITHUB_TOKEN`) and ECR Public
   (OIDC publisher role) by digest, `--preserve-digests`, per-tag
   no-overwrite gates, per-copy digest verification

The job gains `packages: write` + `id-token: write` (job-scoped), a
checkout step (the composite action is in-repo), and updated summary /
footer text. Idempotent re-runs: every leg no-ops on identical digests and
fails on divergent ones.

## Why

Phase 1 completion (runbook §7): with this plus #283, a release is two
approved runs — this workflow (publish→sign→propagate) and `promote-image`
(all-registry promotion). v0.1.6 needed those two approvals **plus** a
manual laptop signing session, a manual Isengard/crane ECR mirror, and
three separate `ghcr-mirror` dispatches with approvals. Atomicity also
closes the ordering hazards we hit live: signing can no longer happen
after promotion (v0.1.5) and no registry ever serves an unsigned
candidate that a `.sig` dispatch could miss.

## Testing done

- `yaml.safe_load` passes.
- The signing steps are the #279 composite action reviewed in that PR; the
  propagation gates replicate the ghcr-mirror gates that have now passed
  live runs for image, signature, and release-tag artifact kinds.
- Not executable pre-merge (main-only environment). Live verification: the
  0.1.7 release, or a re-dispatch of v0.1.6's tag — every leg no-ops on the
  already-current digests except signing, which adds a Rekor-logged second
  signature to the existing `.sig` tag (harmless, and proves the OIDC + KMS
  + Rekor path the same way the planned break-glass rehearsal would).

## Checklist

- [x] I have added or updated tests for new functionality — per-leg gates and digest verification; no-op/live-verification plan above
- [x] I have updated documentation if behavior changed — summary text + deliberate-omissions footer updated; runbook §6/§7 sweep follows the merge

## Breaking changes

None. Tag-push trigger, gate, builds, and the publish approval are unchanged.

---

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -1,10 +1,12 @@
 # Publishes an immutable, commit-addressed CANDIDATE of the ExtendDB PostgreSQL
 # container image to Docker Hub, for linux/amd64 and linux/arm64.
 #
-# This workflow deliberately stops at the candidate. It never creates the
-# version tag or `latest`: promotion, registry mirroring, verification, and
-# signing are maintainer steps recorded in the release runbook. That keeps the
-# blast radius of any workflow defect to a tag no user consumes.
+# This workflow never creates the version tag or `latest` (promotion is
+# promote-image's job, separately approved), but it carries the candidate all
+# the way to signed-and-mirrored: publish + sign under the dockerhub
+# environment approval, then propagation to GHCR + ECR under the ghcr
+# environment approval. Any defect's blast radius stays a tag no user
+# consumes.
 #
 # Shape, and why:
 #
@@ -22,15 +24,18 @@
 #             bit-for-bit what was tested. These jobs hold NO registry
 #             credentials.
 #
-#   publish   Runs once, after both architectures pass. This is the only job
-#             with Docker Hub credentials, so the token never exists in a job
-#             that is executing a freshly built image, and never exists at all
-#             unless both architectures passed. It pushes ONLY the
-#             commit-addressed candidate tags:
+#   publish   Runs once, after both architectures pass. The only job with
+#             Docker Hub credentials; the token never coexists with a job
+#             executing a freshly built image. Pushes ONLY commit-addressed
+#             candidate tags (sha-<full-commit>{,-amd64,-arm64}), verifies the
+#             tag ANONYMOUSLY, then signs the index digest (KMS via OIDC,
+#             Rekor, tag-based .sig).
 #
-#               sha-<full-commit>-amd64
-#               sha-<full-commit>-arm64
-#               sha-<full-commit>          (multi-arch manifest list)
+#   propagate Second approval (ghcr environment): copies image + signature to
+#             GHCR and ECR Public via the copy-artifact composite action —
+#             idempotent, fail-closed, digest-asserted. GHCR writes are gated
+#             by the ghcr environment's own reviewers; the ECR role's OIDC
+#             trust covers both protected environments.
 #
 # Trigger: manual workflow_dispatch only. There is intentionally no
 # `push: tags:` trigger: a tag-triggered run executes the workflow definition
@@ -261,9 +266,22 @@ jobs:
   publish-candidate:
     needs: [gate, build]
     runs-on: ubuntu-latest
-    timeout-minutes: 30
-    environment: dockerhub       # the single approval gate, and the only job with secrets
+    timeout-minutes: 45
+    environment: dockerhub       # approval 1: publish + sign (the only job with Docker Hub secrets)
+    # Signing rides the publish approval (release owner decision 2026-08-18);
+    # propagation is the next job, under the ghcr environment's own gate.
+    outputs:
+      digest: ${{ steps.push.outputs.digest }}
+      candidate: ${{ steps.push.outputs.candidate }}
+      sig-tag: ${{ steps.sign.outputs.sig-tag }}
+      sig-digest: ${{ steps.sign.outputs.sig-digest }}
+    permissions:
+      contents: read
+      id-token: write   # KMS signing role (OIDC)
     steps:
+      - name: Check out (composite signing action lives in-repo)
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
       - name: Download the tested images
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093  # v4.3.0
         with:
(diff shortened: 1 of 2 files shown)


LeeroyHannigan (commented):
Reviewed in depth. This is merge-safe once rebased, and the design is right: signing and propagation ride the existing dockerhub approval, permissions are job-scoped to exactly the publishing job, the INDEX digest is what gets signed, and nothing here touches the latest/no-overwrite guards in promote-image.

First, rebase onto main: five of the six commits are the #279 signing infra, now byte-identical to main, so they collapse to nothing and the reviewable delta becomes just 4f640ba (release-image.yml, +104/-8). I'm holding approval until then only because a rebase push would dismiss it anyway.

One thing to fix before this is relied on operationally:

- The no-overwrite guard exits 1 whenever the candidate tag exists, but this PR appends the failure-prone tail (two OIDC assumes, cosign+Rekor, two cross-registry copies) after the push. Any post-push failure leaves the candidate published-but-unsigned and blocks every re-dispatch at the guard, so recovery means manually deleting the Docker Hub tag. The guard's own comment says identical-artifact re-runs are permitted; make the code do that: if the candidate exists and its digest matches the built artifact, skip the push and continue. The propagate loops are already idempotent, they're just unreachable on a retry today.

Smaller items, none blocking: copy the image leg by @digest rather than tag (promote-image already does; the .sig leg stays tag-addressed), route `${{ github.actor }}` (L375) through env like everything else, capture the digest from the create output instead of re-inspecting the tag, and the checkout pin comment on L275 says v4.2.2 for a SHA that's v4.4.0 elsewhere.

Ping me after the rebase and I'll approve.

LeeroyHannigan (dismissed):

.github/workflows/release-image.yml
```diff
@@ -325,6 +335,98 @@ jobs:
...
           echo "$OUT" | grep -q 'linux/arm64' || { echo "::error::arm64 missing"; exit 1; }
 
+      - name: Verify the candidate anonymously (registry-side, no credential)
+        env:
+          DIGEST: ${{ steps.push.outputs.digest }}
+        run: |
+          set -euo pipefail
+          # The pushed index, fetched as any user would fetch it, must be the
+          # digest we just recorded and contain both platforms.
+          RAW=$(skopeo inspect --raw "docker://${IMAGE_REPO}@${DIGEST}")
```
LeeroyHannigan: **Blocking:** this isn't anonymous and isn't a digest check.

`docker/login-action` at :293-297 already wrote creds to `~/.docker/config.json`, which skopeo reads by documented fallback, so this fetches as the maintainer PAT. And inspecting `@${DIGEST}` is content-addressed, so it can't detect the tag resolving elsewhere: it never looks at the tag. Since `DIGEST` came from this same index at :324, nothing new is proven.

That matters because it's the only gate between publish and "users can pull this", and it'd pass on a repo that isn't anonymously pullable.

```bash
GOT=$(skopeo inspect --no-creds --format '{{.Digest}}' "docker://${IMAGE_REPO}:${CANDIDATE}")
[[ "$GOT" == "$DIGEST" ]] || { echo "::error::tag resolves to $GOT"; exit 1; }
```

(needs `CANDIDATE` in the step env)

.github/workflows/release-image.yml
```diff
@@ -325,6 +335,98 @@ jobs:
...
+          SIG_DIGEST: ${{ steps.sign.outputs.sig-digest }}
+        run: |
+          set -euo pipefail
+          # Constant username: GHCR ignores it with a GitHub token, and it keeps
+          # the operator identity out of the run block.
+          skopeo login ghcr.io -u x-access-token --password-stdin <<< "$GHCR_TOKEN"
+          for pair in "${CANDIDATE}|${DIGEST}" "${SIG_TAG}|${SIG_DIGEST}"; do
+            TAG="${pair%%|*}"
+            WANT="${pair##*|}"
+            if EXISTING=$(skopeo inspect --format '{{.Digest}}' "docker://${GHCR_REPO}:${TAG}" 2>/dev/null); then
```
LeeroyHannigan: **Should-fix:** branches on exit code with stderr discarded, and `skopeo inspect` exits nonzero for blips and auth errors as well as 404. Any nonzero skips the different-digest check and copies; the post-copy assert then compares what was just written, so it always passes and hides that the guard was bypassed. Same at :413, and same pattern in #283's composite.

.github/workflows/release-image.yml
```diff
@@ -261,9 +261,19 @@ jobs:
...
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
     environment: dockerhub       # the single approval gate, and the only job with secrets
+    # One approval covers publish -> verify -> sign -> propagate (release
+    # owner decision 2026-08-18: signing rides the publish approval; the
+    # automated gates between are the safety net).
+    permissions:
+      contents: read
+      packages: write   # GHCR propagation
```
LeeroyHannigan: **Should-fix:** job-scoped, so every step carries these including summarise. Bigger point: a job holds one `environment`, this one is `dockerhub` (:265), but `ghcr-mirror.yml:69` gates GHCR on `ghcr`, so this path bypasses it and ECR gets its first automated write path with no environment of its own. The comment on :265 also still says "the only job with secrets".

.github/workflows/release-image.yml
```diff
@@ -261,9 +261,19 @@ jobs:
...
+      id-token: write   # KMS signing + ECR propagation (OIDC roles)
     steps:
+      - name: Check out (composite signing action lives in-repo)
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.2.2
```
LeeroyHannigan: **Nit:** this SHA is `v4.4.0`, not v4.2.2 (that's `11bd7190...`). The repo's other four call sites say `# v4.4.0`, and #290 is rewording a comment on this same SHA, so worth converging.

.github/workflows/release-image.yml
```diff
@@ -325,6 +335,98 @@ jobs:
...
+          DIGEST: ${{ steps.push.outputs.digest }}
+          CANDIDATE: ${{ steps.push.outputs.candidate }}
+          SIG_TAG: ${{ steps.sign.outputs.sig-tag }}
+          SIG_DIGEST: ${{ steps.sign.outputs.sig-digest }}
+        run: |
+          set -euo pipefail
+          # Constant username: GHCR ignores it with a GitHub token, and it keeps
+          # the operator identity out of the run block.
+          skopeo login ghcr.io -u x-access-token --password-stdin <<< "$GHCR_TOKEN"
+          for pair in "${CANDIDATE}|${DIGEST}" "${SIG_TAG}|${SIG_DIGEST}"; do
```
LeeroyHannigan: **Nit:** latent only, since I checked both sign outputs do exist. But if either were ever empty the pair becomes `"|"`, skopeo gets an empty tag and defaults to `latest`, which is the one tag this workflow says it never writes. `[[ -n "$TAG" && -n "$WANT" ]] || exit 1` covers it.

LeeroyHannigan (commented):
Signing inside the publish approval is the right move. One blocker inline, one below since the code is unchanged.

**Blocking:** the push guard at :311-314 exits 1 whenever the candidate tag exists, and that tag is created at :321 before sign, GHCR and ECR. So any failure past :321 strands a pushed-but-unsigned candidate that no re-dispatch can finish, and it makes the tolerate-existing branches at :382-383 dead code. The comment at :309-310 promises the opposite. Fix: if the tag exists, compare its digest to what was just built and fall through to sign/propagate when identical; only a different digest should error. Same gap you flagged as should-fix 2 on #281.

- **Should-fix:** this duplicates #283's `promote-registry` composite, and they already disagree (digest vs tag, unsigned-refusal vs not). They'll drift.
- **Nit:** headers at :4-6 and :29-33 still say the workflow stops at the candidate.
- **Nit:** summarise step at :430 has no `if:`, so a partial-failure run produces no summary. `if: always()`.
- **Nit:** "stacked on #279" is stale, merge base is main's tip.

Checked and fine: the two sequential OIDC assumptions are correct, and `sig-tag`/`sig-digest` do exist on the composite.

robinnsc:
All addressed in the latest commit:

**Blockers:**
- **Re-entrancy** — the push guard now falls through when the existing candidate holds exactly what this run built, compared by **per-arch image config digest** (the image ID that `docker save/load` preserves and the smoke test certified), with zero registry mutation before the comparison. A different artifact errors before anything is pushed; a tail failure is now finishable by re-dispatch, and the tolerate-existing propagate branches are reachable as designed. (Yes — same gap as my should-fix 2 on #281; consider this me taking my own medicine.)
- **The "anonymous" verify** — good catch, that gate was proving nothing twice over. Now `--no-creds` (you're right that skopeo reads the docker login by documented fallback) and **tag-addressed** with a digest equality assertion, so it actually checks what an anonymous user pulling the tag gets.

**Should-fixes:**
- **Duplication with #283's composite** — extracted a `copy-artifact` composite action (idempotent, three-way fail-closed probes — your :381 pattern fix — digest-asserted, empty-input guard for the :378 latent). The propagate legs are now four action calls. It's deliberately a *sibling* of `promote-registry` rather than a merge: copy-artifact moves candidates/sigs with no-overwrite semantics; promote-registry adds the signature-first + forward-only-latest release rules. Both live in `.github/actions/` with one probe idiom.
- **Environment bypass** — propagation is now its own job under the **ghcr environment**, matching promote-image's two-wave shape from #283: publish+sign under `dockerhub` (approval 1), propagate under `ghcr` (approval 2). The ECR role's OIDC trust already covers both protected environments (updated with read-back during #283's rework). The `permissions` scoping point resolves with the split — the publish job keeps only `id-token` for signing; `packages: write` exists only in the propagate job.

**Nits:** headers rewritten (the workflow no longer "stops at the candidate"), summary is `if: always()`, pin comment corrected to v4.4.0 (I'll converge #290's wording to match), and the stale "stacked on #279" opener is edited out of the description — merge base is main's tip, reviewable delta is the one workflow plus the new action.

The operator picture across #283+#284 is now symmetric: release = dockerhub approval (publish+sign) → ghcr approval (propagate); promotion = dockerhub approval (canary) → ghcr approval (mirrors). Two environments, each gating its own writes, four clicks per release end to end.

LeeroyHannigan (commented):
Re-reviewed at 954748b. Everything from the last round is addressed, and a couple of the fixes are better than what I suggested:

- The anonymous verify is now genuinely anonymous (`--no-creds`) and tag-addressed with a digest assertion - exactly the property the step name claims.
- The re-entrancy fix is the stronger option: instead of just tolerating an existing tag, it compares per-arch image config digests against what this run built and tested, with no registry mutation, and only then falls through to sign/propagate. A different artifact still hard-errors. Nice.
- Propagation moved to its own job under the `ghcr` gate, the `copy-artifact` composite carries the fail-closed three-way probe and the empty-input guard, headers are current, all five checkout pins read `# v4.4.0`, and the summary runs on failure.
- The cosign concern I raised on #283 doesn't apply here: `sign-image` installs a pinned, checksum-verified cosign itself. That's the pattern #283 should borrow.

Two tiny things, neither blocking:

- **Nit:** the PR title still says "(stacked on #279)" - the body was fixed but the title wasn't, and #279 has been on main for a day.
- **Note:** `probe()` is now copy-pasted between `copy-artifact` and #283's `promote-registry`. Fine for now; worth folding into one place whenever one of them next changes.

Good to go from my side.

LeeroyHannigan (approved):
LGTM

Issue #124, opened 2026-05-24
[Bug] Major Performance degradation with 20GB dataset. Timeouts with 200GB dataset.


RossWilliams:
## Describe the bug

Update operations become unreasonably slow if they change a GSI due to lack of an index. Problematic SQL missing indexes:

```sql
DELETE FROM "_ddb_<gsi_table>" WHERE base_pk = $1 AND base_sk_s = $2
```

At just 20GB I'm seeing updates take from 700ms-11s . At 200GB a single update_item call fails with a timeout. I'm using gsi_propagation_delay_ms=0.

## To Reproduce

1. Set gsi_propagation_delay_ms=0.
2. Create a table with 3 gsis
3. Load 20GB of data using all GSIs for the load
4. with concurrency of 5 start updating the existing data with update_item. The updates should update a gsi key.
5. Observe latency and max throughput

## Expected behavior

DynamoDB offers consistent performance for common production databases.

## Actual behavior

Over 30 seconds running:

Requests per second achieved: 4.4
p50 update latency: 730ms
p99 update latency: 11,677ms

A small database shows unreasonable performance degradation, likely severely impacting user-visible latency. A reasonably sized database will cease to operate.

## Environment

- ExtendDB version: 140a1e5
- Operating system: macOS 26
- Rust version (if building from source): 1.94
- Client SDK/driver and version: https 
- Deployment method (binary, container, source): source

## Checklist

- [x] I have searched existing issues for duplicates
- [x] I have included the ExtendDB version and environment details
- [x] I can reproduce this with the latest release

Issue #115, opened 2026-05-22
[Bug] GSI missing index for efficient propagation


parsnips:
## Describe the bug

ExtendDB provisions GSI (secondary-index) tables without an index on `(base_pk, base_sk_b)`. The engine then issues `DELETE FROM "<gsi-table>" WHERE base_pk = $1 AND base_sk_b = $2` to maintain the GSI on `UpdateItem` / `DeleteItem` of the base item. Because no index leads with those columns, every reverse-lookup DELETE falls through to a sequential scan of the entire GSI table.

In a Twisp benchmark with two GSI tables of ~66k rows each, this single statement accounted for **74.4% of all Postgres CPU time** (270 s of 363 s total in a 5×500-tx run), at 12.76 ms mean per call.

Adding `CREATE INDEX … (base_pk, base_sk_b)` by hand dropped the same DELETE from 12.76 ms to 0.021 ms (≈610× faster per call, ≈250× fewer buffers) and lifted the batched-write throughput of the surrounding workload by ~40% (89.4 → 125.3 tx/s).

The schema generator should provision this index alongside the existing primary key and the `(pk, sk_b, base_pk, base_sk_b)` secondary, since the engine itself depends on `(base_pk, base_sk_b)` lookups for GSI maintenance.

## To Reproduce

1. Start ExtendDB against Postgres in the default configuration; create a DynamoDB-style table with at least one GSI.
2. Populate it: enough rows to make a Seq Scan visible against an Index Scan (tens of thousands is enough; the regression is asymptotic in row count regardless).
3. Issue an `UpdateItem` against a base item in a way that triggers GSI maintenance (changes a GSI key, or deletes the item entirely).
4. Inspect the GSI table:

   ```sql
   \d "<gsi-table-name>"
   EXPLAIN (ANALYZE, BUFFERS)
     DELETE FROM "<gsi-table-name>"
     WHERE base_pk = 'nonexistent' AND base_sk_b = '\x00';
   ```

5. Observe a `Seq Scan` on the GSI table for the DELETE rather than an `Index Scan`.

Concrete observation from a local run (Postgres 16, table with 66,072 rows):

```
Delete on "_ddb_ecfa0b6e-…"  (cost=0.00..6799.08 rows=0 width=0) (actual time=20.421..20.421 rows=0 loops=1)
  Buffers: shared read=5808
[...]

PR #288, opened 2026-08-19
fix(sqlite): index GSI base-key columns for propagation deletes

LeeroyHannigan:
## What this does

Adds an index on `(base_pk, base_sk*)` to every GSI/LSI data table on the **SQLite** backend, at creation time.

This is the SQLite half of the defect diagnosed on PostgreSQL in #115 and #124. PR #127 fixes the PostgreSQL side; this is its counterpart. Neither closes the other's backend.

## Why

Index propagation deletes an item's old index row by base-table key:

```sql
DELETE FROM <idx_table> WHERE base_pk = ? AND base_sk_s = ?
```

There is no `pk` predicate, and both indexes on the table lead with `pk`:

| index | columns |
|---|---|
| `PRIMARY KEY` (autoindex) | `(pk, base_pk, base_sk*)` |
| `idx_order_*` | `(pk, sk*, base_pk, base_sk*)` |

So neither can serve the filter and SQLite plans a full `SCAN` of the index table on every propagating write. The cost grows with the size of the index rather than staying flat, which is the same asymptotic problem parsnips measured on PostgreSQL.

Measured on a faithful 66k-row replica of the generated schema, matching the row count in #115:

```
EXPLAIN QUERY PLAN DELETE FROM idx WHERE base_pk = ? AND base_sk_s = ?
  before -> SCAN idx                              6.621 ms mean
  after  -> SEARCH idx USING INDEX idx_base_key_* 0.003 ms mean
```

Same shape as the 12.76 ms to 0.021 ms that #115 measured on PostgreSQL.

## Two decisions worth reviewing

**Creation-time only, no migration.** There are no deployed SQLite databases carrying index tables, so nothing needs backfilling. This also avoids the implicit-rebuild-on-server-startup pattern that was objected to on the PostgreSQL side of this fix, where the reviewer asked for an operator-controlled migration instead. If a SQLite deployment ever does need retrofitting, the catalog-version path already exists.

**The index is unconditional**, not gated on the index declaring a sort key, unlike `idx_order_*` directly above it. The reverse lookup happens either way, and a sort-key-less index gets no ordering index at all, so for that shape this is the only secondary index on the table. There is a test for exactly that case.

## Testing done

Four new tests assert **the query plan SQLite actually chooses**, rather than merely that an index exists by name:

- composite base key
- hash-only base key
- an index with no sort key (the case where this is the only secondary index)
- column order: an index on `(base_sk_s, base_pk)` would satisfy the first three but would not serve a `base_pk`-only lookup, so the order is pinned via `PRAGMA index_info`

**Negative control.** Removing the `CREATE INDEX` execution fails all four with `got plan: SCAN _ddb_probe`, so they discriminate rather than pass vacuously. Re-proven after rebasing onto current `main`, since the vector merge changed this file by +165 lines.

**Live wiring check**, because the unit tests exercise the DDL function but not that it reaches it. Against a running dev-mode server on an isolated catalog, `CreateTable` with a GSI produced this in the real database file:

```
idx_base_key_<id> | CREATE INDEX "idx_base_key_<id>"
                    ON "_ddb_<id>" (base_pk, base_sk_s)

plan read back from that file:
  SEARCH _ddb_<id> USING INDEX idx_base_key_<id> (base_pk=? AND base_sk_s=?)
```

Gates:

```
cargo fmt --all -- --check                                                   # exit 0
cargo clippy --all-targets -- -D warnings                                    # exit 0
cargo clippy --all-targets --no-default-features --features sqlite -- -D warnings  # exit 0
cargo test --workspace     # 1019 passed; 0 failed; 0 filtered out
```

## Also in this diff

The doc comment on `create_index_data_table` stated the primary key as `(pk, sk*, base_pk, base_sk*)`. It does not include the index sort keys: `pk_cols` is `pk`, `base_pk`, then the base sort keys. The base-table key is what makes a row unique. Corrected, and both secondary indexes are now documented there.

## Checklist

- [x] Clippy is clean (`-D warnings` on both feature sets)

ADR / RFC: n/a — no change to the wire protocol, `Storage` trait, auth model, or public CLI surface. The on-disk change is additive and applies only to newly created index tables.

## Breaking changes

None. New index tables gain an index; existing tables and all query results are unchanged.

---

### DIFF
--- a/crates/storage-sqlite/src/data/ddl.rs
+++ b/crates/storage-sqlite/src/data/ddl.rs
@@ -222,8 +222,14 @@ impl SqliteEngine {
     }
 
     /// Create a GSI/LSI data table. GSI keys are not unique, so the primary key
-    /// is `(pk, sk*, base_pk, base_sk*)` to keep one row per base item, and an
-    /// ordering index covers `(pk, sk*, base_pk, base_sk*)`.
+    /// is `(pk, base_pk, base_sk*)`: the base-table key is what makes a row
+    /// unique, keeping one row per base item per index partition.
+    ///
+    /// Two secondary indexes are created:
+    /// - `idx_order_*` on `(pk, sk*, base_pk, base_sk*)`, for sort-key ordering
+    ///   within an index partition. Only when the index declares a sort key.
+    /// - `idx_base_key_*` on `(base_pk, base_sk*)`, for the reverse lookup that
+    ///   index propagation uses to delete an item's old index row. Always.
     pub(crate) async fn create_index_data_table(
         tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>,
         index_id: &str,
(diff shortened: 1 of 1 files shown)


jcshepherd (approved):
LGTM. It might be worth considering whether using INDEXED BY "hints" in the SQLite index entry delete queries would be helpful to better ensure the queries reliably leverage the index. Fine waiting until there's an actual problem that needs to be solved.

PR #290, opened 2026-08-19
ci(sign): break-glass signer accepts an allowlisted image choice

robinnsc:
## What

Adds an `image` **choice** input (`extenddb-postgres` | `extenddb-dev`, default postgres) to the break-glass `sign-image` workflow and derives the repo reference from it. A closed choice list, not free text — nothing user-controlled reaches the registry reference beyond the two allowlisted names. Also corrects the checkout pin's stale version annotation.

## Why

Recovery gap surfaced reviewing #281: the dev image signs inside its single-dispatch release, so if that run fails between publish and sign, the standalone signer is the recovery path — and it was hardcoded to the postgres repo. Complements (not replaces) the idempotent re-entry suggested on #281 itself.

## Testing done

- `yaml.safe_load` passes; input reaches the run only via the choice list; digest/release-tag validation unchanged.
- Default preserves existing behavior exactly — a dispatch that omits the input signs extenddb-postgres as today.

## Checklist

- [ ] Tests / fmt / clippy — not applicable, workflow-only change
- [x] Documentation — header semantics unchanged; input self-describing
- [x] Breaking changes noted below

## Breaking changes

None.

---


.github/workflows/sign-image.yml
```diff
@@ -23,6 +23,13 @@ on:
       release_tag:
         description: 'The release this digest belongs to, e.g. v0.1.5 (annotation only)'
         required: true
+      image:
+        description: 'Which image the digest belongs to'
+        type: choice
+        options:
+          - extenddb-postgres
+          - extenddb-dev
```
LeeroyHannigan, line 39: **Blocking:** a decision rather than a defect, so it needs your call not mine.

Adding dev here means the prod key at :45 signs it too, and key-based verify can't distinguish them: cosign ignores `docker-reference` and binds on digest, and the `org.extenddb.release` annotation isn't enforced unless a verifier passes `-a`. So a dev digest passes `cosign verify --key <release-pubkey>` exactly like a real release, for an image that's plain HTTP with open authorization and isn't built by `release-image.yml`.

Three options: separate key selected off `inputs.image`; drop dev until the dev release path exists; or keep one key deliberately and update the published verify instructions to require `-a org.extenddb.release=vX.Y.Z`.

Separately, dev is non-functional today regardless: `extenddb/extenddb-dev` is 404 on Docker Hub, and since the gate runs after the environment approval, a dev dispatch burns a two-person approval and then dies on an opaque skopeo error.

.github/workflows/sign-image.yml
```diff
@@ -32,7 +39,9 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  IMAGE_REPO: docker.io/extenddb/extenddb-postgres
+  # inputs.image is constrained by the choice list above — a closed allowlist,
+  # not free text — so this concatenation cannot be steered anywhere else.
+  IMAGE_REPO: docker.io/extenddb/${{ inputs.image }}
   KEY_ARN: arn:aws:kms:us-east-1:975306274793:key/654ab7f4-4a11-45e1-a313-3a6989f3721c
```
LeeroyHannigan, line 62: **Nit:** worth a comment noting this key is shared across every image in the choice list, whichever way the question above lands. It currently reads as "the signing key" rather than "the key that vouches for both a release and a dev build".

.github/workflows/sign-image.yml
```diff
@@ -32,7 +39,9 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  IMAGE_REPO: docker.io/extenddb/extenddb-postgres
+  # inputs.image is constrained by the choice list above — a closed allowlist,
+  # not free text — so this concatenation cannot be steered anywhere else.
```
LeeroyHannigan: **Nit:** probably true but undocumented. The REST endpoint does seem to reject out-of-list values, but the sources are a cli/cli issue and a community thread; GitHub's docs describe `type: choice` without mentioning API-side enforcement. Since the comment is asserting a security property, a three-line `case` on `inputs.image` makes it local and costs nothing. Same applies to #283.

.github/workflows/sign-image.yml
```diff
@@ -70,7 +79,7 @@ jobs:
           grep -q '"architecture": *"amd64"' <<< "$RAW" || { echo "::error::amd64 missing from index"; exit 1; }
           grep -q '"architecture": *"arm64"' <<< "$RAW" || { echo "::error::arm64 missing from index"; exit 1; }
 
-      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.2.2
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # releases-v4 backport head, same pin as release-image.yml
```
LeeroyHannigan: **Nit:** good catch that `# v4.2.2` was wrong, that SHA is `v4.4.0`. Only thing is the new wording disagrees with the repo's four other call sites, which all say `# v4.4.0`, and "backport head" moves while the pin doesn't. #284 adds a fresh `# v4.2.2` on this same SHA too.

LeeroyHannigan (commented):
Parameterisation of the signing path is clean and I found no correctness defects in the twelve lines. The multi-arch gate would accept a dev index fine.

One decision for you inline on the `extenddb-dev` option: the prod key would now sign the dev image too, and key-based verify can't tell them apart.

- **Should-fix:** `:38` is still `group: sign-image-extenddb-postgres`, the one image-identity site the diff missed. Dev and postgres signs now serialise under a postgres-named group, so you can't sign both at once, which a key rotation would need. `group: sign-image-${{ inputs.image }}`.
- **Should-fix:** no `reason` input, so a break-glass sign records what but not why. A required `reason` echoed to the summary and passed as an extra `-a` annotation would fix it.
- **Note:** still has a job after #284 lands (key rotation, non-pipeline digests, recovery when publish-stage signing failed after the push). Header and summary don't mention dev though.
- **Note:** worth confirming out of band that the `dockerhub` environment restricts deployment branches to main and the OIDC trust pins the ref. `workflow_dispatch` runs the definition from the dispatched ref, so :51's "main only" is a claim this file can't enforce. Not created by this PR, but it widens what the key signs.

robinnsc:
All applied in 0740302:

- **The key decision** — you're right that it's a decision, not a defect, and it isn't mine to make either: dev is **out of the choice list** pending the release owner's call, with your three options recorded in the inline comment where the next person will trip over it. My own recommendation for when it's decided: option 1 (separate dev key) — the dev/prod key separation already exists in our KMS setup for exactly this reason, and "the release key only ever attests release-pipeline artifacts" is a property worth keeping absolute rather than annotation-conditional. Also noted: dev was non-functional here anyway (no Docker Hub repo, and the gate ordering would burn an approval to find out).
- **Concurrency** — `group: sign-image-${{ inputs.image }}`; a key rotation can sign images in parallel.
- **Reason** — required input, validated non-empty, echoed to the summary **and signed into the payload** as `org.extenddb.break-glass-reason` (the composite action gained an optional validated `extra-annotation` input), so the key attests why the break-glass path ran, not just what it signed.
- **Choice-list enforcement** — local `case` in the first gate; the comment no longer asserts an undocumented API property. Agreed on #283 — same treatment there in its next push.
- **Pin comment** — converged to `# v4.4.0`; #284's fresh `# v4.2.2` was corrected in its last push too.
- **Header** — documents the post-#284 role (rotation, non-pipeline digests, finishing a publish-stage signing failure).

**Out-of-band check you asked for, done:** the `dockerhub` environment has exactly one deployment branch policy, `main` (branch type), verified via the API just now — so a dispatch from any other ref can't enter the environment, and the OIDC sub the role trusts (`environment:dockerhub`) is only mintable by runs that passed that policy plus reviewer approval. The ":51 main only" claim is enforced by the environment, which the file can reference but not prove — agreed that's worth knowing rather than assuming.

robinnsc:
Key decision landed (release owner, today): **one key for every image, all tiers**, with two recorded conditions — verify/admission guidance must distinguish images by the signed annotations or repo (never key alone; a bare key-match admits any ExtendDB build including dev), and the key is explicitly slated to be phased out for **keyless workflow-identity signing**, which makes the tier distinction automatic and deletes the key-management question entirely. Reflected in the workflow comments in the latest commit. Un-parks your inline decision item; dev stays out of the choice list only for the practical reason (repo 404 / no dev release path yet) and rejoins when #281 lands.

LeeroyHannigan (commented):
Re-reviewed at b1d2319. The key question got a real decision, recorded where it needs to be (choice-list comment, KEY_ARN comment, and the d8ae412 commit): one key attests "official ExtendDB build", the signed annotations attest which, admission policies must match annotations/repo rather than key alone, keyless planned. That settles it from my side. The dev option is also genuinely functional now that `extenddb/extenddb-dev` exists on Docker Hub (checked, 200).

Everything else from the last round landed: per-image concurrency group, required `reason` validated non-empty and signed into the payload via the new `extra-annotation` input (nicer than what I asked for), local `case` allowlist, and the checkout pin now reads `# v4.4.0`.

One should-fix inline in the new code, one-line fix.

- **Note:** the annotation-matching requirement ("never key alone") now needs to reach the published verify instructions when the next release notes go out - already on the tracked list with the cosign.pub publication, so just flagging the connection.

.github/workflows/sign-image.yml
```diff
@@ -103,10 +142,11 @@ jobs:
       - name: Summarise
         run: |
           {
-            echo "### Docker Hub image signed (break-glass path)"
+            echo "### ${{ inputs.image }} signed (break-glass path)"
             echo ""
             echo "| field | value |"
             echo "|---|---|"
+            echo "| reason | ${{ inputs.reason }} |"
```
LeeroyHannigan: **Should-fix:** free-text `inputs.reason` is interpolated straight into this run block's script source, which is the classic workflow_dispatch script-injection shape - and this step still holds the signing role's ambient AWS creds. It also breaks the convention this repo states itself (promote-image: "Dispatch inputs are untrusted; they reach the shell only through the environment"). The gate step already routes it through `RAW_REASON`; do the same here: `env: REASON: ${{ inputs.reason }}` and `echo "| reason | ${REASON} |"`. (Line 145's `inputs.image` is fine - allowlist-validated.)

robinnsc:
Fixed — the summary's `reason` echo now goes through `env` like everything else (the composite-action path was already env-routed and format-validated, so the summary was the one leak). Agreed on the verify-instructions connection: the annotation-matching requirement is tracked for the next release notes alongside the pubkey publication.

LeeroyHannigan (approved):
LGTM

PR #293, opened 2026-08-19
docs: allow a backend to factor internal crates beneath its own directory

yesyayen:
Four one-clause amendments to `docs/rfcs/0002-backend-plugin-policy.md`, so that a backend may factor internal crates beneath its own directory.

- `:31` and `:224` permitted exactly one crate per backend. The permission is now explicit, and the published crate still lives at `crates/storage-{backend}/`.
- `:246` said backend crates "depend only on the `storage` trait and common utilities, not on each other", which prohibits a backend depending on its own internal crates. Re-scoped to another backend's crates.
- `:257` promised semver on all crates under a future crates.io model. Internal crates are `publish = false` and carry no semver commitment.

**Nothing downstream of backend selection changes.** `crates/bin/Cargo.toml` keeps `sqlite = ["extenddb-storage-sqlite"]` and `sqlite-memory = ["sqlite", "extenddb-storage-sqlite/memory"]` unchanged, because the outer crate keeps its name and path.

**No enforced check is affected.** The dependency clause states it "will be validated via `cargo build --no-default-features --bin extenddb`", which can only show that the front end builds with no backend and cannot detect a dependency between backend crates. That command appears nowhere under `.github/` or `devtools/`, so this corrects unenforced prose rather than relaxing a gate.

This travels separately from #292, which is the design doc that needs it, so that the policy change is reviewed on its own terms.

### DIFF
--- a/docs/rfcs/0002-backend-plugin-policy.md
+++ b/docs/rfcs/0002-backend-plugin-policy.md
@@ -28,7 +28,7 @@ Without formal policies, we risk inconsistent quality, fragmented implementation
 
 ### Repository structure: Mono-repo with feature flags
 
-All officially supported backends live in the main `extenddb` repository as separate crates under `crates/storage-{backend}/`. Examples:
+All officially supported backends live in the main `extenddb` repository as separate crates under `crates/storage-{backend}/`. A backend may factor internal crates beneath its own directory (for example `crates/storage-{backend}/exec/`) when that separation enforces something the backend needs; the published backend crate stays at `crates/storage-{backend}/`. Examples:
 
 - `crates/storage-postgres/` (reference implementation)
 - `crates/storage-cassandra/`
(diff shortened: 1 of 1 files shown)


LeeroyHannigan (commented):
Checked the claims and they hold: `cargo build --no-default-features --bin extenddb` appears nowhere under `.github/` or `devtools/`, so this corrects unenforced prose rather than relaxing a gate, and splitting the policy change out of #292 is the right call. The four amendments are surgical. One wording gap inline, and it's the substance of the PR so worth getting right.

- **Note:** RFC-0002 is still `Status: Draft`, so this doesn't need the FCP machinery - but it's @jcshepherd's draft, so his ack is the one that matters here.

docs/rfcs/0002-backend-plugin-policy.md
```diff
@@ -243,7 +243,7 @@ To prevent accidental coupling between frontend and specific backend implementat
 
 1. The `extenddb` service binary builds successfully without any backend feature flags enabled (frontend depends only on `storage` trait crate, not concrete backends).
 
-2. Backend crates depend only on the `storage` trait and common utilities, not on each other.
+2. Backend crates depend only on the `storage` trait and common utilities, and not on another backend's crates. A backend's own internal crates may depend on each other.
```
LeeroyHannigan, line 246: **Should-fix:** strictly read, this still forbids the thing the amendment exists to permit. "Backend crates depend only on the `storage` trait and common utilities" excludes a backend's own internal crates, so `crates/storage-sqlite/` depending on `crates/storage-sqlite/exec/` remains prohibited by the first sentence - the second sentence only lets internal crates depend on *each other*. Suggest: "Backend crates depend only on the `storage` trait, common utilities, and crates internal to the same backend - never on another backend's crates."

LeeroyHannigan (approved):

PR #294, opened 2026-08-19
chore: bump workspace version to 0.1.7

LeeroyHannigan:
## What

Workspace version 0.1.6 -> 0.1.7, with **both** licence notices files regenerated in the same commit. The notices embed every workspace crate version, so a bump without regeneration is the exact failure that made 0.1.6 unreleasable (#271, fixed by #272). `licenses.yml` gates both files at PR time now, and the release workflows gate them again at dispatch.

## Why 0.1.7

`v0.1.6` already exists and points 13+ commits behind current main, predating vector search entirely. The `release-dev-image` gate requires the dispatched tag to match `Cargo.toml` at the tagged commit, so shipping the dev image with vector search requires this bump, then a `v0.1.7` tag on the merged commit.

## Stacked on #281 — merge order matters

Base is `feat/dev-container`, deliberately: the DEV notices generator only exists on that branch, and regenerating the DEV notices against 0.1.7 requires it. **Merge #281 first.** GitHub retargets this PR to `main` automatically when #281 lands — then **verify the base actually shows `main` before merging this** (the #274 stranding lesson: merging within seconds of the parent can land the child on the dead parent branch).

## Verification

- `generate-software-license-notices --check` exit 0
- `generate-dev-license-notices --check` exit 0
- `cargo check` clean on `sqlite,dev-mode` and `postgres`
- the release gate comparison (tag `v0.1.7` vs workspace `Cargo.toml`) now matches

## Heads-up, not this PR's problem

`packaging/npm/package.json` on #258 still says 0.1.3; the npm publish gate will refuse that drift against a 0.1.7 workspace the moment #258 merges. #258 needs its manifest bumped before or at merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1107,7 +1107,7 @@ dependencies = [
 
 [[package]]
 name = "extenddb"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "anyhow",
  "extenddb-app",
@@ -1120,7 +1120,7 @@ dependencies = [
 
 [[package]]
 name = "extenddb-app"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -1148,7 +1148,7 @@ dependencies = [
 
 [[package]]
 name = "extenddb-auth"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "async-trait",
  "axum",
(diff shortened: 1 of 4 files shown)


LeeroyHannigan:
Closing in favour of #296, which is cut off `main`.

This branch was cut off `feat/dev-container` one minute before #281 merged, when the DEV notices generator was not yet on main. #281 then merged through the REBASE queue, so its commits were rewritten and `b900ec6` is no longer an ancestor of main — retargeting this PR would render it as 15 commits instead of 1. All prerequisites are on main now, so a fresh cut is the clean fix.

PR #296, opened 2026-08-19
chore: bump workspace version to 0.1.7

LeeroyHannigan:
## What

Workspace version 0.1.6 -> 0.1.7, with **both** licence notices files regenerated in the same commit. They embed every workspace crate version, so a bump without regeneration is the exact failure that made 0.1.6 unreleasable (#271, fixed by #272). `licenses.yml` gates both at PR time; the release workflows gate them again at dispatch.

## Why now

This unblocks the `extenddb-dev` image release. `v0.1.6` already exists and points behind the vector search merges, and `release-dev-image`'s gate requires the dispatched tag to match `Cargo.toml` at the tagged commit — so the dev image with vector search needs this bump, then a `v0.1.7` tag on the merged commit.

## Verification

- `generate-software-license-notices --check` exit 0
- `generate-dev-license-notices --check` exit 0
- `cargo check` clean on `sqlite,dev-mode` and `postgres`

## Note

Supersedes #294, which was cut off `feat/dev-container` one minute before #281 merged. This one is cut off `main`.

`packaging/npm/package.json` on #258 still says 0.1.3 and will fail the npm publish drift gate against a 0.1.7 workspace — needs bumping before or at that merge, not here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1107,7 +1107,7 @@ dependencies = [
 
 [[package]]
 name = "extenddb"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "anyhow",
  "extenddb-app",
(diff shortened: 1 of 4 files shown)


amrith (approved):
sounds like 7 comes after 6, so this must be good?

PR #297, opened 2026-08-21
ci(dev-image): propagate to GHCR and ECR with registry-state re-entry

robinnsc:
## What

`release-dev-image` now covers all three registries in its single dispatch:
after the existing publish+sign job (`dockerhub` environment, approval 1), a
new `propagate` job (`ghcr` environment, approval 2) carries the version
tag, a **state-derived** `latest`, and the signature to GHCR and ECR Public
through the `promote-registry` composite action — signature verified at
each destination before any tag is placed, forward-only `latest` re-guarded
per destination, three-way fail-closed probes throughout.

The publish job also becomes **re-entrant on registry state**: an existing
version tag whose signature verifies is treated as authoritative (skip
push, skip sign); existing-but-unsigned proceeds to signing only; and only
a definite manifest-unknown proceeds to push. Signing is additionally
guarded by a signature-existence check, so a re-dispatch never double-signs.

`move-latest` for the mirrors is derived from Docker Hub state (does
`latest` point at this digest?) rather than re-deciding, so mirrors always
converge on what the canonical registry serves.

## Why

Release-owner decision 2026-08-20: the dev image lands on the same three
registries as the Postgres image (superseding the 08-15 Docker Hub-only
scoping — and matching the three-registry expectation from the #283 review).
The AWS side is already live and read-back-verified: `extenddb-dev` exists
on ECR Public with catalog metadata, and both ECR role policies cover it.

Re-entry has to be registry-state-based rather than #284-style artifact
comparison because this workflow rebuilds on every dispatch and the build
is not bit-reproducible — comparing a fresh build against the released
artifact would refuse every time. An existing version tag with a verifying
signature *is* the release; immutability says never to touch it anyway.

**Immediate use — the v0.1.7 backfill:** dev 0.1.7 is Docker Hub-only
today. After this merges, re-dispatching `v0.1.7` no-ops publish and
signing (tag exists, signature verifies) and runs the propagate wave,
putting the already-released artifact on GHCR and ECR. The first GHCR push
auto-creates the package: verify repo linkage and visibility per runbook
10.6 (the summary reminds the operator).

## Testing done

- `yaml.safe_load` passes.
- The propagate legs are the `promote-registry` action reviewed and merged
  in #283 (four hostile rounds), pointed at the dev repositories; the AWS
  prerequisites were verified by read-back and the policy diffs archived.
- Re-entry decision table traced against live state: v0.1.7 on Docker Hub
  (exists + signed → skip to propagate ✓); a fresh version (absent →
  normal path ✓); the interrupted case (exists + unsigned → sign only ✓);
  probe failure (abort ✓).
- Not executable pre-merge (main-only environment). The v0.1.7 backfill
  dispatch is the live verification, and every gate fails closed.

## Checklist

- [ ] Tests / fmt / clippy — not applicable, workflow-only change
- [x] I have added or updated tests for new functionality — the re-entry decision table and per-destination gates are the checks; live verification plan above
- [x] I have updated documentation if behavior changed — header rewritten; runbook records the decision and completed AWS work

## Breaking changes

None. A normal first-time dispatch behaves as before, plus the propagate wave.

---

### DIFF
--- a/.github/workflows/release-dev-image.yml
+++ b/.github/workflows/release-dev-image.yml
@@ -278,22 +278,44 @@ jobs:
         env:
           SHA: ${{ needs.gate.outputs.sha }}
           VERSION: ${{ needs.gate.outputs.version }}
+          COSIGN_VERSION: v3.1.3
+          COSIGN_SHA256: 4629c757b7618056f8ddd7e2625ae9fdd94c0372a65049520bc7d9df9efc7f71
         run: |
           set -euo pipefail
           REPO='${{ env.IMAGE_REPO }}'
           CANDIDATE="sha-${SHA}"
 
           # A released version tag is immutable. If it exists AND its
-          # signature verifies against the committed release key, the
+          # signature VERIFIES against the committed release key, the
           # published artifact is authoritative: skip pushing and signing
           # (re-dispatch recovery / mirror backfill). Exists but UNSIGNED:
-          # fall through to signing only. Probe failures other than a
-          # definite absence abort (fail closed).
+          # fall through to signing only. Exists with a signature that fails
+          # verification: abort loudly — skipping would report success on an
+          # artifact that can never propagate (the destination verify would
+          # refuse it forever). Probe failures other than a definite absence
+          # abort. Captured output is shape-validated because 2>&1 merges
+          # skopeo warnings into successful captures.
           if OUT=$(skopeo inspect --no-creds --format '{{.Digest}}' "docker://${REPO}:${VERSION}" 2>&1); then
+            if [[ ! "$OUT" =~ ^sha256:[0-9a-f]{64}$ ]]; then
+              echo "::error::digest capture for ${REPO}:${VERSION} is polluted or malformed: ${OUT}"
+              exit 1
+            fi
             EXISTING="$OUT"
             SIG_TAG="sha256-${EXISTING#sha256:}.sig"
             if skopeo inspect --no-creds --format '{{.Digest}}' "docker://${REPO}:${SIG_TAG}" >/dev/null 2>&1; then
-              echo "::notice::${REPO}:${VERSION} already published and signature artifact present; treating published state as authoritative"
+              # The signature must VERIFY, not merely exist: a wrong-key or
+              # garbage manifest at the tag name must abort, not skip.
+              if ! command -v cosign >/dev/null || [[ "$(cosign version 2>/dev/null | awk '/GitVersion/{print $2}')" != "$COSIGN_VERSION" ]]; then
+                curl -sSfL -o /tmp/cosign \
+                  "https://github.com/sigstore/cosign/releases/download/${COSIGN_VERSION}/cosign-linux-amd64"
+                echo "${COSIGN_SHA256}  /tmp/cosign" | sha256sum --check --strict
+                sudo install -m 0755 /tmp/cosign /usr/local/bin/cosign
+              fi
+              if ! cosign verify --key extenddb-signing.pub.pem "${REPO}@${EXISTING}" >/dev/null; then
+                echo "::error::${REPO}:${VERSION} carries a signature artifact that does NOT verify against the release key; refusing to skip. Investigate before re-dispatching."
+                exit 1
+              fi
+              echo "::notice::${REPO}:${VERSION} already published and its signature verifies; treating published state as authoritative"
               echo "digest=$EXISTING" >> "$GITHUB_OUTPUT"
               echo "already-published=true" >> "$GITHUB_OUTPUT"
               exit 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release-dev-image.yml
```diff
@@ -270,9 +283,27 @@ jobs:
...
+          # signature verifies against the committed release key, the
+          # published artifact is authoritative: skip pushing and signing
+          # (re-dispatch recovery / mirror backfill). Exists but UNSIGNED:
+          # fall through to signing only. Probe failures other than a
+          # definite absence abort (fail closed).
+          if OUT=$(skopeo inspect --no-creds --format '{{.Digest}}' "docker://${REPO}:${VERSION}" 2>&1); then
+            EXISTING="$OUT"
+            SIG_TAG="sha256-${EXISTING#sha256:}.sig"
+            if skopeo inspect --no-creds --format '{{.Digest}}' "docker://${REPO}:${SIG_TAG}" >/dev/null 2>&1; then
+              echo "::notice::${REPO}:${VERSION} already published and signature artifact present; treating published state as authoritative"
```
LeeroyHannigan: **Blocking:** the skip decision checks the sig *tag exists*, but the comment and PR body promise the signature *verifies* - same shape as #283's S1, which you fixed there with real `cosign verify`. A wrong-key manifest at that tag name skips signing, reports success, and can never self-heal (every re-dispatch skips again; propagate's destination verify then fails forever). Install pinned cosign (the block sign-image already carries) and gate the skip on `cosign verify --key extenddb-signing.pub.pem docker.io/${REPO}@${EXISTING}`; exists-but-fails-verify should abort loudly.

.github/workflows/release-dev-image.yml
```diff
@@ -270,9 +283,27 @@ jobs:
...
-          if docker buildx imagetools inspect "${REPO}:${VERSION}" >/dev/null 2>&1; then
-            echo "::error::${REPO}:${VERSION} already exists; refusing to overwrite a released tag."
+          # A released version tag is immutable. If it exists AND its
+          # signature verifies against the committed release key, the
+          # published artifact is authoritative: skip pushing and signing
+          # (re-dispatch recovery / mirror backfill). Exists but UNSIGNED:
+          # fall through to signing only. Probe failures other than a
+          # definite absence abort (fail closed).
+          if OUT=$(skopeo inspect --no-creds --format '{{.Digest}}' "docker://${REPO}:${VERSION}" 2>&1); then
+            EXISTING="$OUT"
```
LeeroyHannigan, line 303: **Should-fix:** `2>&1` merges stderr into `OUT` on *success* too, and skopeo emits warnings on successful calls - reproduced locally: one warning line makes `SIG_TAG` = `sha256-WARN[0000]....sig`, the release gets misclassified as unsigned, and the polluted digest lands in `GITHUB_OUTPUT`. Validate shape after capture: `[[ "$OUT" =~ ^sha256:[0-9a-f]{64}$ ]]` or abort.

.github/workflows/release-dev-image.yml
```diff
@@ -338,7 +383,103 @@ jobs:
...
+    steps:
+      - name: Check out (composite action lives in-repo)
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4.4.0
+
+      - name: Derive move-latest from Docker Hub state
+        id: ml
+        run: |
+          set -euo pipefail
+          if OUT=$(skopeo inspect --no-creds --format '{{.Digest}}' "docker://${IMAGE_REPO}:latest" 2>&1); then
+            [[ "$OUT" == "$DIGEST" ]] && MV=true || MV=false
```
LeeroyHannigan, line 446: **Should-fix:** same stderr-pollution as the publish probe, but here it fails *silent*: a warning-polluted `OUT` never equals `$DIGEST`, so `move-latest` quietly becomes false and the mirrors' `latest` is left behind with a green run. Same shape validation fixes it.

LeeroyHannigan (commented):
Read the whole workflow plus verified the live claims: ECR Public `extenddb/extenddb-dev` exists (empty, ready), the 0.1.7 sig tag is on Docker Hub so the backfill trace holds, and the OIDC trust already covers the `ghcr` environment from the #283 work. The state-derived `move-latest` is a nice piece of design - canonical registry decides, mirrors converge, and it's race-safe across the two approvals.

One blocking finding and one should-fix, both in the new probe shell. Detail inline.

robinnsc:
All three in the latest commit:

- **Blocking** — guilty as charged, and of my own #283 offense no less: the code checked existence while every comment promised verification. The skip decision now installs pinned cosign (same checksum pin as everywhere else) and requires `cosign verify --key extenddb-signing.pub.pem` against the existing digest; **exists-but-fails-verify aborts loudly** with an investigate-first message, since skipping would green-light an artifact the destination verify refuses forever. Sanity-checked the exact backfill case live: dev 0.1.7's signature verifies with the committed key, so the v0.1.7 re-dispatch takes the skip path as designed.
- **Publish probe pollution** — the captured digest is now shape-validated (`^sha256:[0-9a-f]{64}$`) before anything derives from it; polluted or malformed captures abort rather than fabricating a `.sig` tag name or landing in `GITHUB_OUTPUT`. Nice reproduction on the warning-line case — that one would have been miserable to debug from a distance.
- **move-latest pollution** — same shape validation; a polluted capture now fails the run instead of silently leaving the mirrors' `latest` behind under a green checkmark, which was the worse of the two pollution outcomes.

Appreciate the live verification of the ECR repo and trust state on your side — between that and the 0.1.7 sig check, the backfill dispatch should be uneventful in the good way.

LeeroyHannigan (commented):
Re-reviewed 936ed94 - all three findings closed, and closed well.

The skip path now runs a real `cosign verify` with the same pinned, checksum-verified cosign as the sign-image composite, and the exists-but-fails-verify case aborts loudly instead of skipping - better than what I asked for, since it makes the unrecoverable-wedge state impossible to enter silently. Both probe sites got the digest shape validation with the mechanism documented inline.

I also checked the one thing the fix could have broken: the verify reads `extenddb-signing.pub.pem` from the checkout, and this job checks out main (not the tag tree), so the v0.1.7 backfill dispatch works even though the key postdates the tag - no repeat of the npm-publish failure class.

Good to go from my side.

LeeroyHannigan (approved):

PR #302, opened 2026-08-22
fix: compile on Windows — gate POSIX daemon/syslog/signal code behind cfg(unix)

LeeroyHannigan:
## Why

npm-publish run [32528088299](https://github.com/ExtendDB/extenddb/actions/runs/32528088299) is the first run to ever pass the gate (thanks to #299 + #301), which means the `win32-x64` build job is the first thing to ever compile ExtendDB for Windows — and it can't. `extenddb-server` fails with 9 errors (`syslog_tracing`, `tokio::signal::unix`, raw `libc::syslog`), and behind it `daemonize` and `libc::kill` in `extenddb-app` fail the same way.

## What

The POSIX daemon subsystem is gated `cfg(unix)` rather than ported — the npm launcher (the only consumer of a Windows binary) always runs `serve --foreground`, which never touches any of it:

- **server**: `syslog-tracing`/`libc` → `[target.'cfg(unix)'.dependencies]`. Syslog writer arm and `log_to_syslog_raw` get unix impls; non-unix `log_to_syslog_raw` falls back to stderr (always attached without daemonize). `shutdown_signal` is cfg-split: Ctrl+C only on non-unix.
- **app**: `daemonize`/`libc` → unix-only deps. `serve` without `--foreground` bails on non-unix with a clear message; `stop` prints an honest unsupported message; `check_config_permissions` is a documented no-op (Windows ACLs, not mode bits); `is_process_alive` conservatively returns `true` on non-unix (guards keep refusing, PID files only exist in unix daemon mode).

## Verification

- `cargo check --target x86_64-pc-windows-msvc --no-default-features --features sqlite,dev-mode -p extenddb`: **clean** (C build scripts stubbed locally; the CI run already proved libsqlite3-sys/aws-lc-sys compile on the real Windows runner — the failures were all in our Rust source).
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, server+app test suites (52 tests): all pass on Linux.

## Release consequence

The build jobs compile from the **tag tree**, so no v0.1.8 dispatch can ever pass — the tag predates this fix. After this merges, npm publishing needs a 0.1.9 bump (both `Cargo.toml` **and** `packaging/npm/package.json`) and a fresh tag.

### DIFF
--- a/crates/app/src/serve_helpers.rs
+++ b/crates/app/src/serve_helpers.rs
@@ -4,9 +4,11 @@
 //! Helper functions for `extenddb serve`: daemonize health checks, PID file
 //! management, config permission checks, and syslog utilities.
 
+#[cfg(unix)]
 use std::path::PathBuf;
 
 /// Platform-appropriate hint for viewing syslog output.
+#[cfg(unix)]
 fn syslog_hint() -> &'static str {
     if cfg!(target_os = "macos") {
         "Check syslog: log show --predicate 'process == \"extenddb\"' --last 5m"
@@ -19,6 +21,7 @@ fn syslog_hint() -> &'static str {
 /// and then verifies the daemon process is still alive. This catches early
 /// startup failures (bad config, missing catalog tables, TLS cert errors)
 /// that would otherwise be invisible because stderr is /dev/null after fork.
+#[cfg(unix)]
 pub fn verify_daemon_started(pid_file: &PathBuf, bind_addr: &str) -> anyhow::Result<()> {
     let hint = syslog_hint();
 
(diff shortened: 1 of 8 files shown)


crates/app/src/util.rs
```diff
@@ -5,12 +5,23 @@
...
     // SAFETY: kill with signal 0 performs error checking without sending a
     // signal. Returns 0 if the process exists and we have permission.
     unsafe { libc::kill(pid, 0) == 0 }
 }
 
+/// Non-unix fallback: there is no dependency-free liveness probe, so assume
+/// the process is alive. This is the conservative direction for every caller:
+/// guards that refuse to run alongside a live server keep refusing, and
+/// `status` at worst reports a stale PID. PID files only exist in daemon
+/// mode, which is unix-only, so this path is effectively unreachable.
```
robinnsc, line 19: The "effectively unreachable" justification here is not quite right: PID files do not only exist in daemon mode. `--write-pid-file` writes one for a foreground server on any platform, and the README documents it precisely so `stop` and `status` can work against a foreground process. So on Windows the reachable sequence is: `serve --foreground --write-pid-file`, then a hard kill (which is what any supervisor stop is on Windows, the npm launcher's `stop()` included), leaving a stale PID file behind. With this fallback always answering true, `status` then reports the dead server as running and any liveness guard keeps refusing until someone deletes the file by hand.

To be clear, I would keep the behavior. Conservative is the right direction when there is no cheap probe. It is the comment I would fix: drop the unreachable claim, and say instead that a stale PID file on Windows misreports as alive until removed. If you want the probe to be real at some point, `OpenProcess` + `GetExitCodeProcess` does it without a new dependency via the `windows-sys` crate, but that is a follow up, not this PR.

crates/app/src/serve_helpers.rs
```diff
@@ -72,35 +75,50 @@ pub fn verify_daemon_started(pid_file: &PathBuf, bind_addr: &str) -> anyhow::Res
...
+                },
+            )?;
+        }
+
+        Ok(())
+    }
+    // Windows has no POSIX permission bits; access control is via ACLs, which
+    // have no meaningful 0600 equivalent to enforce here. The dev/CI use case
+    // on Windows (the npm launcher) runs with an in-memory backend and no
+    // config file, so this check is a no-op rather than a porting project.
```
robinnsc, line 118: Small accuracy point: the launcher runs file backed by default, not in memory. Memory mode being opt in was the deliberately argued contract decision in #258. The half of this sentence that actually carries the conclusion is "no config file" (the launcher configures through env vars from a scratch cwd), and that part is right, so the no-op stands. Just drop or fix the in memory claim so this does not contradict the launcher contract it cites.

robinnsc (approved):
The gate rather than port approach is the right call given the npm launcher is the only Windows consumer and always passes `--foreground`. What I checked and liked: the unix paths are wrapped verbatim rather than modified (including the daemonize control flow and panic hook placement in `cmd_serve.rs`), `LogTarget::Syslog` is handled totally on non unix with a bail instead of a silent fallback, the dependency moves use the proper `[target.'cfg(unix)'.dependencies]` mechanism, and the `shutdown_signal` comment is honest about `TerminateProcess` skipping the drain and why that is safe. Skipping the config permission check on Windows is a reasonable, documented tradeoff.

Approving with two accuracy findings inline. Neither changes behavior, but both comments will mislead the next reader, and one of them implies a doc gap: the main README's `--write-pid-file` paragraph promises `stop` and `status` work against a foreground server, and on Windows `stop` now unconditionally exits 1. A sentence there (or in troubleshooting) about the Windows situation would close the loop.

The release consequence note in the description is appreciated; agreed that v0.1.8 can never pass the gate and the fix rides 0.1.9 with a fresh tag.

PR #303, opened 2026-08-22
ci(dockerhub-description): publish automatically when an Overview file lands on main

LeeroyHannigan:
## Why

#289 shipped the Overview publisher, and the postgres front page then sat blank for two days — the workflow was `workflow_dispatch`-only, linked to nothing, and nobody remembered the follow-up dispatch (it had **zero runs** until yesterday's manual backfill). A fix that depends on remembering isn't a fix.

## What

- **New `push` trigger**: any change to `docker/dockerhub-overview*.md` landing on main queues a publish for exactly the image(s) whose file changed. Changed files come from the compare API (`before...after`), matched with `grep -qx` so `...md.bak`-style names can't false-match. Unresolvable ranges (forced push) publish both — idempotent, so over-publishing is harmless and beats silently skipping one.
- **The `dockerhub` environment gate is unchanged and applies on every path** — the trigger removes the remembering, not the review. `workflow_dispatch` stays for re-publishing without a content change.
- **`validate`/`publish` become a `fail-fast: false` matrix** over the planned image list, preserving #289's property that a missing Docker Hub repo fails one image alone. The image→file mapping is duplicated between the two jobs deliberately: matrix job outputs are last-writer-wins.

## Verification

- YAML parses.
- The plan derivation logic was executed locally under `set -euo pipefail` with synthetic data for all six cases: postgres-only, dev-only, both, fallback-on-empty, no-match (errors loudly), and the substring trap. Both grep short-circuit directions run standalone without tripping `set -e` (non-final AND-list commands are exempt).
- The credentialed publish/verify steps are unchanged from #289 apart from reading the matrix variable instead of the dispatch input.

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -7,27 +7,40 @@
 # repository metadata: the Overview lives on the Docker Hub repository object,
 # not on any image. GHCR derives its description from the OCI labels the
 # Dockerfiles already set, so the same image reads as documented there and blank
-# on Docker Hub. That is why this exists as its own step rather than falling out
-# of a push.
+# on Docker Hub. That is why this exists as its own workflow rather than falling
+# out of a push.
 #
 # Sourced from `docker/dockerhub-overview*.md` so the published Overview is
 # reviewed like any other content and cannot drift from the repository.
 #
-# Two jobs: `validate` is credential-free and ungated, so a dead link, an
-# oversize file, or a missing Docker Hub repository is discovered on dispatch
-# without spending a reviewer approval. `publish` needs validate and carries
-# the `dockerhub` environment gate, because it uses the same credential that
-# can push images. Deliberately NOT wired into promote-image: the Overview is
+# Runs on two triggers. A push to main that changes an Overview file queues a
+# publish for exactly the image(s) whose file changed, so landing an Overview
+# edit is sufficient — nobody has to remember a follow-up dispatch (the first
+# Overview sat unpublished for two days because publishing was a separate
+# manual step). Manual dispatch remains for re-publishing without a content
+# change (e.g. after Docker Hub support intervention). Both paths still pause
+# on the `dockerhub` environment gate before anything touches the credential.
+#
+# Deliberately NOT wired into promote-image or releases: the Overview is
 # version-neutral, so a wording fix should not require a release, and a
 # release should not silently rewrite it.
 #
-# One image per dispatch rather than a matrix over both, because a Docker Hub
-# repository that does not exist yet must fail loudly for that image alone
-# instead of failing a run that also had work to do for the other.
+# validate/publish run as a fail-fast:false matrix over the planned image
+# list, so a Docker Hub repository that does not exist yet fails loudly for
+# that image alone instead of failing work the run also had to do for the
+# other. `validate` is credential-free and ungated, so a dead link, an
+# oversize file, or a missing Docker Hub repository is discovered before a
+# reviewer approval is spent. `publish` carries the `dockerhub` environment
+# gate, because it uses the same credential that can push images.
 
 name: Docker Hub description
 
 on:
+  push:
+    branches: [main]
+    paths:
+      - 'docker/dockerhub-overview.md'
+      - 'docker/dockerhub-overview-dev.md'
   workflow_dispatch:
     inputs:
       image:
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-description.yml
```diff
@@ -7,27 +7,40 @@
...
 # version-neutral, so a wording fix should not require a release, and a
 # release should not silently rewrite it.
 #
-# One image per dispatch rather than a matrix over both, because a Docker Hub
-# repository that does not exist yet must fail loudly for that image alone
-# instead of failing a run that also had work to do for the other.
+# validate/publish run as a fail-fast:false matrix over the planned image
+# list, so a Docker Hub repository that does not exist yet fails loudly for
+# that image alone instead of failing work the run also had to do for the
+# other. `validate` is credential-free and ungated, so a dead link, an
```
robinnsc, line 31: This claim only holds inside `validate`'s own reporting. `publish` needs the whole `validate` job, and a `needs` on a matrix aggregates across legs, so if one image's validate leg fails, every publish leg is skipped. On a both image push where one Hub repo does not exist yet, which is exactly the scenario this property was written for, neither image publishes. Under the old one image per dispatch shape, the other image was a separate run and unaffected.

I would not restructure for it. Matrix legs cannot have per leg needs, and recovery is easy (create the repo, rerun failed jobs, the surviving validate leg passes and publish proceeds for both). But the comment should describe the real behavior: `fail-fast: false` gets you a full failure report across images, not independent publishes.

.github/workflows/dockerhub-description.yml
```diff
@@ -41,33 +54,99 @@ on:
...
         id: src
         env:
-          IMAGE: ${{ inputs.image }}
+          IMAGE: ${{ matrix.image }}
         run: |
           set -euo pipefail
           # Docker Hub caps the short description at 100 characters; asserted below.
+          # This mapping is duplicated in `publish` below: matrix job outputs
+          # are last-writer-wins across the matrix, so each job derives its
+          # own rather than passing outputs between jobs.
```
robinnsc, line 149: The last writer wins point is true for matrix job outputs, but `plan` is not a matrix job. It could emit an `include` array (`[{image, file, short}, ...]`) and both matrices would read `matrix.file` and `matrix.short` directly, no duplication needed.

That buys more than tidiness. Right now this job asserts the 100 character cap against its copy of `SHORT` while `publish` sends its own copy, and nothing checks the two agree, so someone editing the short description in one place ships a string that was never validated. Deriving both from `plan` makes the validated string and the published string the same string by construction.

.github/workflows/dockerhub-description.yml
```diff
@@ -41,33 +54,99 @@ on:
...
+          if [[ "$EVENT" == "workflow_dispatch" ]]; then
+            case "$INPUT_IMAGE" in
+              extenddb-postgres|extenddb-dev) ;;
+              *) echo "::error::unknown image '$INPUT_IMAGE'"; exit 1 ;;
+            esac
+            IMAGES=$(jq -cn --arg i "$INPUT_IMAGE" '[$i]')
+          else
+            CHANGED=""
+            if [[ -n "$BEFORE" && "$BEFORE" != "0000000000000000000000000000000000000000" ]]; then
+              CHANGED=$(curl -sS --fail-with-body \
```
robinnsc, line 100: One edge worth handling: the compare API returns at most 300 files. A large push to main that includes an Overview edit past that cap produces a `CHANGED` list without the file, which lands in the loud error branch below with a message saying no Overview file changed when one did. The failure is loud and dispatch exists as a fallback, so nothing is silently lost, but the message would point whoever debugs it in the wrong direction.

Cheap fix: if the response contains exactly 300 files, treat the range as unresolvable and take the publish both path, same as a forced push.

robinnsc (approved):
The design is right: the push trigger removes the remembering while the `dockerhub` environment gate still reviews every publish, the dispatch input is validated against an allowlist, event SHAs only reach the shell through env vars, and biasing unresolvable ranges toward publish-both is the correct call for an idempotent PATCH. The `grep && assignment` short circuits are genuinely safe under `set -e` (non final AND list commands are exempt), and unifying the concurrency group is correct now that one push can carry both images.

Approving with three findings inline. The first is a claim in the comments that does not hold end to end, the other two are small design improvements. None blocking.

PR #304, opened 2026-08-24
fix(vector): align the vector index lifecycle and rejection messages with DynamoDB

LeeroyHannigan:
## What

Seven fixes to the vector index surface, each qualified against real DynamoDB in **both us-east-1 and eu-west-2** (2026-08-24) before being written. Every behavior and message below is what the live service does; nothing here is inferred.

### Lifecycle

- **An index created with its table now reports `CREATING` while the table is `CREATING`, and flips `ACTIVE` atomically with the table.** Measured on the service (three runs polling at 250 ms): both reach `ACTIVE` in the same DescribeTable poll, with no gap in either direction. Previously the index claimed `ACTIVE` from the first instant, so a client following the documented wait-for-the-index readiness loop was released onto a still-CREATING table and got `ResourceNotFoundException` from its first write or search. This was the root cause behind most of the observed divergences, including four write-rejection messages that were already correct but unreachable.
- **UpdateTable-created indexes hold `CREATING` for a configurable floor** (`vector_index_min_creating_ms`, default 1000; zero disables) before the `ACTIVE` flip. The service's online-index machinery never completes an added index instantly (~17 min observed for a 25-item table), so the documented walk (`CREATING`/`Backfilling: false`, then `true`, then `ACTIVE` with the member absent) is always observable there; a local backfill over a small table finished in milliseconds and collapsed that walk into an unobservable instant. Searches keep answering the documented not-ready rejection and writes keep queuing behind the CREATING hold for the duration, exactly as during a real backfill.
- **Startup reconciliation and the stuck-build sweep are scoped to `backfilling IS NOT NULL`**, so a with-table CREATING index (which has no build task) is never rebuilt early or flipped ACTIVE ahead of its own table.
- **DeleteTable under an in-flight index build is refused** with `ResourceInUseException` and the documented sentence: `Cannot delete table while indexes are being created, updated, or deleted.` Cancelling the build with a `Delete` action remains allowed, as on the service.

### Validation and messages

- `Dimensions` below 1 rejects at the **request-model layer** (`1 validation error detected: Value '0' at 'vectorIndexes.1.member.dimensions' failed to satisfy constraint: Member must have value greater than or equal to 1`), captured live with client-side SDK validation disabled. The semantic `between 1 and 4096 inclusive` message now covers the upper bound only, which is the split the service makes.
- Two indexes on one vector attribute with different `Dimensions` reject with the service's `Conflicting attribute definition for '<attr>'...` message (previously accepted).
- More than one action in `VectorIndexUpdates` rejects with `LimitExceededException`: `Subscriber limit exceeded: Only 1 online index can be created or deleted simultaneously per table`.
- An empty string or binary value in a SearchSchema HASH attribute rejects with the secondary-index empty-key sentence, `IndexName`/`IndexKey` suffix included (previously accepted silently).
- Two SearchVectors messages completed to their full wire text: the invalid-values rejection now names the 32-bit float constraint, and the comparator rejection carries its `Invalid SearchConditionExpression: ` prefix.

## Verification

- `cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, workspace tests green (623 tests, one adapted to the new one-online-action rule).
- End-to-end against the dev server in `:memory:` mode: full external verification pass, 12 previously-diverging behaviors now match the service, zero regressions.
- SQLite backend only: Postgres does not implement vector indexes.

### DIFF
--- a/crates/storage-sqlite/src/store.rs
+++ b/crates/storage-sqlite/src/store.rs
@@ -295,6 +295,30 @@ impl SqliteEngine {
         }
     }
 
+    /// Minimum milliseconds an UpdateTable-created vector index stays `CREATING`
+    /// before its `ACTIVE` flip. See
+    /// [`extenddb_core::settings_keys::VECTOR_INDEX_MIN_CREATING_MS`] for why the
+    /// hold exists. Defaults to 1000 when unset or unparseable; zero disables it.
+    pub(crate) async fn vector_index_min_creating_ms(&self) -> u64 {
+        const DEFAULT_MS: u64 = 1_000;
+        let live: Result<Option<(String,)>, _> =
+            sqlx::query_as("SELECT value FROM settings WHERE key = ?")
+                .bind(extenddb_core::settings_keys::VECTOR_INDEX_MIN_CREATING_MS)
+                .fetch_optional(&self.pool)
+                .await;
+        match live {
+            Ok(row) => row
+                .and_then(|(v,)| v.parse::<u64>().ok())
+                .unwrap_or(DEFAULT_MS),
+            Err(e) => {
+                tracing::debug!(
+                    "vector_index_min_creating_ms: live read failed, using {DEFAULT_MS}: {e:?}"
+                );
+                DEFAULT_MS
+            }
+        }
+    }
+
     /// Handle to the GSI propagation notifier, woken after an enqueue.
     pub(crate) fn gsi_notify(&self) -> Arc<tokio::sync::Notify> {
         Arc::clone(&self.gsi_notify)
--- a/crates/core/src/expression/search_condition.rs
+++ b/crates/core/src/expression/search_condition.rs
@@ -76,7 +76,8 @@ pub fn validate_search_condition_expression(
     // Only equality is supported. Every other comparator contains one of these.
     if trimmed.contains(['<', '>', '!']) {
         return Err(invalid(
-            "Invalid comparator used in SearchConditionExpression",
+            "Invalid SearchConditionExpression: Invalid comparator used in \
+             SearchConditionExpression",
         ));
     }
 
@@ -430,7 +431,7 @@ mod tests {
                 Some(&n),
                 Some(&v)
             )),
-            "Invalid comparator used in SearchConditionExpression"
+            "Invalid SearchConditionExpression: Invalid comparator used in SearchConditionExpression"
         );
     }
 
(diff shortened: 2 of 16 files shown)


crates/storage-sqlite/src/store.rs
```diff
@@ -295,6 +295,30 @@ impl SqliteEngine {
         }
     }
 
+    /// Minimum milliseconds an UpdateTable-created vector index stays `CREATING`
+    /// before its `ACTIVE` flip. See
+    /// [`extenddb_core::settings_keys::VECTOR_INDEX_MIN_CREATING_MS`] for why the
+    /// hold exists. Defaults to 1000 when unset or unparseable; zero disables it.
+    pub(crate) async fn vector_index_min_creating_ms(&self) -> u64 {
```
robinnsc, line 302: The key this reads is not registered anywhere a caller can set it. `KNOWN_KEYS` in `ops_settings.rs` carries `VECTOR_BACKFILL_BATCH_DELAY_MS` but not `VECTOR_INDEX_MIN_CREATING_MS`, and `cmd_settings.rs` validates against that list and rejects unknown keys. So `extenddb settings set vector_index_min_creating_ms 0` fails with "Unknown setting", the management API refuses it too, and the only way to configure the floor today is a manual row insert into the settings table.

That makes the documented "zero disables" escape hatch unreachable, and it is the one people will actually want: with the default floor, every UpdateTable index creation holds that table's queued index writes for about a second, which test suites and CI will notice. Given the sibling key from the same subsystem is registered, this looks like an oversight rather than a choice. One entry in `KNOWN_KEYS` (plus the console settings page if it lists keys) closes it.

robinnsc (approved):
This one lands well. It closes the backfill collision I flagged on #244 exactly the right way: `VectorRowConflict::KeepExisting` for the backfill caller while the write path keeps its loud plain INSERT, with a regression test that reproduces the wedged CREATING scenario and a comment explaining why an existing row is always the same or newer generation. The lifecycle work is coherent too: the `backfilling IS NULL` versus `NOT NULL` split cleanly separates with table indexes from UpdateTable builds, the reconciler and stuck build sweep are scoped to match, the with table index flip rides the same transaction as the table flip and is written as an idempotent self healing catch all, and the DeleteTable refusal is checked inside `BEGIN IMMEDIATE` in both branches so a build committing while the call waits on the write lock is still caught.

Approving with one real finding inline (the new settings key is not registered, so the documented knob cannot actually be set), plus two notes:

1. The one second default hold is a behavior change beyond parity: every UpdateTable index creation now queues that table's index writes behind the CREATING hold for the floor's duration. The mechanics are sound, and the build registry correctly keeps the stuck build sweep from firing during the sleep since the task stays alive and registered. But it is one more reason the knob needs to be reachable.
2. The empty value rejection covers the HASH position only. If the service also refuses empty strings in INLINE_FILTER attributes, that is unmeasured territory; a one line scope note on the match arm would keep the next reader from assuming the coverage is complete.

Also appreciated the honest annotation on the swap test flip: saying plainly that the old must-pass reasoning was never measured is what keeps this suite trustworthy.

PR #305, opened 2026-08-24
fix: align capacity accounting, restore metadata, and rejection behavior with DynamoDB

LeeroyHannigan:
## What

Eighteen behaviors qualified against real DynamoDB in **both us-east-1 and eu-west-2** (2026-08-24): each one diverges here and matches the live service in both regions, so none is a regional split. Companion to #304 (vector surface); the two are independent and both branch from main.

### ConsumedCapacity

- **Query on a secondary index with `INDEXES`** reports the breakdown: the index carries the read, the table arm is zero, aggregate = sum (previously the table arm alone carried everything and the index map was absent).
- **An identical PutItem overwrite charges no index writes**: the table arm alone. An earlier measurement had put replacements charging the index regardless of change; re-measured, the service skips unchanged projected entries.
- **BatchWriteItem with `INDEXES`** reports each per-table entry with the arms the batch actually touched (an indexed put and a sparse put answer total 3 = table 2 + gsi 1). The old images needed for attribution are fetched only when `INDEXES` is requested, so the aggregate-only shapes pay nothing new.
- **A same-token TransactWriteItems replay** reports read capacity recomputed against the item size (`2 x ceil(size/4KB)`), never the stored write magnitude relabelled: a ~1.5KB put answers 4 WCU on the first call and 2 RCU on the replay, which only agree below 1KB.

### RestoreTableFromBackup

- The response's `TableDescription` carries `RestoreSummary` with `SourceBackupArn` and `RestoreInProgress: true`.

### Validation and messages

- Empty-binary secondary-index key values say **"binary"**, not "string", in both the put-form (with `IndexName`/`IndexKey` suffix) and update-expression-form messages.
- The **nesting-depth limit applies to every ExpressionAttributeValue** on UpdateItem, condition-only placeholders included: a 32-level value answers the nesting `ValidationException` instead of evaluating the condition. A 31-level value still passes. (An earlier measurement had condition-only values exempt; both regions now apply the limit.)
- **BatchGetItem with empty `RequestItems`** answers the request model's length constraint (`Value at 'RequestItems' ... length greater than or equal to 1`). BatchWriteItem keeps its distinct required-parameter sentence, which is likewise measured.
- **Duplicate BS members** answer the binary form naming the set type, with the service's own missing space (`]of type BS contains duplicates`).
- **Select SPECIFIC_ATTRIBUTES without a projection** carries the `1 validation error detected: ` envelope on Query and not on Scan, the same per-operation split the sibling Select rejection already implements.
- **Invalid enum aggregation matches each operation**: PutItem aggregates three (`ReturnConsumedCapacity`, `ReturnItemCollectionMetrics`, `ReturnValues`), DeleteItem aggregates two, UpdateItem stops at the first invalid enum and reports one, and Query names `returnConsumedCapacity` ahead of `select`.

## Verification

- `cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, workspace tests green.
- Full external verification pass: all eighteen behaviors now match the service, zero regressions. The only remaining divergences on this branch are the vector items #304 carries.


crates/core/src/types/capacity.rs
```diff
@@ -217,6 +217,44 @@ impl ConsumedCapacity {
...
     }
 
+    /// Build a `ConsumedCapacity` for a read served by a secondary index.
+    ///
+    /// The index carries the read and the table's share is zero, with the
+    /// aggregate being their sum; that is what makes the INDEXES breakdown a
+    /// breakdown rather than a copy of the total (pinned by the ground-truth
+    /// runs of 2026-08-24, us-east-1 and eu-west-2). `is_global` selects the
+    /// map the index entry lands in.
+    pub fn read_on_index(table_name: &str, index_name: &str, cu: f64, is_global: bool) -> Self {
```
robinnsc, line 227: One question about the zero table arm. For GSIs this is unconditionally right, but LSIs have a documented service behavior this builder cannot express: a query that requests attributes beyond the index projection (say `ALL_ATTRIBUTES` against a `KEYS_ONLY` LSI) makes DynamoDB fetch from the base table and charge additional read capacity to the table arm. If the ground truth runs only exercised reads served from the projection, the zero here over generalizes to exactly the case where the service splits the charge.

If that case was measured and the service really does report zero, ignore me and maybe note it in the doc comment. If it was not, worth either measuring it or scoping the comment to projection served reads so the next person does not treat the zero as universal.

crates/engine/src/backup.rs
```diff
@@ -147,12 +147,28 @@ pub(crate) async fn handle_restore_table_from_backup(
...
+    // from and that the restore is under way: SourceBackupArn and
+    // RestoreInProgress: true, pinned by the ground-truth runs of 2026-08-24
+    // (us-east-1 and eu-west-2). Set here rather than in each backend because
+    // the summary is response metadata about this call, not table state the
+    // backends persist.
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_secs_f64())
+        .unwrap_or_default();
+    desc.restore_summary = Some(extenddb_core::types::RestoreSummary {
```
robinnsc, line 166: The scoping decision here is defensible but it creates a client visible divergence worth writing down. On the real service `RestoreSummary` persists: DescribeTable keeps reporting it after the restore, with `RestoreInProgress` flipping to false once it completes. Here the summary exists only on this one response, so a client that reads `RestoreInProgress: true` and then polls DescribeTable waiting for the flip never sees a `RestoreSummary` again. And since the restore is synchronous on this side, the `true` describes work that has already finished.

I would not persist it just for this, the response shape parity is the part that matters for SDK compatibility. But this is a new divergence with a plausible failure mode for real client code, so it belongs in `docs/differences-from-dynamodb.md`, either in this PR or as an immediate follow up.

robinnsc (approved):
Went through all eighteen behaviors. The mechanical quality is high, and the honesty in the test flips deserves a mention: annotating that an earlier live check was contradicted by remeasurement, rather than quietly rewriting the assertion, is exactly what keeps a parity suite trustworthy.

Particular things that hold up well: the per operation enum divergence encoding (aggregate vs stop at first, per operation clause order, and the `EnumClause::Bare` variant reproducing the service's ReturnValues clause with its `UPDATE_NEW` typo while keeping the real accepted member list separate from the displayed one), BatchWriteItem fetching old images only when INDEXES is requested so the aggregate shapes pay no new read, the nesting depth boundary tested from both sides with the TransactWriteItems exemption measured rather than assumed, and the replay capacity builder documenting its own imprecision for the key only sub ops.

Approving with two observations inline, both about places where the measured model may over generalize. Neither is blocking, but the second probably wants a line in `docs/differences-from-dynamodb.md`.

PR #306, opened 2026-08-24
fix: measured vector conformance gaps, and the index-build lifecycle as shared code

yesyayen:
## What

Vector conformance fixes, each pinned to a wire capture from Amazon DynamoDB, plus the ADR-0005 lifecycle extraction. Both are prerequisites for a second backend implementing vector search.

- SearchVectors `ConsumedCapacity` reports `VectorSearchUnits` alongside `VectorSearchRequestBytes` (always equal, as the service returns them).
- The SearchVectors invalid-values rejection and the write-path validation family now match the measured whole strings. The write-path family uses two envelope constants because the envelope goes with the error kind (doubled for wrong type / wrong size / out-of-range component, single sentence for an element-level type error), not with nesting level.
- SQLite refuses a vector-index create when the request's net billing mode is PROVISIONED; a combined switch-to-PAY_PER_REQUEST + create in one UpdateTable stays accepted, both as measured.
- `StorageError::ResourceInUse` with its engine mapping (plumbing for the measured allocation-phase delete refusal, tracked as F-18), and the missing `Unsupported` arm on UpdateTable: a backend refusal surfaced as a 500 where CreateTable answers 400.
- ADR-0005: the SQLite index-build lifecycle moves to `extenddb_storage::vector_lifecycle` as shared drivers over one backend trait (backend-owned cursor type, defaulted heartbeat hook). Behavior-preserving: zero test-expectation changes.

Every value below was captured from Amazon DynamoDB via raw signed requests (us-west-2, 2026-08-19).

|---|---|---|
| SearchVectors `ConsumedCapacity` | `{"VectorSearchRequestBytes": 1024.0}` | `{"VectorSearchRequestBytes": 1024.0, "VectorSearchUnits": 1024.0}` |
| SearchVectors query vector with NaN / Infinity / value outside f32 range | `Search vector contains invalid values` | `Search vector contains invalid values. All values in the search vector must be a 32-bit floating-point number attribute` |
| PutItem vector with wrong dimension count | `One or more parameter values were invalid. Invalid size for parameter emb, Expected: 4, Actual: 3. IndexName: vidx` | `One or more parameter values are not valid. One or more parameter values were invalid . Invalid size for parameter emb, Expected: 4, Actual: 3. IndexName: vidx` |
| PutItem vector of wrong type, or with a component outside f32 range | single-sentence envelope, same shape as above | doubled envelope, same shape as above |
| UpdateTable creating a vector index while the net billing mode is PROVISIONED | index created | `ValidationException: One or more parameter values were invalid: Vector indexes are only supported for PAY_PER_REQUEST tables` |
| UpdateTable switching to PAY_PER_REQUEST and creating the index in one request (control) | accepted | accepted |

## Why

Conformance gaps found by comparing ExtendDB against Amazon DynamoDB. The previous strings were pinned by tests asserting the wrong text.

Closes # n/a

## Testing done

Captured every expected string and response shape from Amazon DynamoDB first (raw wire log, 83 request/response pairs), then validated dual-target.

- New Rust wire tests in `tests/rust` covering each gap, including the per-item TransactWriteItems `CancellationReasons` shape for an invalid vector.
- Extraction gate: the full wire suite ran green before and after the lifecycle move (488 tests, all 50 vector tests executed under `EXTENDDB_EXPECT_VECTORS=1`). The wire run exercises propagation delay 0; the delay>0 queue path is covered by 13 unit tests.
- `cargo test --workspace`, `cargo fmt --check`, and `cargo clippy --all-targets -- -D warnings` all pass.

## Checklist

ADR / RFC: docs/adr/0005-index-build-lifecycle-ownership.md (marked Accepted with an outcome section in this PR)

## Breaking changes

n/a

---


crates/core/src/validation/vector_item.rs
```diff
@@ -26,6 +26,29 @@ fn invalid(msg: impl Into<String>) -> DynamoDbError {
...
+/// type, a component out of f32 range, and the per-item TransactWriteItems
+/// cancellation reason, which repeats the PutItem wording verbatim). Both
+/// oddities are the service's own: the sentence is stated twice, in two
+/// different tenses, and there is a space before the second full stop.
+///
+/// Not universal. Probe P13 measured the element-level TYPE error using
+/// [`INVALID_PARAMETER_VALUES_ONCE`] instead, so the envelope goes with the error
+/// KIND rather than with this file. Reading the four captures above as one rule
+/// is what got it wrong once already: they are all doubled-envelope kinds.
+const INVALID_PARAMETER_VALUES_TWICE: &str =
```
LeeroyHannigan: **Blocking** - the doubled-vs-single split isn't per error kind, it's per region. us-east-1 and us-west-2 give the single envelope for every kind; eu-west-1 gives the doubled one for every kind. So these byte-exact tests match no single region. Pick one region's shape and use it for all four kinds.

crates/storage/src/error.rs
```diff
@@ -63,6 +63,13 @@ pub enum StorageError {
     /// never claimed, so it is not a bug and must not be logged as one.
     #[error("Not supported by this storage backend: {0}")]
     Unsupported(String),
+    /// The request targets a resource whose current state forbids the change.
+    /// Maps to `ResourceInUseException`, which the service uses for a vector
+    /// index deleted while its creation is still allocating resources (measured
+    /// 2026-08-19). Carries the whole message because the state, and therefore
+    /// the wording, is known only to the backend that holds it.
+    #[error("{0}")]
+    ResourceInUse(String),
```
LeeroyHannigan: **Should-fix** - this overlaps `IndexesInUse` already on main (both map to ResourceInUseException) and has no producer yet. Worth folding into `IndexesInUse` on the rebase rather than carrying two.

crates/storage-sqlite/src/data/vector_index.rs
```diff
@@ -576,134 +468,176 @@ async fn backfill_vector_batch(
...
+        let outcome = backfill_vector_batch(&mut tx, &plan, limit, cursor).await?;
+        tx.commit()
+            .await
+            .map_err(crate::sqlite_util::map_sqlx_err)?;
+        Ok(outcome)
+    }
+
+    async fn set_backfilling(&mut self) -> Result<(), StorageError> {
+        sqlx::query(
+            "UPDATE vector_indexes SET backfilling = 1 WHERE table_id = ? AND index_id = ?",
```
LeeroyHannigan, line 595: **Blocking** - this is the pre-fix backfill. Rebasing needs to keep main's KeepExisting/`INSERT OR IGNORE` path and the min-CREATING window - both are dropped as the branch stands, so a write landing mid-backfill can wedge the build again.

LeeroyHannigan (commented):
Nice work, the shared lifecycle module and pinning strings to real captures is the right call. Two things holding it up, both from the branch being cut before the vector fixes landed on main. Rebase onto current main and re-apply the extraction on top, and I think it's most of the way there.

crates/core/src/validation/vector_item.rs
yesyayen: good catch - re-probed all 3 regions and you're right, it's per region not per kind. us-east-1/us-west-2 single, eu-west-1 doubled.

pinned everything to the us shape in e5af651 commit. Two message bodies also changed since our original capture (type error is prose now, size error dropped a period) so those got pinned too.

crates/storage-sqlite/src/data/vector_index.rs
yesyayen: yes, the extraction was based on a pre-#304 snapshot so both would've been dropped.

Fixed it during the rebase. The KeepExisting/INSERT OR IGNORE path is back in the backfill apply hook, and the min-CREATING window moved into the shared complete_build (new earliest_active param, repair paths skip it).

crates/storage/src/error.rs
yesyayen: Yep, agreed - both mapped to the same exception and ours had no producer left after the rebase. Folded it into IndexesInUse. The variant now carries the full message composed by the backend, and the engine just maps it to ResourceInUseException as-is. The delete-table sentence moved into its producer unchanged so nothing moved on the wire.

Unit tests cover the passthrough and pin the allocation-phase string to the probe capture, and the wire suite exercises both refusals end to end.

LeeroyHannigan (approved):
🚀

PR #307, opened 2026-08-24
feat(postgres): vector indexes and SearchVectors

yesyayen:
## What

PostgreSQL implements the vector index surface end to end: CreateTable / UpdateTable create and delete with a real backfill, DescribeTable status reporting, SearchVectors exact scan (pgvector, one `vector(N)` data table per index), index maintenance at all six write sites through the shared `gsi_pending` queue, crash recovery, and cross-process build ownership (advisory lock plus heartbeat; the lock holds its own session because a pooled connection's session outlives its checkout, which would otherwise defeat the stuck-build sweep).

pgvector is runtime-detected: `init` attempts `CREATE EXTENSION` and tolerates failure; `serve` probes once and caches. Without the extension every vector operation fails closed at validation with a typed error, proven over the wire by a CI job on a plain postgres image. The catalog migrates 0.0.2 -> 0.0.3, and the version gate refuses symmetrically in both directions.

Because the shared wire suite is backend-blind, SQLite gains the same measured rules in this PR: the allocation-phase delete refusal, the observable CREATING with `Backfilling: false` window, and f64 score arithmetic. Zero vectors index and score 1.0 as measured. Ties break on the base key; the service's tie order is measured-unstable, so any order conforms.

One deliberate residual is documented in `docs/differences-from-dynamodb.md`: pgvector's internal f32 accumulation collapses cosine distance to three values for query vectors whose squares underflow; SQLite computes the true value. Known follow-ups are tracked in `docs/technical-debt.md` (F-19: SQLite restore silently drops vector indexes where PostgreSQL refuses; F-20: SQLite's zero-delay inline maintenance bypasses the backfill hold).

Every value below was captured from Amazon DynamoDB via raw signed requests (us-west-2, 2026-08-19).

|---|---|---|
| SearchVectors on the PostgreSQL backend | not supported (backend refusal) | supported |
| UpdateTable deleting a vector index still in the resource-allocation phase | index deleted, 200 | `ResourceInUseException: Attempt to change a resource which is still in use: Index creation is in resource allocation phase. Retry deletion during backfilling phase or when the index is active. Table: {table} Index: {index}` |
| UpdateTable deleting a vector index during backfilling, or when ACTIVE (control) | accepted | accepted |
| SearchVectors where the score arithmetic under- or overflows f32 | `"Score": null` on HTTP 200 | finite 32-bit score |
| SearchVectors with a zero query vector under COSINE | score exactly `1.0` (SQLite; PostgreSQL n/a) | score exactly `1.0`, zero on either side |

## Why

The second storage backend for vector search, built on the lifecycle primitives extracted in #306 (ADR-0005): the backend implements a cursor and apply hooks; the shared drivers own backfill, status sequencing, recovery, and holds. Storage and scoring decisions (vector(N) layout, operator-derived scores, f64 norms, fail-closed detection) are recorded in ADR-0006.

Stacked on #306: the first 19 commits are that PR; review the last 9 here.

Closes # n/a

## Testing done

Measured against the live service throughout: 91 recorded wire pairs, including the phase-dependent delete, net-state billing, sparse-index semantics, and the dimension-mismatch message family.

- Wire suite: 496/496 on PostgreSQL and 496/496 on SQLite, both at `EXTENDDB_EXPECT_VECTORS=1` (58 vector tests). The PostgreSQL leg was run against PostgreSQL 15.18 with pgvector 0.8.0; CI runs the same suite on the `pgvector/pgvector:pg16` image, and a separate job runs the fail-closed refusal leg on a plain `postgres:16` image (the new CI job shape).
- Storage-level suite with mutation-verified failing-first coverage; the `"Score": null` fix was verified by watching the test fail against the unfixed build.
- The catalog migration exercised on a live deployment with both binary versions; refusal is symmetric.
- `cargo test --workspace`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets -- -D warnings` pass at the final tip.

## Checklist

ADR / RFC: docs/adr/0006-pgvector-storage-and-scoring.md (lifecycle ownership: docs/adr/0005-index-build-lifecycle-ownership.md, via #306; RFC context: #236)

## Breaking changes

n/a

---


crates/storage-postgres/src/data/vector_index.rs
```diff
@@ -0,0 +1,1105 @@
...
+    //
+    // Deliberately an over-approximation. A GSI row pending under a per-index delay
+    // while the global delay is zero also forces this write to queue. The queued
+    // row is ready immediately and applies in order, so the cost is eventualness,
+    // never a wrong index row. Inline only exists at delay zero, where the queue
+    // holds build-era rows and drains to empty, so the inline path resumes; there
+    // is no steady-state traffic that could strand it.
+    let queue_empty = if delay_ms == 0 && metas.iter().any(|(_, status)| status == "ACTIVE") {
+        let pending: Option<(i32,)> =
+            sqlx::query_as("SELECT 1 FROM gsi_pending WHERE table_id = $1 LIMIT 1")
```
LeeroyHannigan, line 187: **Should-fix** - maintenance runs through the gsi_pending queue, so a search right after a write can come back empty until it drains. Worth a line in the docs so nobody reads that as data loss.

LeeroyHannigan (commented):
This is solid - control plane, backfill, SearchVectors, the advisory-lock ownership all hold up, and I ran it against pgvector end to end. Only real issue is it's stacked on #306, so it picks up #306's blockers; rebase once that's sorted. One small doc ask inline.

crates/storage-postgres/src/data/vector_index.rs
yesyayen: Agreed, worth documenting. It's the same eventual-consistency model as a GSI, so the write is never lost, just queued. Added a note to the vector propagation row in `differences-from-dynamodb.md` saying exactly that: the window is ordering, not data loss. Done in commit 941821e.

LeeroyHannigan (approved):
LGTM 🚀

LeeroyHannigan (approved):

PR #309, opened 2026-08-25
fix(mongodb): clean up account, index, and tag lifecycle data

diegotoledano95:
## What

  Fix MongoDB lifecycle cleanup for accounts, secondary indexes, and table tags.

  Changes:

  - Read account ownership from `_id.account_id` during account deletion checks.
  - Drop physical `_ddb_<index_id>` collections when indexes are deleted.
  - Handle already-missing physical collections safely.
  - Remove table tags when tables are deleted.
  - Preserve safe catalog/cache/collection cleanup ordering.

  ## Why

  Account deletion could bypass the table dependency check because the account ID
  was stored inside `_id.account_id`, not at the queried top-level path.

  Physical index collections and table tags could also survive table deletion.
  Recreating a table with the same account and name could expose stale metadata.

  ## Testing done

  - MongoDB unit tests.
  - Account deletion dependency integration test.
  - DynamoDB API test that verifies physical index collections are deleted.
  - Table deletion and recreation test verifying tags do not survive.

  ## Checklist

  ## Breaking changes

  None.


crates/storage-mongodb/src/table_engine.rs
```diff
@@ -572,10 +572,19 @@ impl MongoEngine {
             .await
             .map_err(|e| StorageError::Internal(e.to_string()))?;
 
-        // Delete index entries
+        // Drop each physical GSI/LSI collection before deleting its catalog
+        // metadata. The index documents are stored separately from the table
+        // document, so dropping the base collection alone does not remove
+        // `_ddb_<index_id>` collections.
+        self.drop_index_collections_for_table(&desc.table_id)
```
LeeroyHannigan, line 589: **Nit**: DeleteTable drops physical collections before deleting catalog docs, while DeleteIndex deliberately goes catalog-first for reader safety (the comment around line 950 explains why). Safe here since the table is already DELETING, but a one-line note would stop a future reader assuming the DeleteIndex invariant holds on this path too.

LeeroyHannigan (commented):
All three gaps check out as real, the account-delete guard on main filters on a field that does not exist in the catalog docs, so it never fired. Ran the physical-cleanup and isolation suites live and everything this PR touches passes, including the mongosh assertions that the index collections are actually gone. Heads up on one unrelated thing I hit while running your tests: the two TestConsoleIsolation tests in this file fail under our harness because the pre-existing `_console_login` helper sets `session.verify = False` without `trust_env = False`, so the exported CA bundle wins. Not your bug and not this PR's code, we will fix it separately.

LeeroyHannigan (dismissed):

crates/storage-mongodb/src/table_engine.rs
```diff
@@ -917,17 +929,37 @@ impl MongoEngine {
...
+                        .await
+                        .map_err(|e| StorageError::Internal(e.to_string()))?;
 
                     // Invalidate cache — may still have other GSIs
                     self.gsi_cache_invalidate(&desc.table_id);
+
+                    // The physical cleanup is still reported to the caller if
+                    // it fails, but readers no longer observe the deleted
+                    // index as available while that cleanup is in progress.
+                    self.drop_index_collection(&index_id).await?;
```
LeeroyHannigan, line 963: **Blocking** Deleting a GSI that's still backfilling races the worker: it never re-checks the catalog between batches, and its upsert implicitly recreates the collection we just dropped here, so we end up with a fully populated orphan no cleanup path can ever see. Cheapest fix is in the worker: treat matched_count == 0 on the per-batch cursor update as "index deleted", abort the job, and re-drop.

LeeroyHannigan (commented):
Had another pass over this side by side with #310 since they share the backfill worker, and I spotted a couple of things I missed first time round. Nothing structural, the cleanup logic itself holds up well.

crates/storage-mongodb/src/table_engine.rs
```diff
@@ -573,10 +573,22 @@ impl MongoEngine {
...
 
-        // Delete index entries
+        // The table is already DELETING, so data-plane readers are rejected;
+        // unlike individual index deletion, physical cleanup can safely occur
+        // before deleting the catalog metadata.
+        // Drop each physical GSI/LSI collection before deleting its catalog
+        // metadata. The index documents are stored separately from the table
+        // document, so dropping the base collection alone does not remove
+        // `_ddb_<index_id>` collections.
+        self.drop_index_collections_for_table(&desc.table_id)
```
LeeroyHannigan, line 589: **Blocking** The base-collection drop just above isn't NamespaceNotFound-tolerant like your new drop_index_collection, so if any of these later cleanup steps fails transiently, the table wedges in DELETING and every retry dies on code 26 at the base drop. The same code-26 arm you added for index collections fixes it.

crates/storage-mongodb/src/table_engine.rs
```diff
@@ -917,17 +929,37 @@ impl MongoEngine {
                     let desc = self
                         .describe_table_impl(account_id, &input.table_name)
                         .await?;
-                    let result = self.catalog_db.collection::<Document>("indexes")
-                        .delete_one(doc! { "_id": { "table_id": &desc.table_id, "index_name": &delete.index_name } })
+                    let indexes_coll = self.catalog_db.collection::<Document>("indexes");
+                    let index_filter = doc! { "_id": { "table_id": &desc.table_id, "index_name": &delete.index_name } };
+                    let index_doc = indexes_coll
+                        .find_one(index_filter.clone())
```
LeeroyHannigan: **Should-fix** Two concurrent DeleteIndex calls both report success now that the delete result is ignored. find_one_and_delete gives you the doc for index_id extraction and the loser gets IndexNotFound, one fewer round trip too.

crates/storage-mongodb/src/ttl_worker.rs
```diff
@@ -167,13 +175,20 @@ async fn run_gsi_backfill_job(storage: &MongoEngine, job: &Document) -> Result<(
...
+                    doc! { "index_id": &index_id, "index_status": "CREATING" },
                     doc! { "$set": { "backfill_cursor": last_id.clone() } },
                 )
                 .await
                 .map_err(|e| StorageError::Internal(e.to_string()))?;
+            if result.matched_count == 0 {
+                tracing::info!(
+                    "GSI backfill worker: index_id={index_id} was deleted during backfill; dropping orphan collection"
+                );
+                storage.drop_index_collection(&index_id).await?;
```
LeeroyHannigan: **Blocking:** with two instances sharing a catalog (the documented mode in lib.rs), the worker that loses the backfill race sees matched_count == 0 here after the winner flips the index ACTIVE, concludes the index was deleted, and drops a live index's collection; queries then silently return empty. On zero match, re-check the row with find_one on index_id alone and only drop when the row is truly gone.

LeeroyHannigan (commented):
Re-checked the new commits. All three asks are fixed exactly as requested and the tests are the real thing, thank you.

**Blocking (branch structure):** the rebase stacked this branch on #310's full commit stack plus replayed copies of nine already-merged main PRs (27 commits total), so #309 is now a strict superset of #310: squash-merging it lands all of #310's content in one commit under this PR's title and leaves #310 empty. Please tell me how you want to sequence this against #310, and ideally trim the branch to just its own commits on current main.

One inline comment on the new worker abort path. Could you also reply on the open threads so I can resolve them?

crates/storage-mongodb/src/table_engine.rs
diegotoledano95: Addressed. Added a comment explaining that DeleteTable first marks the table DELETING, which rejects data-plane readers. Physical cleanup can therefore safely occur before catalog deletion, unlike individual index deletion, which must remove catalog visibility first.

crates/storage-mongodb/src/table_engine.rs
diegotoledano95: Addressed. DeleteTable now uses the same missing-collection-tolerant cleanup path for the base collection as it does for index collections. A MongoDB code 26 error no longer wedges the table in DELETING during a retry.

crates/storage-mongodb/src/table_engine.rs
diegotoledano95: Addressed. DeleteIndex now uses find_one_and_delete to atomically claim the catalog row. The winning request receives the index identifier and performs physical cleanup; the losing request receives IndexNotFound and cannot report a second success.

crates/storage-mongodb/src/table_engine.rs
diegotoledano95: Addressed. The worker now treats a zero-match cursor update as a lost backfill claim, stops processing that job, and rechecks the catalog before cleaning up the physical collection. This prevents a deleted index from being left behind after its collection is recreated by a backfill.

crates/storage-mongodb/src/ttl_worker.rs
diegotoledano95: Addressed. Both zero-match worker paths now recheck the catalog with find_one using only index_id. The physical collection is removed only when the catalog row is genuinely gone. If another worker has already changed the row to ACTIVE, the collection is retained. The new regression test covers this status race, while the existing deletion test covers genuine catalog deletion.

diegotoledano95:
@LeeroyHannigan Agreed. #310 must merge first. Because #309 currently carries #310’s shared foundation and replayed main-branch commits, I will not merge it as-is. After #310 lands, I will rebase and trim #309 onto the updated main, retaining only the lifecycle-cleanup commits and this race fix. I will verify the final commit list and diff before pushing the cleaned branch.

LeeroyHannigan (approved):

PR #310, opened 2026-08-25
fix(mongodb): mongodb gsi backfill race safe

diegotoledano95:
## What

Make live MongoDB GSI backfills safe against concurrent updates and deletes.

  Changes:

  - Read bounded batches with majority read concern.
  - Process each live item in its own short transaction.
  - Conditionally claim the exact base document before writing its GSI entry.
  - Upsert the GSI entry only after the claim succeeds.
  - Remove the temporary guard before committing.
  - Retry transient transaction conflicts with bounded backoff.
  - Keep restore backfills on a separate non-transactional path.
  - Compile deterministic backfill test hooks only with
    `mongodb-test-hooks`.
  - Verify the required test-hook capability before running the race test.

  ## Why

  A backfill could read an item, then a concurrent delete or update could remove
  its GSI entry. The old backfill could later recreate that stale entry.

  The new live path verifies that the base document still matches the item read by
  the backfill before writing the GSI entry.

  ## Testing done

  - Deterministic integration test that pauses after the backfill read, deletes
    the item through the DynamoDB API, and verifies no stale GSI entry remains.
  - MongoDB builds with and without test hooks.
  - MongoDB integration tests using the `mongodb-test-hooks` build.
  - Package-scoped Clippy checks.

  ## Integration dependency

  This PR must merge before `fix/mongodb-backup-metadata`.

  The backup branch also calls `backfill_gsi_batch`, but currently uses the old
  function signature. After this PR merges:

  1. Rebase `fix/mongodb-backup-metadata` onto the updated `main`.
  2. Update its restore call to construct the new backfill context.
  3. Explicitly use `GsiBackfillMode::Restore`, not `Live`.
  4. Run the combined build and MongoDB integration tests.
  5. Push the rebased backup branch with `--force-with-lease`.

  Using `Live` for restore would compile but incorrectly add guard writes and
  per-item transactions to the restore path.

  ## Checklist

  ## Breaking changes

  None.
  
Push this branch first:

  git push -u fork fix/mongodb-gsi-backfill-race-safe


crates/storage-mongodb/src/data_engine.rs
```diff
@@ -2907,59 +2921,268 @@ impl MongoEngine {
...
+            &projected,
+            context.idx_key_schema,
+            &context.key_info.key_schema,
+            &context.key_info.attribute_definitions,
+        )?;
+        let replace_opts = mongodb::options::ReplaceOptions::builder()
+            .upsert(true)
+            .build();
+
+        if context.mode == GsiBackfillMode::Restore {
```
LeeroyHannigan: **Should-fix**: this `Restore` branch has no caller in this PR, its only consumer arrives with the backup-metadata branch, so it merges unexercised. A unit test driving `Restore` here would cover it until then. Longer term a separate `backfill_gsi_restore` entry point would make it impossible for a rebase to default to `Live` on the restore path.

crates/server/src/management/ops_settings.rs
```diff
@@ -37,6 +37,10 @@ pub const KNOWN_KEYS: &[(&str, Validator)] = &[
         extenddb_core::settings_keys::VECTOR_BACKFILL_BATCH_DELAY_MS,
         validate_backfill_batch_delay_ms,
     ),
+    (
+        extenddb_core::settings_keys::GSI_BACKFILL_TEST_GATE,
```
LeeroyHannigan: **Nit**: this key is writable in a production build but the gate is compiled out, so `settings set` succeeds and silently does nothing. Worth cfg-gating the KNOWN_KEYS entry behind `test-hooks` too, or a doc note that it is inert in prod.

LeeroyHannigan (commented):
Ran this against a live replica set with the deterministic gate and the race fix holds up, the per-item claim transaction is exactly right. Checked the test-hooks gating too, the production build compiles it out cleanly. Two things below, then this is good to go. One note for sequencing: this needs to land before the backup-metadata PR, and that branch will need its backfill call updated to pass `Restore` when it rebases.

crates/storage-mongodb/src/ttl_worker.rs
```diff
@@ -65,10 +66,10 @@ pub(crate) async fn stream_record_cleanup_worker(storage: Arc<MongoEngine>) {
 /// Live writes during the backfill window continue to route through
 /// `sync_indexes` / `sync_indexes_in_session`, which write to
 /// CREATING indexes too (indexes catalog membership, not status, is
-/// what gates the write path). All writes are upserts on the same
-/// `_id` shape, so a base item touched by both the backfill and a
-/// concurrent write converges regardless of interleaving —
-/// RFC-0003 §2.4.
+/// what gates the write path). Both paths use a transaction over the
+/// base and index rows, so a concurrent mutation serializes with or
```
LeeroyHannigan: **Blocking** This overclaims: when gsi_cache_get_fresh returns a stale false, PutItem and DeleteItem take the sessionless fast paths with no index maintenance at all, so a write during that window leaves the ACTIVE index missing an item or holding a phantom. The single-instance version is a TOCTOU where a stale sync's gsi_cache_set(false) lands after UpdateTable's set(true); a generation counter on the cache would close that, and the cross-instance TTL window probably deserves its own issue.

.github/workflows/integration-mongodb.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install Python dependencies
         run: pip install -r requirements.txt
       - name: Build release (mongodb backend)
-        run: cargo build --release --no-default-features --features mongodb
+        run: cargo build --release --no-default-features --features mongodb-test-hooks
```
LeeroyHannigan, line 36: **Blocking** Both jobs now build mongodb-test-hooks only, so nothing in CI compiles the plain mongodb feature graph anymore, including the cfg(not(test-hooks)) arm. A cargo check --release --no-default-features --features mongodb job would cover it cheaply.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -2907,59 +2925,268 @@ impl MongoEngine {
...
         let last_id = docs.last().and_then(|d| d.get("_id").cloned());
 
-        for doc in docs {
-            let item = document_to_item(&doc)?;
-            if !item_has_index_keys(&item, idx_key_schema) {
+        // The test gate is outside any transaction. It lets an API test commit
+        // a DeleteItem after this batch has been read but before the per-item
+        // claim transaction begins, without consuming transaction lifetime.
+        if context.mode == GsiBackfillMode::Live {
+            self.wait_for_gsi_backfill_test_gate().await?;
```
LeeroyHannigan: **Should-fix** The gate is a global singleton, so any table's backfill can claim it under parallel test execution and the race test silently stops exercising the interleaving while still passing. Scoping the gate value by table name keeps it deterministic.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -2907,59 +2925,268 @@ impl MongoEngine {
...
+        // reached the end of the base collection. Upstream flips the
+        // index to ACTIVE when that happens.
+        Ok(progress)
+    }
+
+    /// Backfill one item. Live backfills claim the exact base snapshot and
+    /// write its index row in one short transaction. Restore backfills skip
+    /// the claim because the target table remains unavailable until the copy
+    /// and all index rows are complete.
+    async fn backfill_gsi_item(
```
LeeroyHannigan: **Should-fix** One pre-existing item with a wrong-typed key attribute makes the batch error and retry forever, so the index never leaves CREATING with no signal beyond a warn. Skipping per-item conversion failures with a counted log matches what the real backfill does with non-conforming items. Happy for this to be a follow-up issue if you'd rather keep this PR focused.

LeeroyHannigan (commented):
Came back to this after tracing how the claim mechanism interacts with the GSI cache, which I didn't do on my first pass. The claim transaction itself holds up under every interleaving I could construct on the session paths, nice work on that.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -1735,6 +1757,7 @@ impl MongoEngine {
             return Ok(());
         }
 
+        let cache_generation = self.gsi_cache_generation(&key_info.table_id);
```
LeeroyHannigan: **Blocking:** this capture runs inside the write transaction after the snapshot is established (the base write comes first), so an UpdateTable landing between snapshot and capture bumps the generation before we read it, the guarded publish still passes, and a fresh stale false sits in the cache for 60s, re-opening the fast-path bypass. Capture the generation before the transaction's first operation, or only publish false from a non-transactional catalog read.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -238,6 +238,28 @@ impl DataEngine for MongoEngine {
     }
 }
 
+/// Distinguishes live GSI creation, where concurrent writes are possible,
+/// from restore backfills, where the target table is still unavailable.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum GsiBackfillMode {
+    Live,
+    #[allow(dead_code)]
```
LeeroyHannigan, line 271: **Should-fix:** this silences the warning instead of exercising the variant; a small unit test driving backfill_gsi_item in Restore mode would cover it, or note it explicitly as tracked with #311.

LeeroyHannigan (commented):
Good round: the production-feature CI job, the table-scoped gate, the poison-row handling, and the prod-key validation are all exactly right.

Two inline comments remain, plus the cheapest ask from the original thread: the comment at ttl_worker.rs:66-73 still says both paths serialize with the backfill, which is not true for the fast paths; please update it to state the actual invariant the code relies on. Could you also file the follow-up issue for the cross-instance cache window, and reply on the threads so I can resolve them?

crates/server/src/management/ops_settings.rs
diegotoledano95: Addressed. The test gate is no longer included as a normal production setting. It is accepted only as a table scoped setting when the MongoDB test-hook feature is compiled in, so production builds reject it as an unknown setting instead of silently accepting an inert value.

crates/storage-mongodb/src/ttl_worker.rs
diegotoledano95: Addressed. The cache now uses a generation counter so an older catalog observation cannot publish a stale “no indexes” result after an UpdateTable invalidation. I also created the follow-up issue: #333 for the remaining cross-instance time-to-live window, where another process can retain a negative cache entry until it expires.

.github/workflows/integration-mongodb.yml
diegotoledano95: Addressed. The MongoDB workflow now includes a separate production feature-graph check:

`cargo check --release --no-default-features --features mongodb`

This compiles the normal MongoDB configuration, including the code excluded from test-hook builds.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Addressed. The test gate is now scoped by table name, so parallel backfills on different tables cannot claim the same gate or cause the race test to pass without exercising its intended interleaving.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Addressed. The backfill now skips malformed source items and invalid index-key conversions individually, records the skipped-item count, and continues processing the batch. A single non-conforming item no longer causes the entire backfill to fail and retry indefinitely.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Addressed, the cache generation is now captured before the transaction’s first operation and passed through to the guarded cache publication. If an UpdateTable invalidates the generation while the transaction is in progress, the stale “no indexes” result cannot be published.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Agreed. Restore-mode execution coverage is tracked in PR #311, where the Restore path is exercised by the restore backfill flow. This avoids duplicating the test while the Restore caller is being moved behind the worker there.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Agreed. I added a focused unit test that drives backfill_gsi_item in Restore mode, so this branch is exercised in #310 before #311 is merged. #311 will provide the production Restore caller and end-to-end coverage when it follows immediately after #310. The current explicit Restore mode is therefore covered without duplicating #311’s production implementation.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -2315,6 +2353,7 @@ impl MongoEngine {
                 // stale index entries when this write changes or removes a
                 // GSI key attribute, and (c) supply OldImage to any attached
                 // stream capture.
+                let cache_generation = self.gsi_cache_generation(&key_info.table_id);
```
LeeroyHannigan: **Blocking:** TransactWriteItems still captures the generation here, inside the started transaction (the idempotency pre-check or an earlier operation's read has already established the snapshot), so the stale-false publish we just fixed for put, delete, and update survives on this path. Capture once before start_transaction, same as d0dec075, and thread it through the per-op sync calls.

LeeroyHannigan (commented):
The put, delete, and update captures in d0dec075 are exactly right, the ttl_worker comment now states the real invariant, and #333 is exactly the issue I wanted on file. Two things left. One inline on the TransactWriteItems path. The other: your reply mentions a Restore-mode unit test, but I do not see it on 828561dc; the tautology test and the #[allow(dead_code)] are still there. Did a push go missing?

Your merge sequence works for me; this PR is now the only thing holding the queue.

LeeroyHannigan (approved):
Thanks @diegotoledano95 🚀

PR #311, opened 2026-08-25
fix(mongodb): backup metadata

diegotoledano95:
## What

Preserve DynamoDB secondary-index metadata and throughput settings when creating and restoring MongoDB-backed backups.

Changes:

  - Store and restore GSI metadata.
  - Store and restore LSI metadata.
  - Preserve provisioned read and write capacity.
  - Preserve pay-per-request billing mode.
  - Preserve table class and encryption settings.
  - Omit empty GSI/LSI fields where DynamoDB omits them.
  - Retain compatibility fallback behavior for older backups.

## Why

  Backups previously restored indexed tables without their indexes and restored
  provisioned tables with hardcoded `5/5` capacity settings.

  This could change query behavior and billing configuration after restore.

## Testing done

  - Unit tests for provisioned-throughput preservation.
  - Unit tests for pay-per-request restoration.
  - Legacy-backup fallback tests.
  - Integration tests restoring GSIs and LSIs.
  - DescribeTable assertions for indexes, billing mode, and throughput.

## Integration dependency

  This PR depends on `fix/mongodb-gsi-backfill-race-safe` and must merge after it.

  The GSI PR refactors `backfill_gsi_batch` and introduces the explicit
  `GsiBackfillMode::Restore` path. After the GSI PR merges:

  1. Fetch the updated upstream `main`.
  2. Rebase this branch onto `origin/main`.
  3. Update `backfill_restored_indexes` to use the new context API.
  4. Explicitly use `GsiBackfillMode::Restore`.
  5. Run the combined build and MongoDB integration tests.
  6. Push the rebased branch with `--force-with-lease`.

This PR should remain a draft or be marked as dependent until those steps are complete.

## Checklist

  ## Breaking changes

  None.

Push only after rebasing onto the merged GSI change:

  git push --force-with-lease -u fork fix/mongodb-backup-metadata


crates/storage-mongodb/src/backup_engine.rs
```diff
@@ -52,6 +55,105 @@ fn now_epoch_secs() -> f64 {
...
+                .map_err(|_| StorageError::Internal("missing restored index_id".to_string()))?
+                .to_owned();
+            let index_key_schema: Vec<KeySchemaElement> =
+                decode_required(&index_doc, "key_schema")?;
+            let projection: Projection = decode_required(&index_doc, "projection")?;
+
+            let mut cursor_id = None;
+            loop {
+                let progress = self
+                    .backfill_gsi_batch(
```
LeeroyHannigan: **Should-fix**: the PR description says this depends on the backfill-race branch and its `GsiBackfillMode::Restore`, but this branch compiles and passes standalone against the current signature. When the race PR lands first, this call needs updating to pass `Restore` explicitly, defaulting to `Live` would put guard writes and per-item transactions on the restore path. Can you confirm the rebase plan covers that?

LeeroyHannigan (commented):
Confirmed both defects against main, restore really did force 5/5 and drop every index. Ran the new tests plus the wider backup suite against a live replica set, all green, and the GSI query round-trip after restore is a great assertion. Leaving streams/TTL/deletion-protection unset matches real RestoreTableFromBackup behavior, so no concerns there. One clarification below.

crates/storage-mongodb/src/backup_engine.rs
diegotoledano95: Yes. this PR’s rebase plan covers that:

  - Merge PR310 first.
  - Rebase PR311.
  - Update backfill_restored_indexes to the new signature.
  - Pass GsiBackfillMode::Restore explicitly.

  It will not use Live, otherwise restore would unnecessarily perform guard writes and per-item transactions.

crates/storage-mongodb/src/backup_engine.rs
```diff
@@ -52,6 +55,105 @@ fn now_epoch_secs() -> f64 {
...
+            read_capacity_units: 5,
+            write_capacity_units: 5,
+        }))
+    } else {
+        None
+    }
+}
+
+impl MongoEngine {
+    async fn backfill_restored_indexes(&self, table_id: &str) -> Result<(), StorageError> {
```
LeeroyHannigan: **Blocking** This runs the whole index backfill synchronously inside the RestoreTableFromBackup request, one replace_one per item per index; a 100k-item table with 2 GSIs is ~200k sequential round trips, so client timeouts fire first, and any error or disconnect strands the table in CREATING forever (the deferred flip stays Null and the worker only picks up CREATING GSIs, these are born ACTIVE). Once this rebases onto #310, moving it behind the worker in Restore mode resolves both.

crates/storage-mongodb/src/backup_engine.rs
```diff
@@ -176,13 +328,25 @@ impl BackupEngine for MongoEngine {
                 "key_schema": key_schema_bson,
                 "attribute_definitions": attr_defs_bson,
```
LeeroyHannigan, line 298: **Blocking** The table doc, the indexes, and the $out snapshot are three separate unsessioned reads, so an UpdateTable landing between them can capture a GSI without its key attribute definition. On restore, sk_info returns None and the index silently builds partition-only, wrong query results forever from a backup that looked healthy. Cheapest guard: after reading indexes, verify every captured index key attr exists in the captured attribute definitions and fail the backup loudly if not.

tests/test_mongodb_backup_restore.py
```diff
@@ -0,0 +1,206 @@
...
+        ]
+        assert throughput["ReadCapacityUnits"] == 0
+        assert throughput["WriteCapacityUnits"] == 0
+    finally:
+        _cleanup(dynamodb_client, restored, source)
+        if backup_arn:
+            dynamodb_client.delete_backup(BackupArn=backup_arn)
+
+
+def test_restore_preserves_provisioned_throughput_and_secondary_indexes(
```
LeeroyHannigan, line 125: **Should-fix** The legacy fallback coverage is helper-level only; the fixture already fetches the container name, so one test that $unsets the new fields on a real backup doc via mongosh and restores it end to end would cover genuine pre-fix documents. A query against the restored LSI would be good too, right now it's never read after restore.

LeeroyHannigan (commented):
The metadata capture and back-compat read path look right, and the ACTIVE-flip ordering avoids the restore bug we fixed on the sqlite side. Two things below I'd want settled before this lands. Also, the body says this should stay draft until #310 merges, but it's currently open and mergeable; worth flipping to draft so it can't land out of order.

LeeroyHannigan (commented):
The capture-atomicity guard and the legacy end-to-end test are exactly what I asked for, and persisting the mode on the index doc settles the Restore question nicely.

One blocking issue in the new worker lifecycle, inline. Two smaller notes: a client disconnect during the $out copy still leaves the table stuck in CREATING (a much smaller window now, fine as a follow-up issue), and this branch still carries #310's commits, so let's settle sequencing there before either merges. Please reply on the threads so I can resolve them.

crates/storage-mongodb/src/backup_engine.rs
```diff
@@ -534,20 +676,15 @@ impl BackupEngine for MongoEngine {
...
+            // The base `$out` copy is complete, but restored secondary-index
+            // collections are still empty. Mark the table as pending restore
+            // completion and leave its indexes CREATING so the shared worker
+            // can backfill them in bounded, restartable batches.
             let status_update = doc! {
-                "$set": { "item_count": item_count, "table_status": "ACTIVE" },
-                "$unset": { "status_transition_at": "" },
+                "$set": {
+                    "item_count": item_count,
+                    "restore_backfill_pending": true,
```
LeeroyHannigan, line 693: **Blocking:** the restored index docs are already CREATING before the $out copy starts, and the worker picks up any CREATING index without checking this flag, so a worker tick landing mid-copy scans the empty collection, reports done, and flips the index ACTIVE with zero rows; the table then activates with permanently empty indexes and silently wrong query results. Gate restore-mode jobs on restore_backfill_pending (or defer CREATING until after the copy), and add a test that forces a worker tick during the copy.

crates/storage-mongodb/src/backup_engine.rs
diegotoledano95: Addressed. RestoreTableFromBackup now copies the base data with $out, returns without performing one index write per restored item, and records the indexes as Restore-mode CREATING jobs. The shared worker processes those jobs in bounded batches, and the table becomes ACTIVE only after the restored indexes are complete. The client-disconnect case is being tracked separately in #334 .

crates/storage-mongodb/src/backup_engine.rs
diegotoledano95: Addressed. After reading the captured index metadata, the backup path now verifies that every index key attribute exists in the captured attribute definitions. If an index refers to a missing attribute, backup creation fails explicitly instead of producing metadata that could restore with an incorrect partition-only index.

tests/test_mongodb_backup_restore.py
diegotoledano95: Addressed. The test now removes the new metadata fields from a real MongoDB backup document through mongosh, restores that backup end to end, and queries the restored Local Secondary Index. This verifies compatibility with an actual pre-change backup document rather than only testing the decoding helper.

crates/storage-mongodb/src/backup_engine.rs
diegotoledano95: Addressed. Restore-mode jobs now require restore_backfill_pending before the worker can scan or activate them. The integration test uses the existing table-scoped test gate to hold the restore after its CREATING index metadata exists, forces a worker tick before the completion marker is set, releases the restore, and verifies that the restored Global Secondary Index contains all expected items.

diegotoledano95:
@LeeroyHannigan Agreed. I created #334  to track recovery when a restore request is interrupted before the base copy completes and restore_backfill_pending is recorded. The current change narrows the unsafe window and prevents premature index activation; the recovery behavior is isolated in the follow-up issue.

 PR #310 must merge first. I will rebase #311 onto the updated main immediately afterward, verify the explicit Restore mode and worker guard again, and keep #311 from merging before that rebase.

LeeroyHannigan (approved):

PR #312, opened 2026-08-25
fix(mongodb): dotted attribute paths

diegotoledano95:
## What

Preserve DynamoDB’s literal-field behavior for valid attribute names containing dots.

Changes:

  - Prevent unsafe condition pushdown for dotted attribute names.
  - Evaluate dotted conditions using literal-field semantics.
  - Read dotted TTL attributes with `$getField`.
  - Convert dotted TTL values numerically before filtering.
  - Prevent expired items from being starved during cleanup.
  - Avoid creating misleading physical TTL indexes for dotted attributes.
  - Correct TTL index cleanup when TTL is disabled.
  - Clarify that `ttl_index_ready` indicates TTL cleanup readiness; dotted
    attributes use the literal-field fallback without a physical index.

## Why

  DynamoDB permits attributes such as `expires.at` and `a.b` as literal names.
  MongoDB normally interprets dots as nested-object separators.

  Without these changes, conditions and TTL cleanup could read a different field
  from the one addressed by DynamoDB.

## Testing done

  - Literal dotted-field unit tests.
  - Dotted TTL filtering and numeric conversion tests.
  - TTL starvation tests.
  - Conditional-expression tests.
  - DynamoDB API tests for enabling and disabling TTL.
  - Direct MongoDB assertions for physical TTL-index lifecycle.
  - `cargo test --release -p extenddb-storage-mongodb ttl_query_`
  - `git diff --check`

## Checklist

  - [x] Documentation was updated

## Breaking changes

  None.


tests/rust/src/update_item_edge.rs
```diff
@@ -47,6 +47,42 @@ async fn update_with_nested_map() {
     assert_eq!(map.get("inner").unwrap().as_s().unwrap(), "nested_value");
 }
 
+#[tokio::test]
+async fn update_literal_dotted_attribute_with_alias() {
```
LeeroyHannigan: **Should-fix**: none of the new tests writes BOTH a literal `a.b` and a nested map `a = {b: ...}` on the same item and asserts they read back independently. A backend that exploded dots on write and re-collapsed on read would pass everything here. One item carrying both forms closes that hole.

LeeroyHannigan (commented):
Nice work on this one, the three-layer split (pushdown refusal, native-update bailout, $getField for TTL) is the right shape and I confirmed literal `a.b` and nested `a` into `b` stay structurally distinct in storage. Live TTL tests pass including the no-physical-index path for dotted names. The postgres and sqlite signature changes are behavior-preserving. One test ask below.

LeeroyHannigan (commented):
Went back over this one tracing the TTL delete pipeline end to end. The routing approach is the right architecture for the dot problem, and the literal/nested test pair is exactly the right discriminator. Two things to close out:

tests/rust/src/ttl.rs
```diff
@@ -86,6 +86,38 @@ async fn enable_ttl() {
     assert_eq!(ttl.attribute_name().unwrap(), "expires_at");
 }
 
+#[tokio::test]
+async fn enable_ttl_with_dotted_attribute_name() {
```
LeeroyHannigan, line 90: **Blocking** This checks describe metadata but nothing inserts an expired dotted item and asserts it actually gets deleted; the dotted delete path is a five-hop chain and a break anywhere in it is invisible to the suite. The plain-attribute sweep test shows the harness supports it, a dotted twin would close this.

crates/storage-mongodb/src/metadata_engine.rs
```diff
@@ -343,12 +429,16 @@ impl MetadataEngine for MongoEngine {
...
-
             let table_id = table_doc
                 .get_str("table_id")
                 .map_err(|_| StorageError::Internal("missing table_id".to_string()))?;
 
-            let ttl_attribute = table_doc.get_str("ttl_attribute").unwrap_or_default();
+            // Dotted attributes never have a physical TTL index: MongoDB's
+            // field-path syntax would reinterpret the literal attribute name.
+            if ttl_attribute.contains('.') {
+                return Ok(());
```
LeeroyHannigan: **Blocking** Deployments that enabled dotted TTL before this fix have a physical idx_ttl_<name> index from the old code, and this early return makes it permanently undroppable through the API. Still attempting drop_one on the legacy name and swallowing IndexNotFound (the caller already tolerates drop failures) cleans those up for free.

crates/storage-mongodb/src/data_engine.rs
```diff
@@ -35,6 +35,20 @@ use crate::pushdown::{Pushable, is_pushable};
 
 use extenddb_core::types::{Projection, ProjectionType};
 
+/// Resolve a single-component update path for the native MongoDB fast path.
+/// Literal attribute names containing dots cannot be represented safely by
+/// MongoDB's ordinary update paths, so they must use the Rust/session path.
+fn native_attribute_name(path: &[PathElement], maps: &ExpressionMaps) -> Option<String> {
```
LeeroyHannigan, line 42: **Should-fix** Same bug class one character over: an attribute literally named $foo makes the native path emit an update Mongo rejects, so the customer sees an internal error where the session fallback would have handled it fine, exactly as it does for dots. Extending this predicate to dollar-prefixed names routes them the same way.

crates/engine/src/ttl.rs
```diff
@@ -111,9 +111,12 @@ pub async fn handle_update_time_to_live(
     } else {
         // Disable path: metadata already updated (sweeper won't pick up this table).
         // Drop the index. Safe because sweeper checks ttl_index_ready which is now FALSE.
+        let ttl_attribute = current.attribute_name.as_deref().ok_or_else(|| {
```
LeeroyHannigan: **Nit** This fires after update_ttl has committed, so a metadata inconsistency turns a succeeded disable into a customer-visible 500, while an index-drop failure on the next line is just a warn. Resolving the attribute before the commit keeps the failure policy consistent.

crates/storage-mongodb/src/pushdown.rs
```diff
@@ -67,9 +67,56 @@ impl Pushable {
...
 
+/// Return whether an expression contains a literal DynamoDB attribute name
+/// with a dot. Such a name must be read as one field in `item_data`, while a
+/// MongoDB dotted field path means nested-document traversal. The compiler
+/// cannot safely push these expressions down, so callers must use the Rust
+/// evaluator instead.
+fn contains_dotted_attribute(expr: &Expr, maps: &ExpressionMaps) -> bool {
+    match expr {
+        Expr::Path(elements) => elements.iter().any(|element| match element {
+            PathElement::Attribute(name) => {
```
LeeroyHannigan, line 83: **Should-fix:** same bug class as the dots: this gate only refuses dotted names, so a ConditionExpression on an attribute literally named $foo still passes and we emit an item_data.$foo path into the Mongo filter on the pushdown paths. Mirror the native_attribute_name check here (refuse a leading $) and add one condition-path test.

LeeroyHannigan (commented):
Both blockers and the earlier test gap are fixed exactly as asked, nice work, and the TTL end-to-end test is a good addition.

One small remainder inline. Same sequencing note as the siblings: this branch still includes #310's commits, so the diff here shrinks once that lands. Please reply on the threads so I can resolve them.

tests/rust/src/update_item_edge.rs
diegotoledano95: Addressed. The test now writes both forms on the same item and verifies they remain independent when read back. This confirms literal dotted attributes are not confused with nested document paths.

tests/rust/src/ttl.rs
diegotoledano95: Addressed. Added expired_item_with_dotted_ttl_attribute_is_deleted, which inserts an expired item using a dotted time-to-live attribute and verifies the worker deletes it.

crates/storage-mongodb/src/metadata_engine.rs
diegotoledano95: Addressed. Disabling time-to-live still attempts to remove the legacy physical index. A missing-index error is tolerated, while existing legacy indexes are cleaned up.

crates/storage-mongodb/src/data_engine.rs
diegotoledano95: Addressed. Native update path handling now rejects dollar-prefixed names and routes them through the session fallback, just as it does for dotted names.

crates/engine/src/ttl.rs
diegotoledano95: Addressed. The time-to-live attribute is resolved before the metadata update commits. An invalid attribute therefore fails before the disable operation is recorded, keeping the failure behavior consistent with index-drop failures.

crates/storage-mongodb/src/pushdown.rs
diegotoledano95: Addressed. The condition pushdown analyzer now rejects literal attribute names beginning with $, including names resolved through aliases. Added dollar_prefixed_attribute_in_condition_is_not_pushable to verify the condition takes the safe fallback path.

LeeroyHannigan (commented):
The gate fix is exactly right: unsafe_attribute_reason covers dots, leading $ (alias-resolved too), and unresolved aliases across the same recursion the dot check had, and both pushdown callers route through it.

Two things before this can queue: the condition-path ask landed as a predicate unit test only, please add one end-to-end ConditionExpression test with an attribute literally named $foo (the dotted test at tests/rust/src/condition_expressions.rs:167 is the template); and the branch is conflicted with main after #314, so your newest commit has no CI on it. Per your sequence, rebasing onto main after #310 lands resolves both at once.

LeeroyHannigan (approved):
LGTM 🚀

PR #313, opened 2026-08-25
fix(mongodb): catalog connection limit

diegotoledano95:
## What

Honor MongoDB catalog connection settings and make numeric pool configuration overrides parse consistently.

Changes:

  - Apply `max_catalog_connections`.
  - Share the underlying MongoDB client pool between catalog and authentication
    operations.
  - Add shared string-to-number configuration coercion.
  - Support environment overrides such as
    `EXTENDDB__STORAGE__MONGODB__MAX_CONNECTIONS=10`.

## Why

  The catalog connection setting was declared but ignored. Catalog and
  authentication clients also created separate pool handles, making the effective
  connection budget unclear.

  Numeric environment overrides could additionally fail to parse because MongoDB
  configuration did not use the shared coercion behavior.

## Testing done

  - MongoDB configuration unit tests.
  - Environment-variable override tests.
  - Connection-pool configuration tests.
  - MongoDB build and integration validation.

## Checklist

  ## Breaking changes

  Unconfigured pool ceilings change. When `max_pool_size` / `max_catalog_connections` are omitted, the effective limit now comes from the connection string or the MongoDB driver default of 10, where previous builds applied fixed defaults of 50 (data) and 20 (catalog). Operators relying on the old implicit ceilings should set the values explicitly.


crates/storage-mongodb/src/lib.rs
```diff
@@ -123,6 +128,7 @@ fn server_components_factory(
             .and_then(|d| d.get_str("value").ok().map(std::borrow::ToOwned::to_owned))
             .ok_or(BackendError::MissingEncryptionKey)?;
 
+        let auth_client = catalog_client.clone();
```
LeeroyHannigan, line 153: **Should-fix**: auth now shares this pool with catalog admin ops and auth sits on every request, so a catalog burst can starve it. Fine to keep one pool, but the sizing guidance should say `max_catalog_connections` must cover auth concurrency, not just admin work. The postgres section of the sample config has that note; mongodb should too.

crates/storage-mongodb/src/config.rs
```diff
@@ -16,10 +16,16 @@ pub struct MongoStorageConfig {
...
+        deserialize_with = "extenddb_storage::config::string_coerce::u32"
+    )]
     pub max_connections: u32,
     /// Maximum concurrent connections for catalog/management operations
-    #[serde(default = "default_max_catalog_connections")]
+    #[serde(
+        default = "default_max_catalog_connections",
+        deserialize_with = "extenddb_storage::config::string_coerce::u32"
+    )]
     pub max_catalog_connections: u32,
```
LeeroyHannigan: **Should-fix**: `extenddb.sample.toml` still documents `max_pool_size` under `[storage.mongodb]`, which is not a field on this struct, and since there is no deny_unknown_fields an operator who uncomments it gets silently ignored config. Since this PR is about honoring these two fields, worth updating the sample to name `max_connections` and `max_catalog_connections` with their real defaults.

LeeroyHannigan (commented):
Verified the claims here, `max_catalog_connections` really was declared and never threaded through, and the clone-shares-the-pool comment is accurate for the Rust driver. Unit tests and a live boot with custom limits both pass. Two asks below.

crates/storage-mongodb/src/lib.rs
```diff
@@ -102,13 +103,17 @@ fn server_components_factory(
...
-            .map_err(|e| BackendError::ConnectionFailed {
-                backend: "mongodb".to_string(),
-                details: format!("Failed to create catalog client: {e}"),
-            })?;
+        // Create one shared catalog/management client. MongoDB Client clones
+        // share the underlying pool, so max_catalog_connections limits the
+        // combined catalog and auth traffic rather than creating two
+        // independent pools with twice the configured ceiling.
+        let catalog_client =
+            connect_guarded(&connection_string, Some(max_catalog_connections), false)
```
LeeroyHannigan: **Blocking** With the serde default this is now always Some, and connect_guarded unconditionally overwrites options.max_pool_size, so anyone tuning pools via mongodb://...?maxPoolSize=100 silently drops to 20 on upgrade. Either Option<u32> fields deferring to the URI when unset (postgres has opt_u32 for this), or an explicit breaking-change note in the description.

crates/storage-mongodb/src/config.rs
```diff
@@ -16,10 +16,16 @@ pub struct MongoStorageConfig {
...
-    #[serde(default = "default_max_connections")]
+    #[serde(
+        default = "default_max_connections",
+        deserialize_with = "extenddb_storage::config::string_coerce::u32"
+    )]
     pub max_connections: u32,
     /// Maximum concurrent connections for catalog/management operations
-    #[serde(default = "default_max_catalog_connections")]
+    #[serde(
+        default = "default_max_catalog_connections",
```
LeeroyHannigan: **Blocking** "0" parses fine here and then kills the server at startup with a connection-failure-shaped driver error (the driver rejects maxPoolSize=0). A >=1 bound at deserialization gives the operator an actionable message instead.

crates/storage-mongodb/src/config.rs
```diff
@@ -61,3 +67,45 @@ impl TryFrom<toml::Table> for MongoStorageConfig {
...
+mod tests {
+    use super::MongoStorageConfig;
+    use extenddb_storage::config::StorageConfig;
+
+    #[test]
+    fn numeric_connection_limits_accept_string_values() {
+        let config: MongoStorageConfig = toml::from_str(
+            r#"connection_string = "mongodb://localhost:27017"
+max_connections = "31"
+max_catalog_connections = "7""#,
```
LeeroyHannigan, line 151: **Blocking** These pin parsing but nothing asserts the catalog client actually receives the limit; reverting the lib.rs hunk leaves the whole suite green, which is the exact declared-but-ignored bug this PR fixes. One test asserting the assembled ClientOptions carries max_pool_size == configured closes the loop.

docs/design/13-storage-mongodb.md
```diff
@@ -870,6 +870,11 @@ max_connections = 50
 max_catalog_connections = 20
 ```
 
+The data-operation client uses `max_connections`. Catalog and authentication
+traffic share the client configured by `max_catalog_connections`; this is one
```
LeeroyHannigan: **Should-fix** Small correction: most catalog traffic (the table/metadata/backup engines and the workers) reaches extenddb_catalog through the data client's pool, not this one; this knob governs the management and auth store surface. Worth naming what rides each pool so operators size the right knob.

LeeroyHannigan (commented):
Verified the shared-client mechanism, the Arc-backed clone does share one pool, so the budget math works and there's no default regression. A few things around the edges from a closer look than I gave it last time:

LeeroyHannigan (commented):
All four asks are fixed cleanly, thank you.

Two small things before merge: the display mismatch inline, and the unconfigured defaults changed with this fix (data pool ceiling 50 to 10, catalog side similar) while the description still says no breaking changes; please add a note there, or restore 50/20 as the no-URI fallback. Then reply on the threads and I think we are done here.

crates/config/src/display.rs:77: **Should-fix:** when the fields are omitted this still prints 50/20, but the effective ceiling is now the URI value or the driver default of 10; print the configured value when set and something like 'from URI / driver default' otherwise.

crates/storage-mongodb/src/lib.rs
diegotoledano95: Addressed. The documentation now explains that the catalog connection limit is shared by management and authentication traffic, so it must be sized for their combined concurrency rather than administrative work alone.

crates/storage-mongodb/src/config.rs
diegotoledano95: Addressed. The sample configuration now uses the actual fields: max_connections and max_catalog_connections, with their real defaults and behavior documented.

crates/storage-mongodb/src/lib.rs
diegotoledano95: Addressed. Both fields remain optional. When omitted, the configured connection string pool setting is preserved. The explicit configuration value only overrides the connection string when the user provides one.

crates/storage-mongodb/src/config.rs
diegotoledano95: Addressed. Deserialization now rejects zero and requires every supplied connection limit to be at least one, producing an actionable configuration error.

crates/storage-mongodb/src/config.rs
diegotoledano95: Addressed. Added coverage that parses the connection options and verifies the resulting client options contain the configured pool limit.

docs/design/13-storage-mongodb.md
diegotoledano95: Addressed. The documentation now distinguishes the two pools: the data pool carries table, metadata, backup, worker, and related catalog operations, while the separate catalog pool carries management, credentials, and authentication traffic.

LeeroyHannigan:
Added the breaking-change note to the description (the unconfigured ceilings moved from 50/20 to URI or driver default 10) so the changelog picks it up. With that, this one is ready on my side.

LeeroyHannigan (approved):
LGTM 🚀

PR #319, opened 2026-08-27
fix: make the Query/Scan refusal on a vector index reachable

yesyayen:
## What

Query and Scan naming an existing vector index returned the index-not-found message, indistinguishable from a nonexistent index. The intended refusals existed in both handlers but were dead code: a vector index is not a row in the `indexes` catalog, so `index_info_by_table_id` failed with `IndexNotFound` and the error propagated before the `IndexType::Vector` guard could run.

Both handlers now re-resolve an unresolved index name against the table's vector index metadata (engine-level, so every backend inherits it; backends without vector support are unchanged) and refuse with the messages and precedence captured from Amazon DynamoDB.

All Amazon DynamoDB messages below were captured directly from the service (2026-08-20), on a table with an ACTIVE vector index `vidx` unless stated. Every case is `400 ValidationException` except the two controls.

| Scenario | ExtendDB (before this PR) | Amazon DynamoDB |
|---|---|---|
| `Query` with `IndexName` = vector index | `The table does not have the specified index: vidx` | `Query operation not supported on this index type.` (trailing period) |
| `Scan` with `IndexName` = vector index | `The table does not have the specified index: vidx` | `Scan operation not supported on this index type` (no period) |
| `Query`/`Scan` with a nonexistent `IndexName` | `The table does not have the specified index: nosuchindex` | same (unchanged contrast case) |
| `Query` on vector index + `ConsistentRead: true` | index-not-found message | `Consistent reads are not supported on global secondary indexes` (fires before the refusal, GSI wording even for a vector index) |
| `Scan` on vector index + `ConsistentRead: true` | index-not-found message | `Scan operation not supported on this index type` (refusal fires first) |
| `Query` on vector index, no `KeyConditionExpression` | index-not-found message | `Either the KeyConditions or KeyConditionExpression parameter must be specified in the request.` (fires before the refusal) |
| `Query`/`Scan` on vector index in CREATING, before backfill | index-not-found message | index-not-found message (same wording as a nonexistent index) |
| `Query` on vector index mid-backfill | index-not-found message | `Query operation not supported on this index type.` |
| `Scan` on vector index mid-backfill | index-not-found message | `Cannot read from backfilling global secondary index: vidx` |
| `Query`/`Scan` on the base table or a GSI of the same table | 200 | 200 (controls, unchanged) |

## Why

Conformance gap found by comparing ExtendDB against Amazon DynamoDB. The wrong message actively misleads: it denies that an index exists when the request's actual problem is naming a vector index in an operation that can never serve it.

Closes # n/a

## Testing done

Captured every expected message and ordering from Amazon DynamoDB first (including the lifecycle matrix, observed live through a backfill window), then validated dual-target. The branch was rebased onto current main after the vector-search backend PRs merged; all gates below were re-run at the rebased tip, not carried over.

- New pytest `tests/test_vector_index_query_scan.py` (12 cases: both refusals whole-string, not-found contrast, four measured precedence orderings, base-table and GSI controls).
- New wire tests in `tests/rust`: `query_and_scan_on_a_vector_index_are_refused` (refusals, contrast case, precedence, control) and `scan_during_backfill_gets_the_backfilling_wording`, which holds the backfill window open via the batch-delay setting, waits until `CREATING` with `Backfilling: true` is observed, and pins both mid-backfill messages in the same window.

## Checklist

ADR / RFC: n/a

---


LeeroyHannigan (approved):
LGTM 🚀

PR #321, opened 2026-08-27
fix(sqlite): serialize all writers through the engine write lock

yesyayen:
## What

The SQLite backend's concurrency design (D1, `crates/storage-sqlite/src/store.rs`) serializes every writer through the engine `write_lock` and states that `SQLITE_BUSY` cannot arise from competing writers. Seven writers ran outside the lock:

| Writer | Runs |
|---|---|
| `create_table_impl` (catalog + data-table DDL, two transactions) | every CreateTable |
| `create_ttl_index` (`CREATE INDEX` over a populated data table) + TTL metadata | TTL worker (60s), UpdateTimeToLive |
| `tag_resource` / `untag_resource` | TagResource, UntagResource |
| `refresh_table_size` (UPDATE) | worker (300s) |
| idempotency-token cleanup (DELETE) | worker (600s) |
| stream-record cleanup (DELETE) | worker (3600s) |
| `delete_backup` (bulk DELETE, one row per backed-up item) + `update_continuous_backups` | DeleteBackup, and automatically after every PITR restore |

An unserialized writer with a slow commit holds the SQLite file lock past a locked writer's 5s `busy_timeout`; the locked writer then fails an unrelated request with `database is locked`, which the engine maps to a 500. This is the `run-integration-sqlite` failure on main ([run 33078534390](https://github.com/ExtendDB/extenddb/actions/runs/33078534390)): a plain `PutItem` on a fresh table returned `InternalServerError` and took 7s where neighbors take under 1s (the busy-timeout signature).

All seven writers now take the lock. Contention queues on the in-process mutex, which waits without failing.

Also: the integration jobs now dump the server syslog on failure (`journalctl -t extenddb`). `devtools/run-tests` restarts the server daemonized, so its logs go to syslog and the original failure left zero server-side evidence in CI.

## Why

Reproduced the mechanism locally at `e41f370`: an uncoordinated writer holding the file lock fails a plain `PutItem` with the byte-identical client error, and the server logs `storage internal error internal_error=error returned from database: (code: 5) database is locked`.

Closes # n/a

## Testing done

- 6 new tests pin the D1 invariant: hold `write_lock`, spawn the writer, assert no progress until release. Failing-first verified: with the lock removed from `create_table_impl`, the idempotency cleanup, `delete_backup`, and `update_continuous_backups`, the matching tests fail with `completed while the engine write lock was held (D1 violation)`.
- Deadlock audit: no caller of a newly locked method holds the lock at entry (backup restore takes it only after `create_table` returns; the PITR path acquires and releases per step).
- Full wire suite against the fixed release build: pytest 942 passed, comprehensive 331 passed (includes the failed test), Rust integration 496/496.

## Checklist

ADR / RFC: n/a (enforces the existing D1 design decision; no wire, trait, or on-disk change)

## Breaking changes

n/a

---

### DIFF
--- a/crates/storage-sqlite/src/store.rs
+++ b/crates/storage-sqlite/src/store.rs
@@ -17,6 +17,17 @@
 //! `SQLITE_BUSY_SNAPSHOT` (a deferred read-then-write whose snapshot is
 //! invalidated by another pool committing) rather than surfacing it as a 500.
 //! Reads run concurrently from the pool against WAL snapshots and take no lock.
+//!
+//! "All writers" includes the control-plane paths (CreateTable's DDL, TTL
+//! metadata, tagging) and the periodic maintenance workers (table-size
+//! refresh, TTL index creation, stream-record and idempotency-token cleanup),
+//! not just the item write paths. A writer outside the lock contends at the
+//! SQLite level instead, and when its commit is slow (a large `CREATE INDEX`,
+//! a stalled fsync on a loaded CI host) a concurrent locked writer exhausts
+//! `busy_timeout` and fails an unrelated request with `database is locked`,
+//! which the engine maps to a 500. Measured 2026-08-27: an uncoordinated
+//! writer holding the file lock fails a plain `PutItem` with exactly the
+//! `InternalServerError` seen in the `run-integration-sqlite` CI flake.
 
 use std::sync::Arc;
 use std::sync::atomic::AtomicU64;
--- a/crates/storage-sqlite/src/store.rs
+++ b/crates/storage-sqlite/src/store.rs
@@ -28,6 +28,13 @@
 //! which the engine maps to a 500. Measured 2026-08-27: an uncoordinated
 //! writer holding the file lock fails a plain `PutItem` with exactly the
 //! `InternalServerError` seen in the `run-integration-sqlite` CI flake.
+//!
+//! Deliberate exclusions from the lock, so the invariant stays auditable:
+//! init-time bootstrap in this file (runs before the server serves traffic),
+//! and the management/credential stores, which write through the separate
+//! catalog pool in `lib.rs`. For a file-backed database that second pool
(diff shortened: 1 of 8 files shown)


crates/storage-sqlite/src/store.rs
```diff
@@ -17,6 +17,24 @@
...
+//! refresh, TTL index creation, stream-record and idempotency-token cleanup),
+//! not just the item write paths. A writer outside the lock contends at the
+//! SQLite level instead, and when its commit is slow (a large `CREATE INDEX`,
+//! a stalled fsync on a loaded CI host) a concurrent locked writer exhausts
+//! `busy_timeout` and fails an unrelated request with `database is locked`,
+//! which the engine maps to a 500. Measured 2026-08-27: an uncoordinated
+//! writer holding the file lock fails a plain `PutItem` with exactly the
+//! `InternalServerError` seen in the `run-integration-sqlite` CI flake.
+//!
+//! Deliberate exclusions from the lock, so the invariant stays auditable:
```
LeeroyHannigan, line 32: **Should-fix** `set_backfilling` and `mark_active` in `data/vector_index.rs` still run `UPDATE vector_indexes` on the engine pool with no write lock (their siblings `backfill_batch` and `reset_data_table` on the same struct take it), so a slow locked commit can exhaust their busy_timeout and fail a build step. Take the lock in both, or add them to this exclusion list so the invariant stays auditable.

yesyayen: good call! Confirmed and fixed: both flips now take the write lock, with mark_active holding it for the whole method. 

This finding is also a fair point about the pattern itself: the lock is a per-site convention today. Planning a follow-up that packages lock + BEGIN IMMEDIATE behind a single write_tx() helper so an unserialized write transaction cannot be expressed; kept out of this PR to avoid scope creep.

LeeroyHannigan (approved):

PR #325, opened 2026-08-31
ci(sign): keyless signing leg (default off) + zero-prod-contact rehearsal

robinnsc:
## What

**Draft — staged rollout; merging changes zero production behavior.**

1. The `sign-image` composite action gains a `keyless` input (**default
   `false`**): when enabled, cosign exchanges the job's OIDC token for a
   short-lived Fulcio certificate bound to the workflow run's identity and
   signs **additively** — the keyless signature lands as another layer on
   the same tag-based `.sig` artifact, next to the KMS signature, per the
   recorded dual-signing decision. `key-arn` becomes optional (empty skips
   the keyed leg — rehearsal use only).
2. A `keyless-rehearsal` workflow proves the entire path with **zero
   production contact**: scratch image in a dedicated GHCR package, signed
   through the production composite action, then — certificate identity
   captured from the signature itself, artifact shape asserted against the
   mirror/promote gate requirements, verification by identity, and a
   wrong-identity **negative control**.

## Infrastructure changes required

**None.** No AWS (KMS untouched; dual-signing keeps it), no new GitHub
environments or secrets: Fulcio needs `id-token: write` (signing jobs
already carry it), Rekor reachability from runners is proven since v0.1.8.
The trust infrastructure is Sigstore's public instance — which is also the
honest caveat: keyless verification depends on Fulcio/Rekor availability
and TUF roots, which is why the KMS signature remains the air-gap answer.

## Why keyless

Ends public-key distribution (verify by identity, not by PEM), and
**resolves the documented one-key caveat**: dev and postgres images are
signed by different workflows, so identity-based verification distinguishes
image tiers natively — something key-based verification cannot do.

## Rollout plan (the high-confidence gate)

- **A.** Merge (inert — flag off everywhere).
- **B.** Dispatch `keyless-rehearsal` repeatedly (it is cheap and
  credential-free). It empirically pins the cosign v3 flag set for
  keyless + tag-based storage on real runners — the exact class of flag
  interaction that produced the ECR storage-mode incident — and prints the
  ACTUAL certificate identity strings.
- **C.** From the captured identities: write the verify documentation and
  admission-policy guidance (identity per workflow = per image tier).
- **D.** One-line `keyless: 'true'` in `release-image` /
  `release-dev-image` (separate trivial PR). First dual-signed release
  verifies both ways; nothing existing changes — `cosign verify --key`
  continues to pass against the KMS layer, so the drift verifier, all
  runbook procedures, and every published instruction remain valid.

## Testing done

- `yaml.safe_load` on both files; the keyless step reuses the pinned,
  checksum-verified cosign install and the validated annotation handling.
- **Deliberately not claimed:** that the keyless flag set is correct — that
  is what the rehearsal exists to establish on real runners before any
  production caller enables the flag. The action's keyless leg self-checks
  (verify-by-identity) and the rehearsal adds the negative control, so a
  wrong flag set fails loudly in rehearsal, not in a release.

## Checklist

- [ ] Tests / fmt / clippy — not applicable, CI only
- [x] Added or updated tests — the rehearsal (with negative control) is the test, run pre-enablement by design
- [x] Documentation — headers document semantics; verify docs come in phase C from captured identities
- [x] Breaking changes noted below

ADR / RFC: n/a — CI tooling; the signing-model decision (dual, additive) is already recorded in the release runbook.

## Breaking changes

None. Default-off; the keyed signature and all existing verification paths are unchanged.

---

### DIFF
--- a/.github/actions/sign-image/action.yml
+++ b/.github/actions/sign-image/action.yml
@@ -29,8 +29,11 @@ inputs:
     description: 'Release the digest belongs to (vX.Y.Z); caller must have validated it'
     required: true
   key-arn:
-    description: 'KMS key ARN'
-    required: true
+    description: 'KMS key ARN (empty skips the keyed leg — rehearsal use only; production always dual-signs)'
+    default: ''
+  keyless:
+    description: 'Also sign keyless: Fulcio certificate bound to this workflow run identity, Rekor logged. Requires id-token: write in the calling job. Additive — the keyed signature is unaffected.'
+    default: 'false'
   cosign-version:
     description: 'Pinned cosign version'
     default: v3.1.3
@@ -69,6 +72,7 @@ runs:
         cosign version
 
     - name: Sign - attached, legacy tag-based storage, Rekor transparency log
+      if: inputs.key-arn != ''
       shell: bash
       env:
         IMAGE_REPO: ${{ inputs.image-repo }}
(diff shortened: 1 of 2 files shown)


.github/workflows/keyless-rehearsal.yml
```diff
@@ -0,0 +1,139 @@
...
+  CRANE_VERSION: v0.21.9
+  CRANE_SHA256: 5c16d8ddb971cb1d5e6ed8b1e743da8224414eeba2c2762d8f1a61b2f095699e
+
+jobs:
+  rehearse:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write    # Fulcio: the certificate binds to this run's identity
```
LeeroyHannigan, line 55: **Blocking:** this packages: write token reaches the production GHCR package: ghcr-mirror.yml pushes ghcr.io/extenddb/extenddb-postgres with secrets.GITHUB_TOKEN, GITHUB_TOKEN package scope cannot be narrowed to a single package, and this job has no environment while the mirror job sits behind the 4-reviewer ghcr gate. Attach the job to a branch-restricted environment (or move the scratch package out of this token's reach), and correct the header comment claiming the scratch package is the only capability.

.github/actions/sign-image/action.yml
```diff
@@ -69,6 +72,7 @@ runs:
         cosign version
 
     - name: Sign - attached, legacy tag-based storage, Rekor transparency log
+      if: inputs.key-arn != ''
```
LeeroyHannigan, line 75: **Should-fix:** this guard moves key-arn from fail-loud to silent-skip: with an unset KEY_ARN in a production caller, both keyed steps skip, and on a digest already carrying a stale .sig tag the gate passes green with no fresh keyed signature, despite the input description saying production always dual-signs. Add a leading validation step that fails unless key-arn is non-empty or keyless is true, and have production callers assert their ARN.

LeeroyHannigan (commented):
Strong work on the gating. I verified the default-off claim holds (composite input, no production caller passes it, no fork-reachable trigger, no injectable input), the pinning is exemplary, and the release artifact shape is unchanged while the leg is off. One Blocking inline on the rehearsal job's token and one Should-fix inline on the key-arn guard. Two more here:

**Should-fix:** the keyless leg pins only --new-bundle-format=false while the keyed leg also pins --use-signing-config=false and --tlog-upload=true, so cosign v3 defaults and the TUF signing config govern behavior the keyed leg deliberately pins. Pin the same set, or document which defaults are load-bearing and why the legs diverge.

**Should-fix:** the identity regexp in the self-check accepts any workflow in this repo at any ref. Fine as a smoke check, but annotate it as smoke-check-only in both files so it never gets copied into verify docs or an admission policy; published verify instructions should pin the release workflow file and a tag ref.

Nit: COSIGN_VERSION/COSIGN_SHA256 in the rehearsal env are dead (the action call passes neither), so they can drift; pass them through or delete them.

PR #326, opened 2026-08-31
fix(postgres): serialize transactional writes racing to create the same item

LeeroyHannigan:
## What

Fixes a create race in the Postgres backend's `TransactWriteItems` Put, Update, and Delete arms when the target item does not exist yet.

The write arms guard same-item concurrency with a locking read (`SELECT ... FOR UPDATE`), but a row that does not exist cannot be locked, so two transactions racing to create the same item both observe it as absent and both pass their condition checks. The loser's upsert (`ON CONFLICT DO UPDATE`) then silently overwrote the winner without re-checking its condition, so concurrent `attribute_not_exists` puts all reported success. On tables with synchronous index maintenance (any LSI, or a GSI with propagation delay 0) the loser also skipped the index-row delete (its stale pre-image said no old item existed) and its bare index `INSERT` collided with the winner's committed index row, surfacing `SQLSTATE 23505` as a 500 `InternalServerError`. The Delete arm could likewise physically delete a row committed by a racing creator after its locking read saw nothing, emitting no stream record and orphaning index rows.

The fix arbitrates the create with an atomic `INSERT ... ON CONFLICT DO NOTHING` (new `insert_item_if_absent_in_tx` helper):

- A loser re-reads `FOR UPDATE` (the arbiter insert already waited out the winner's commit) and re-evaluates its condition against the winner's committed item, so `attribute_not_exists` cancels with `ConditionalCheckFailed`, with `ReturnValuesOnConditionCheckFailure=ALL_OLD` carrying the winner's item. A condition that passes against the winner overwrites, as it should.
- A transactional Update re-applies its update expression on top of the winner's item and re-runs post-update validations on the recomputed item, matching the merge semantics on the non-transactional UpdateItem path.
- Index maintenance receives the winner as the old item, so it replaces the index row instead of colliding; the loser's stream record is a MODIFY with the winner as old image.
- Retry exhaustion (sustained create-then-delete churn) cancels with a `TransactionConflict` reason, matching the MongoDB backend's exhaustion path, never a 500.
- The Delete arm only issues its DELETE when the locking read saw (and therefore locked) a row; deleting a nonexistent item is a no-op on the real service (measured), so the skipped write is the faithful serialization against a racing creator.

Same race family as the earlier fixes on the non-transactional UpdateItem and conditional PutItem paths; the transactional arms had not received the pattern. SQLite is unaffected (process-wide write serialization).

## Why

Concurrent conditional Puts inside single-op `TransactWriteItems` targeting the same new item (an idempotent-create pattern) return 500 `InternalServerError` on tables with a GSI/LSI under same-item write contention (`SQLSTATE 23505: duplicate key value violates unique constraint` on the index table), and on plain tables every racer reports success, losing the conditional guarantee entirely.

## Testing done

Five racing integration tests in `tests/rust/src/transact_write_concurrency.rs` (barrier-synchronized writers, 20 rounds each, SDK retries disabled so a 500 cannot be hidden by client retries), run via `devtools/run-tests --extenddb --rust-integration --backend postgres`:

- conditional Put race, plain table: exactly one winner; every loser's typed cancellation is a single `ConditionalCheckFailed` reason whose ALL_OLD item equals the winner's committed item; stored item is the winner's.
- conditional Put race, LSI table: same, plus no 500 and the index holds exactly the winner's row (consistent-read Query).
- condition passing against the winner: all racers must succeed (overwrite, not cancel).
- conditional Update race: one winner, losers cancel with the winner's item.
- unconditional Update race on a fresh key: all succeed and attributes merge.

Negative control: the conditional-Put and merge tests fail against a pre-fix build exactly as reported (all racers succeeding on a plain table; `InternalServerError` under an LSI) and pass post-fix. Full Postgres `--rust-integration` sweep: 498 passed; the 3 `capacity_throttling` failures reproduce identically on the pre-fix binary in this local setup and are unrelated. All transact/transaction modules (54 tests) green after the Delete-arm change. `extenddb-storage-postgres` unit tests, rustfmt, and clippy clean.

Measured against the real service (us-east-1, 8 barrier-synchronized writers, 6 rounds, SDK retries disabled): exactly one `attribute_not_exists` Put wins every round; losers overlapping the winner's in-flight transaction cancel with a `TransactionConflict` reason ("Transaction is ongoing for the item", no Item), and losers arriving after commit cancel with `ConditionalCheckFailed` ("The conditional request failed", ALL_OLD Item = the winner's item). This backend serializes the overlap window instead of conflict-cancelling it, so its losers deterministically take the measured post-commit CCF shape; the exhaustion path's `TransactionConflict` message matches the service verbatim. Simultaneous passing-condition puts and unconditional transactional updates conflict-cancel on the live service (clients reach the all-succeed outcome only after retrying), so those assertions are backend-specific and the tests skip against real DynamoDB.

## Checklist

ADR / RFC: n/a (bug fix restoring documented DynamoDB transaction semantics; no interface change)

---


jcshepherd (approved):
LGTM. I'm looking forward to seeing how Cassandra performs with the new tests.

This general issue of racing writes has me thinking more and more about the effects of queuing writes to the same key on the DDB storage node, hot key behavior, and if/how we want to emulate that in ExtendDB. We're able to work around the absence (and, I think, difficulty) of queuing, but I'm not sure about emulating hot keys, split for heat, etc. (assuming we want to).

PR #330, opened 2026-09-01
feat(cassandra): DynamoDB-compatible TTL expiration

robinnsc:
## What

Adds DynamoDB-style TTL to the Cassandra backend.

Cassandra cannot query an arbitrary TTL attribute inside a JSON item efficiently, and native Cassandra TTL would remove only the base row — it would not clean ExtendDB's secondary-index tables or write the DynamoDB-compatible service `REMOVE` stream record. This implementation therefore keeps a sharded expiration queue in each account keyspace and removes expired items through ExtendDB's normal data model.

The queue is partitioned by table, TTL-enable generation, day bucket, and one of 64 key shards, so lookups use complete partition keys and an ordered clustering range with no `ALLOW FILTERING`. Enabling TTL backfills existing items and publishes the queue as ready only for the generation it actually scanned; disable, re-enable, table deletion, cleanup, and concurrent workers are all fenced by that generation.

Deletion is a durable state machine: `PENDING` → `CLAIMED` → (`CLAIM_ABORTING` | `EFFECTS_APPLYING`) → `EFFECTS_APPLIED`. A claim persists the old item image, a stable delete timestamp, and a deterministic stream identity; the two exits from `CLAIMED` are mutually exclusive LWTs, so an abort and the must-complete effects transition cannot release each other's fence. The worker seals the base owner, cleans synchronous indexes and writes the stream record while it owns the base row, then performs an exact conditional base-row delete, then completes the queue entry — and every conditional destroy of a live queue row is preceded by a quorum-durable repair marker, so an ambiguous outcome or crash always leaves an obligation a dedicated repair worker can act on. Every phase is idempotent and entered only from durable state, so a crash at any point retries the same work without deleting a renewed item or publishing a second visible `REMOVE`. ADR-0010 has the full crash-point/recovery table.

Once TTL is enabled, ordinary Put and Update take the same exact row claim, and their base mutation, TTL outbox/queue work, synchronous index changes, and stream record commit in one logged batch. Non-TTL tables keep the target branch's existing fast paths and same-key OCC behavior.

## How this compares to the PostgreSQL TTL implementation

The bar for this feature is the production readiness of the PostgreSQL TTL implementation, so it matters which limitations are Cassandra's and which are the shared TTL design. ADR-0010 carries the same breakdown with file references.

PostgreSQL derives expiry from the item itself: a `CREATE INDEX CONCURRENTLY` expression index over `(item_data->'ttl'->>'N')::BIGINT`, an indexed range scan to find candidates, and then an ordinary conditional `delete_item` whose base delete, index cleanup, stream record, and async-GSI enqueue are one `BEGIN`/`COMMIT`. There is no queue, no per-item work state, and nothing to coordinate. Cassandra can do neither half of that — no expression index over JSON, and no cross-partition atomic conditional write — which is the entire reason for the queue, the claims, and the state machine.

**Shared with PostgreSQL — same shared handler or same worker shape, not Cassandra limitations:**

| Behaviour | Why it is shared |
|---|---|
| ~100 expirations per table per minute, and no gain from a larger fleet | Both backends use `SCAN_INTERVAL = 60s` and `BATCH_SIZE = 100`. PostgreSQL runs without a lease, but every host issues the same `ORDER BY ttl LIMIT 100` query, so hosts contend for the same rows and the losers' deletes fail their TTL condition. |
| `UpdateTimeToLive` blocks instead of returning immediately | The shared handler awaits `create_ttl_index` on every backend. |
| No `ENABLING`/`DISABLING` status | `TimeToLiveStatus` has only two variants. |
| No five-year cutoff on old timestamps | PostgreSQL matches with `BETWEEN 1 AND now`. |
| TTL deletion bypasses write-capacity accounting and throttling | Both workers call the storage layer directly, below the request capacity path. |
| TTL configuration is not cached | Neither backend caches it. Cassandra pays a per-write catalog read because it needs the configuration on the write path at all; PostgreSQL never reads it. |

**Where Cassandra is now stricter than PostgreSQL:**

| Behaviour | Detail |
|---|---|
| Fractional TTL values | Cassandra ignores `N: "1.5"`. PostgreSQL casts the stored text to `BIGINT` in both its expression index and its sweep query, so a fractional value looks able to raise a runtime error rather than being ignored. Worth confirming separately — not this PR's to fix. |
| Backfill restart safety | A failed Cassandra backfill loses only its scan position; already-registered entries survive because inserts are conditional. A failed PostgreSQL `CREATE INDEX CONCURRENTLY` can leave an invalid index of the same name, after which `IF NOT EXISTS` no-ops and `ttl_index_ready` is set anyway. |
| Duplicate sweep work | Cassandra's per-table lease means the work is done once; PostgreSQL hosts each re-read the same candidate rows every cycle. |

**Genuinely Cassandra-specific:**

| Gap | Cause | Fixable? |
|---|---|---|
| Deletion is a crash-recoverable saga, not one atomic transaction, so a brief internal effects-before-delete window exists | Cassandra cannot combine a Paxos condition on the base row with mutations in other partitions. | **No — this is the fundamental blocker.** It is what the queue, the claims, and the state machine exist to compensate for. |
| TTL cannot be enabled alongside an asynchronously propagated GSI | The async GSI queue has no version-conditional replay fence, so an old TTL delete could overtake a recreated item's insert. PostgreSQL enqueues async GSI cleanup inside the delete transaction and has no such race. | No — needs versioned, conditionally applied GSI mutations. **This is the only functional capability PostgreSQL has that Cassandra does not.** |
| Enabling TTL on a live table needs writes quiesced | The queue must be backfilled, and a write that began before enable cannot join the new generation. `CREATE INDEX CONCURRENTLY` covers live writes with no transition window. | No — needs a durable background backfill that admits pre-enable writes. |
| Writes on a TTL-enabled table cost one claim LWT and a logged batch (the extra catalog read is overlapped with the index read; the release is folded into the batch, with an exact conditional release off the request path) | The claim protocol is on the write path; PostgreSQL writes touch nothing TTL-related. | Partly — see the open question at the end. |
| Writes for a table must be served from one datacenter | All claims and lifecycle LWTs use `LOCAL_QUORUM`/`LOCAL_SERIAL`. | No — global serial consistency would cost cross-region Paxos on every write. |
| The claim and exact delete fence on `item_data` string equality | There is no version column to condition on. | Yes, as a follow-up. |
| Transaction *recovery* reconciles insert-only | The ledger does not persist the pre-commit image. | Yes, as a follow-up. |

Net: this matches PostgreSQL on every shared behaviour, is stricter on two, and the remaining differences all trace back to the absence of cross-partition atomic conditional writes.

## Design decisions and the risks they carry

Five decisions carry most of the risk. Each is a deliberate trade, and each has a named failure mode.

**1. Expiry is a durable queue, not a derived index.** Cassandra has no expression index over JSON, so the queue is unavoidable. The cost is that the queue is now a second source of truth that can drift from the items: an item with no queue entry never expires, and an entry with no item is garbage. Every mitigation in this PR exists to keep those two in step — the in-batch queue mutation, the durable outbox as a fallback, and revalidation before any delete. **Risk: silent non-expiration.** It is the failure mode to watch, because nothing errors when it happens.

**2. Deletion is a saga fenced by a base-row claim, not a transaction.** The claim serializes expiry against ordinary writes; the state machine makes each phase idempotent. `EFFECTS_APPLIED` is the one phase that must go forward — index rows are already deleted and the `REMOVE` already published, so abandoning it would leave a live item with a missing index. Generation cleanup honours the same rule. **Risk: the effects-before-delete window.** Bounded by the claim, and now by the drain rule, but not eliminated.

**3. Claims are time-bounded, and contention is retried rather than surfaced.** No claim is unbounded: an unbounded claim makes a key permanently unwritable if the queue row that drives recovery is lost. Losing a claim is treated as contention — the write re-reads and retries — so ordinary concurrent writes stay last-writer-wins. **Risk: a claim leak still costs write availability for that key** until the claim expires, and sustained contention on one key eventually surfaces `TransactionConflictException`.

**4. Absence decisions read at quorum; everything else tolerates a stale read.** Claim and delete LWTs re-verify the exact image, so a stale read is caught. Decisions about *absence* — retire this queue work, drop this outbox row, retire this bucket registration — have nothing to condition on, and the default read consistency is `ONE`. **Risk if this rule is broken: silent non-expiration**, which is exactly what it was.

**5. Lifecycle changes take a lease rather than being rejected.** Enable, disable, generation cleanup, and index creation that could invalidate a live sweep all take the same lease and retry briefly, rather than failing the caller with a table-state error. **Risk: `ResourceInUseException` under sustained sweep collision**, and a lease leak blocks lifecycle changes until it expires.

## Supported envelope

* **Expiration rate: about 100 items per table per minute.** The shared TTL worker's rate, not a Cassandra property. A table expiring faster than roughly 144k items/day accumulates backlog.
* **Single-datacenter writes per table**, because claims and lifecycle LWTs are `LOCAL_QUORUM`/`LOCAL_SERIAL`.
* **Enable before opening a new table to traffic**, or briefly quiesce an existing table until enable and backfill return. Not enforced by the API.
* **No asynchronously propagated GSI on a TTL-enabled table.** Rejected in both directions.
* **Synchronized clocks**, and request deadlines well below the 120-second request-claim lifetime.
* **Verified against Cassandra 4.1.** Nothing here uses a version-gated CQL feature, but no older version has been exercised.
* **Monitor** TTL backlog, worker health, and Cassandra LWT latency.

## Next round of changes

Separated by confidence, because the reasons differ.

### High confidence — I would take these as written

**1. Lease the sweep per `(table, shard)` instead of per table.** Raises the ceiling from ~100 to ~6,400 items/table/minute. High confidence because it needs **no protocol change**: every queue transition is already conditional on the exact work UUID, which is why two concurrent sweeps already pass the exactly-one-`REMOVE` test. The queue is already sharded 64 ways into disjoint partitions. The only new work is moving lifecycle fencing off the sweep lease — and sweeps already re-read the TTL config and renew between rows, so a generation bump already stops them. This is the one change that would take Cassandra past the PostgreSQL rate rather than matching it.

**2. Add a version column and fence on it instead of `item_data`.** High confidence because the OCC path already maintains a `version` column, so the mechanism exists; this is a narrowing of what the LWT condition compares. It removes the dependency on a byte-stable JSON encoding (currently pinned only by a unit test) and stops shipping whole items — up to 400 KB — as LWT condition values.

**3. Persist the pre-commit image in the transaction ledger.** High confidence because the live path already does this correctly; only recovery is insert-only, and the fix is one column plus reusing `reconcile_ttl_transition`.

**4. Checkpoint the backfill scan cursor.** High confidence in the mechanism — the scan is already paged and resumable in-process, so this is persisting the page token. It does not by itself fix the blocking `UpdateTimeToLive`.

**5. Add a Cassandra integration workflow.** There is none. `ttl_integration.rs` requires a live Cassandra and never runs in CI, so every test in this PR is a local-only signal. High confidence because the other backends already have workflows to copy.

### Potential — worth doing, but I would want to prove something first

**6. Remove the claim from the ordinary write path.** Potentially the largest win: it would delete two Paxos rounds and a catalog read from every write on a TTL-enabled table, and remove the contention class entirely. The reason I would not commit to it yet is that I cannot currently prove the claim is unnecessary. Its remaining value is closing the window between `EFFECTS_APPLIED` and the base delete, which the bounded-claim and drain changes here narrow but do not obviously eliminate. **What would settle it:** a written argument that the worker's exact conditional delete alone is sufficient, plus a benchmark quantifying what the claim actually costs. If the cost is small, the safety is worth keeping.

**7. Cache the per-table TTL configuration.** Removes a catalog read from every write. Potential rather than high confidence because the correctness risk is the wrong shape: a stale "disabled" entry means writes skip the queue entirely and those items never expire, and a stale generation means writes enqueue into a retired one. **What would settle it:** deciding the invalidation story — probably a short TTL plus generation validation on use — and confirming the read is actually a measurable cost rather than an assumed one.

**8. Report `ENABLING`/`DISABLING`.** Straightforward for Cassandra, where `ttl_index_ready` already carries the state. Potential because `TimeToLiveStatus` is a shared core enum, so this changes every backend and the API surface. **What would settle it:** agreement that it is a product-wide change rather than a Cassandra one.

**9. Support TTL alongside asynchronously propagated GSIs.** The only functional gap versus PostgreSQL. Potential because it needs a versioned, conditionally applied GSI replay fence — its own design, and larger than this feature. **What would settle it:** whether any customer wants the combination, since rejecting it is honest and safe today.

**10. Make `sweep_once` scopeable rather than global.** Every sweep lists every TTL-ready table across every account. Correct, but the cost grows with the size of the deployment, and it is why this suite's runtime grows with the age of the local Cassandra. Potential because it interacts with (1) — per-shard leasing may want a different work-discovery shape anyway.

### Deliberately not planned

A backend-wide durable mutation protocol. It is the strongest general answer to Cassandra's cross-partition limits and would subsume several items above, but it is a storage-engine rewrite: it changes every write path, adds journal traffic to tables that do not use TTL, and needs its own capacity, compaction, and migration design. It should be evaluated on its own merits, not as part of TTL.

## Risk to existing behavior

Shared write paths are touched, so the intended blast radius:

* `put_get_item.rs`, `update_item.rs`, `delete_item.rs` — every TTL branch is gated on `ttl_config.is_some()`. With TTL disabled the paths are the target branch's, including the `attribute_not_exists` → `INSERT IF NOT EXISTS` fast path, which is now free of TTL work that previously cost it an extra catalog read.
* `concurrent_owner_error()` preserves the existing `TransactionCanceledException` when TTL is disabled and returns the retryable `TransactionConflict` only when it is enabled.
* `get_item_quorum` is additive; `get_item_impl` and therefore `GetItem` are unchanged.
* `stream_util.rs` — `stream_record_statement` keeps its behavior; the identity-aware variant is new and used only by TTL.
* `index.rs` — only a visibility widening on `effective_delay`.
* `update_table.rs` — the control lease is taken only for GSI creation on a TTL-enabled table or for an async GSI; other `UpdateTable` shapes are unchanged apart from the body moving into `apply_table_update`.
* `delete_table.rs` — the data table is now dropped before the TTL queue is cleared, so a failed drop cannot leave a live TTL-enabled table with an emptied queue.
* The schema is folded into the initial catalog and data migrations rather than added as new versioned ones — the backend is pre-release, so no install needs the in-place upgrade path. That leaves `migrations.rs`, `bootstrapper.rs`, and `operations.rs` untouched by this change. A keyspace that already recorded `V001` will not pick the new columns up, which is the accepted trade.

## Testing done

Real Cassandra 4.1, single node, `127.0.0.1:9042`:

* `cargo test -p extenddb-storage-cassandra --test ttl_integration -- --test-threads=1` — **24 passed, 0 failed**
* `cargo test -p extenddb-storage-cassandra --test metadata_integration` — **1 passed, 0 failed**
* `cargo test -p extenddb-storage-cassandra --lib` — **43 passed, 0 failed**
* `cargo clippy -p extenddb-storage-cassandra --all-targets` — 6 warnings remain and **all 6 pre-date this branch** (verified present at `5ee3fb2`: `index.rs:395`, `query.rs:518`, `transaction_ledger.rs:31`, `transaction_ledger.rs:94`, `transactions.rs:215`, `stream_util.rs:139`). This branch introduces none. CI runs `cargo clippy --all-targets -- -D warnings`, so those pre-existing warnings fail the clippy job on the target branch independently of this change.
* `cargo fmt --all -- --check`, `git diff --check` — clean

Coverage: lifecycle and generation fencing, renewal and stale-candidate races, transaction reconciliation, synchronous GSI cleanup, deterministic stream failure recovery, exactly-one visible `REMOVE`, queue-claim conflicts, orphaned-owner recovery, metadata-only row recreation, non-TTL fast-path behavior, both directions of the async-GSI restriction, concurrent ordinary writes on a TTL-enabled table, draining a `CLAIMED` row versus completing an `EFFECTS_APPLIED` one, drained-bucket retirement, `EFFECTS_APPLYING` recovery with a changed image under the sealed owner, shared-key GSI restoration through recovery, deterministic repair-scan progress (300 markers traversed in exactly two passes; a full page of poison outbox rows does not hide the record behind it), late destroys, outbox fairness, and lifecycle recovery for the new queue states.

**Not covered, and listed as follow-ups:** an orphaned base-row claim with no corresponding queue row; disable racing a genuinely in-flight sweep (the test holds the lease manually); multi-host concurrent backfill; the exact-delete-false interleaving itself (it cannot be forged mid-flow — the survivor path it routes into is what the forged-state tests cover).

⚠️ **There is no Cassandra integration workflow.** `ttl_integration.rs` and `metadata_integration.rs` require a live Cassandra at `127.0.0.1:9042` and are neither `#[ignore]`d nor feature-gated, so `cargo test --workspace` fails without one and these tests never run in CI. See follow-up 5.

A full core-workspace build is still blocked before Cassandra validation by an unrelated dependency mismatch on the target branch: `pgvector 0.4.2` uses SQLx 0.9 traits while `extenddb-storage-postgres` uses SQLx 0.8.6. The Cassandra package and its affected dependency path build and test successfully.

## Review history

The first revision was reviewed; this revision addresses those findings.

1. **Claim contention is no longer a client-visible error.** The losing writer previously got `TransactionCanceledException` with no retry — an error `PutItem` does not have, and one SDKs will not retry. The dominant path turned out to be the pre-read seeing `prepared_txn_id`, not the claim LWT, so both now report the retryable error and all three write paths retry against a re-read image.
2. **No unbounded claims.** The worker's base-row claim had no TTL, so any path that removed its queue row left the key permanently unwritable. It is now bounded, which is safe because the exact base delete already conditions on both the work UUID and the item image.
3. **Retiring a generation drains it instead of deleting it.** Cleanup removes only `PENDING` rows; `CLAIMED` work is released and abandoned, `EFFECTS_APPLIED` work completes its base delete. This closes the case where disabling TTL, or deleting a table whose drop then failed, could leave a live item with deleted index rows and nothing to recover it. The racy `has_inflight_ttl_work` pre-check is gone.
4. **Sweep fan-out is bounded.** Bucket-registry rows were never retired, so per-cycle query fan-out grew linearly with table age. Drained past buckets are now retired with a timestamp-guarded delete, and partitions visited per cycle are capped.
5. **The backfill is leased**, so one host scans a table at a time instead of every host rescanning it every cycle.
6. **Lease hygiene in `update_table`.** The control lease leaked on several `?` returns, and synchronous GSI creation was not fenced at all, letting a sweep see an ACTIVE index whose data table did not exist.
7. **Transactional writes retire the previous queue entry** by reconciling against the pre-commit image; `Delete` ops are reconciled too, where before they were skipped.
8. **Absence decisions now read at quorum.** The worker and the outbox both treated an absent item as authoritative and acted on it destructively, using reads at consistency `ONE`. A lagging replica reads as absent for an item that exists, which leaves that item with no expiration entry — **it would never expire**.
9. **One unusable outbox row no longer starves reconciliation.** Found by the full-suite run: a per-row read failure propagated out of the whole pass, so a single bad row blocked reconciliation for every other item. Failures are now confined to their row.
10. **Test assertions tightened.** Seven tests would have passed while the behaviour they name was broken — a unit test that only counted batch statements, a stream test that never checked the delete completed, a bucket test that never checked the drain happened, and several `is_some()`/`Ok` checks that never verified an effect. The four tag tests, which test nothing TTL-related, moved to `metadata_integration.rs` as one phased test.
11. **Parity framing corrected.** An earlier draft presented the throughput ceiling, the blocking `UpdateTimeToLive`, and the missing `ENABLING`/`DISABLING` as Cassandra limitations. They are product-wide.
12. **ADR corrected.** It claimed the logged batch used a "request-start timestamp"; it is pinned immediately after the claim, and the ADR now states the invariant that actually holds. `differences-from-dynamodb.md` advertised a configurable `ttl_deletion_target_seconds` (default 300s) that appears in no Rust code anywhere in the repo.

### Second review pass (jcshepherd)

13. **TTL schema folded into the initial migrations** rather than added as `V004`/`V003`. Also removes the `{{keyspace}}` templating that existed only to let those two files qualify their statements, so three files drop out of the diff.
14. **ADR gained a "How it works" section** — prose walking the lifecycle end to end, since the decision bullets read as detail without framing.
15. **`push_query` helper** in `data/ttl.rs` replaces the `mem::replace` batch-append repeated at four call sites.
16. **Module doc on `ttl_integration.rs` corrected** — it still said "tag operations" from the file it was copied from. The tag tests it referred to have moved to `metadata_integration.rs`.

### Third review pass (jcshepherd, performance) and two hardening rounds on it

17. **Post-commit reconcile removed** (a redundant catalog read plus a Paxos insert per Put/Update) — the durable outbox already guarantees the same repair. The worker also receives the sweep's config instead of re-reading the catalog per row, and the config read remaining on the write path is overlapped with the index read.
18. **The outbox stays, with its cost cut.** Removing it loses items: the queue table takes concurrent Paxos writes from the worker, which can supersede a batch's queue insert by cell timestamp and erase it outright with an abort tombstone; a worker crash before its compensating reconcile then leaves an item that never expires. Instead, the outbox row is skipped when the item carries no valid TTL, and the background pass answers with a plain quorum read (no Paxos) when the entry is already present.
19. **Claim release folded into the batch** — the batch clears the claim columns at its pinned timestamp, and a bounded detached task performs the exact conditional release off the request path. Net per TTL-enabled write: three Paxos operations become one, three catalog reads become two, overlapped.
20. **The version-column fence is deferred with its prerequisite named**: ordinary Put does not maintain `version`, so fencing on it today would let the worker delete an item a Put had just renewed. It needs every write to bump `version` first; then it delivers what was intended — smaller LWT payloads and one concurrency mechanism instead of two.
21. **Queue destroys gained a quorum-durable repair handshake** (found while pressure-testing point 18: a repair obligation could be discharged in the gap between being written and the destroy it guards). Every conditional destroy of an active-generation row now writes an inflight marker first, in one logged batch with its discovery-registry row; a definitive LWT result hands off to the normal outbox before the marker is removed; ambiguity or a crash leaves the marker for a dedicated repair worker that reconciles it until quorum metadata proves the generation retired. No wall-clock or skew assumption remains in the protocol.
22. **The state machine gained `CLAIM_ABORTING` and `EFFECTS_APPLYING`** — mutually exclusive exits from `CLAIMED`, so lifecycle cleanup and a sweep cannot release each other's fence. The base owner is sealed (no TTL, exact owner+image LWT) immediately before `EFFECTS_APPLYING`; from that durable intent onward the work must complete, and recovery is fenced on the exact owner alone — never the image, which a stale writer's older-pinned unconditional batch can legally change under the seal. A failed exact base delete re-reads at quorum and takes the same survivor path. In both cases the survivor's synchronous index rows are rebuilt from quorum-read metadata at a timestamp strictly above the replayed effects tombstones, before the sealed owner is released.
23. **Background repair progress is deterministic, not probabilistic**: the marker scan resumes each cycle where the last page ended and wraps (N markers traversed in ceil(N/256) cycles), and the outbox pass pages past persistently unreconcilable rows on a keyset cursor that advances on the row id alone, so a poisoned prefix cannot hide valid records and a half-readable row cannot stall the cursor on itself.
24. **Honest cost accounting.** The request path is cheaper; total Cassandra work is not: background reconciliation is roughly nine or ten mostly-sequential quorum operations per TTL-bearing write, plus about four extra quorum writes per successful expiration and a per-account repair scan each minute — all bounded by the sweep's ~100 expirations/table/minute ceiling. A benchmark is a named follow-up before any broad production-performance claim.

**Still open — the claim on the ordinary write path.** See follow-up 6. It is an architectural change rather than a review fix, so it is not in this revision; happy to take it as a separate PR if reviewers prefer that shape.

## Checklist

* [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
* [x] The affected Cassandra package builds and all applicable tests pass against real Cassandra
* [x] Code is formatted (`cargo fmt --all -- --check`)
* [x] This branch introduces no new Clippy warnings
* [x] Tests cover the new behavior, the merge-conflict regressions, and the review fixes
* [x] Documentation describes behavior, trade-offs, production limits, PostgreSQL parity, and known gaps
* [x] No public wire-protocol, `Storage` trait, auth, or CLI breaking change is introduced
* [x] The backend schema and design decision are recorded in the ADR


crates/storage-cassandra/migrations/catalog/V004__ttl_generation.cql
```diff
@@ -0,0 +1,9 @@
+-- Copyright 2026 ExtendDB contributors
+-- SPDX-License-Identifier: Apache-2.0
+-- Fence TTL enable, backfill, sweep, and cleanup cycles.
+
+ALTER TABLE {{keyspace}}.tables ADD IF NOT EXISTS ttl_generation uuid;
```
jcshepherd: I'll leave this to your judgment, but since the plug-in is still in development, I'm not opposed to just updating the original migration/schema in-place. On the other hand, it's nice to exercise the migration functionality. Your call.

docs/adr/0010-cassandra-ttl-expiration-queue.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Context
+
+DynamoDB TTL is more than storage expiry. Removing an expired item must also remove its secondary-index rows and emit a Streams `REMOVE` record whose identity is the DynamoDB service. Cassandra native TTL deletes a base row internally, bypassing both effects, and an item stored as JSON cannot be efficiently searched by an arbitrary configured TTL attribute.
+
+Cassandra also cannot atomically combine a Paxos condition on the base item with queue, index, and stream mutations in other partitions. The PostgreSQL backend is therefore a semantic reference, not an implementation whose ACID internals can be reproduced exactly.
+
+## Decision
+
+Use a generation-fenced, durable TTL-specific state machine.
```
jcshepherd, line 50: This section could benefit from a paragraph or two of narrative about how TTL works (the process): customer enables TTL, we backfill something on the table, when a delete happens X, Y, Z occurs, there's a periodic TTL sweeper that looks for work in T, when it finds an entry that's passed its TTL, it does A, B, C. It's hard to get a clear picture from the bullet points, though they're probably great for filling in the detail once the framing is clear.

crates/storage-cassandra/tests/ttl_integration.rs
```diff
@@ -0,0 +1,1310 @@
+// Copyright 2026 ExtendDB contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration tests for MetadataEngine tag operations.
```
jcshepherd: Copy-pasta?

crates/storage-cassandra/src/data/ttl.rs
```diff
@@ -0,0 +1,1153 @@
...
+                "TTL reconciliation item missing key attribute {}",
+                element.attribute_name
+            ))
+        })?;
+        key.insert(element.attribute_name.clone(), value.clone());
+    }
+    let key_data = serde_json::to_string(&key).map_err(|error| {
+        StorageError::Internal(format!("Serialize TTL reconciliation key: {error}"))
+    })?;
+    let partition = (crc32fast::hash(key_data.as_bytes()) % TTL_SHARDS as u32) as i32;
```
jcshepherd, line 269: Could be pulled into a helper function.

jcshepherd (commented):
A couple comments. I'm going to need some time on this.

crates/storage-cassandra/migrations/catalog/V004__ttl_generation.cql
robinnsc: Swapped that to the in-place update approach to cut down on moving pieces with this overall change

docs/adr/0010-cassandra-ttl-expiration-queue.md
robinnsc: Sure, added a section for background context and the user flow

crates/storage-cassandra/tests/ttl_integration.rs
robinnsc: Yeah, fixed

crates/storage-cassandra/src/data/ttl.rs
robinnsc: Done

jcshepherd (commented):
Thanks for addressing my earlier comments. I'm still working through this. I think I understand the enable/disable process and the sweep process now. My main concern is around the overhead (in terms of Cassandra roundtrips) in the path of item-mutating operations on TTL tables. I'm beginning to think we have future work to cache table metadata, similarly to how auth metadata is cached today, to reduce roundtrips to the Catalog keyspace. However, I think there may be some simplifications to this implementation that will also reduce roundtrips without loss of correctness.

1. Reconciliation table updates are redundant with the inline queue mutations

Every write to a TTL-enabled table currently does two things to keep the queue current:

• add_ttl_queue_mutations — inline in the logged batch, retires the old entry and inserts the new one into ttl_expirations
• add_ttl_reconciliation_mutation — also inline in the same batch, writes a key-only record to ttl_reconcile_pending so the background worker will later re-read the item and re-insert the queue entry

These are doing the same job and since both are in the same logged batch, they either both commit or neither does. If the batch commits, the queue is already correct; the reconciliation entry just causes the worker to redundantly re-read the item  and re-insert an entry that's already there.

The reconciliation record would then only be written in cases where the inline queue write can't happen, which would matter for transaction recovery paths where the queue write happens separately from the base mutation.

Removing the reconciliation update from ordinary writes eliminates one statement from every logged batch and one background worker round-trip per write.

2. The post-commit reconcile_ttl_item call is a third catalog read per write

After the batch commits, put_item_impl and update_item_impl both call reconcile_ttl_item, which calls ttl_config_for_table again. The config was already fetched at the top of the function and is in scope. reconcile_ttl_item could accept the config directly, or the post-commit call could be dropped entirely if the outbox redundancy (point 1) is resolved, since the inline queue mutations already did the work.

3. The claim + release pair can potentially use the OCC version column

The ADR's "known gaps" section flags this: the claim conditions on item_data string equality, which is fragile. The version column added for OCC would be a cleaner fence — IF version = ? AND prepared_txn_id = null is already the OCC write guard. The TTL claim could reuse that same condition rather than shipping the entire serialized item as an LWT condition value. This would also eliminate the item_data equality dependency that the ADR calls out as a protocol risk.

This is a bigger change because it requires the worker to read the version alongside the item, and the claim/release to use version rather than item equality. But it would reduce the LWT condition payload size significantly for large items, and it unifies the two concurrency mechanisms rather than having them be parallel systems.

robinnsc:
Thanks, for those points:

1. Reconciliation table updates are redundant with the inline queue mutations

  
The base mutation, queue mutation, and outbox record are in the same logged batch, so they do commit together. The
complication is that ordinary writes are not the only writers to the queue table. The expiration worker also updates and deletes queue rows through Paxos. Its updates can supersede a batch insert by cell timestamp, and its abort path can leave a row tombstone that removes a concurrent insert. If the worker crashes before repairing that state, the item can be left without a queue entry and never expire. Because of that, I don’t think it is safe to remove the outbox with the current queue design.
  
Reduced its costs so writes no longer create an outbox record when the new item has no valid TTL. During reconciliation, the worker first reads the queue row at quorum. If it is already PENDING, which is the normal case, no conditional insert is needed.
  
Working through also found a problem with the original repair mechanism where a repair record could be processed and removed while the queue deletion it was protecting was still in flight. That's now handled with an explicit handshake. Before conditionally deleting an active generation queue row, the worker writes an inflight repair marker at quorum. A completed LWT writes a normal outbox record before removing that marker. If the LWT result is ambiguous or the process crashes, the marker remains and a separate repair worker continues reconciling it until quorum metadata shows that its table generation has been retired. This part of the repair protocol no longer relies on a clock-skew allowance or time-based grace period.
  
Also tightened the deletion state machine. CLAIMED now transitions through either CLAIM_ABORTING or EFFECTS_APPLYING, using mutually exclusive LWTs, so cleanup and expiration cannot release each other’s ownership fence. Recovery in EFFECTS_APPLYING checks the owner rather than requiring the original item image to remain unchanged. If a stale write changes the image and causes the exact base delete to fail, recovery rereads the surviving item, restores any synchronous index rows removed by the replay, releases the owner, and reconciles the survivor’s TTL entry.

2. The post-commit reconcile_ttl_item call is a third catalog read per write

Yeah good callout, removed the post commit reconcile_ttl_item calls from Put and Update. The sweep now passes its existing TTL configuration into the row processor rather than reading it again. On the write path, the remaining TTL configuration read runs concurrently with the index metadata read.
  
3. The claim + release pair can potentially use the OCC version column

I do agree, though would be dependent on some changes i currently was planning as follow up work. The existing version column can't be used safely yet, as ordinary Put does not currently increment it, only put_item_if_not_exists and the OCC Update path do. An IF version = ? expiration check could therefore delete an item that an ordinary Put had renewed.
  
The prerequisite would be to make every relevant write path advance the same monotonic revision. Once that is in place, expiration can use the revision instead of sending the complete serialized item in its LWT condition.

jcshepherd:
Last thing, though I'm having a heck of time finding it in GitHub's diff view, but I do see it in the patch. There are a number of checks like:
+            if indexes.is_empty()
+                && stream_stmt.is_none()
+                && ttl_config.is_none()
+                && ttl_claim.is_none()
+                && !super::ttl::reconciliation_outbox_required()

... in update_item.rs and put_get_item.rs. tt:reconciliation_outbox_required() is hardcoded to return false, so this check always forces updates to go down the batch patch, regardless of any of the other conditions. I think you could remove this last check and the existing ttl-related checks in the same statement will cover the TTL case, while allowing simple tables (no indexes, streams or TTL) to go down the faster non-batch path.

Beyond that, at this point I think it's time to commit and the iterate on improvements as we need to. Thanks!

robinnsc:
Yeah good call, made those changes, and I squashed and merged this PR's commits into feat/storage-cassandra-in-tree

PR #331, opened 2026-09-02
feat(container): add MongoDB-backed ExtendDB image

diegotoledano95:
## What

  - Add a production `Dockerfile.mongodb` for the MongoDB-backed ExtendDB binary.
  - Add MongoDB license notices and release validation.
  - Add an optional local Compose stack with MongoDB replica-set initialization and ExtendDB bootstrap.
  - Add a MongoDB container smoke test covering startup, hardening, API operations, and persistence after restart.
  - Add release, signing, and registry propagation support for Docker Hub, GHCR, and ECR Public.
  - Document MongoDB container usage and clarify that the bundled unauthenticated MongoDB Compose stack is local-development-only.
  - Add deterministic replica-set checks, TLS-key permission checks, and a MongoDB-specific license compatibility guard.

## Why

This makes the MongoDB backend consumable as a versioned ExtendDB container image while keeping MongoDB as a separate customer-managed service.

  ## Testing done

  - `bash -n devtools/generate-software-license-notices`
  - `bash -n ci/smoke-test-mongodb-container.sh`
  - `git diff --check`
  - `docker compose -f docker-compose.mongodb.yml config --quiet`
  - GitHub Actions workflow YAML parsing
  - MongoDB license notices verified in both argument orders:
  - `--check --backend mongodb`
  - `--backend mongodb --check`
  - MongoDB container smoke test passed, including image hardening, DynamoDB operations, restart persistence, and cleanup.

The smoke test should be rerun after the final `compose build` / `compose up --no-build` startup-order adjustment.

Ran `ci/smoke-test-mongodb-container.sh`. Result: 

`PASS: MongoDB-backed ExtendDB container smoke test complete`

Ran `docker compose -f docker-compose.mongodb.yml config`. Result: 

`PASS: Compose preserved "$$MONGO_CONNECTION_STRING"`

## Checklist

  - [x] Tests were added or updated for the container functionality
  - [x] Documentation was updated
  - [x] No breaking changes
  - [x] No wire protocol, `Storage` trait, auth model, or public CLI changes

## Breaking changes

None. The MongoDB backend remains separately configured, and the production image does not bundle MongoDB itself.

Closes # n/a

ADR / RFC: N/A

---

### DIFF
--- a/docker/dockerhub-overview-mongodb.md
+++ b/docker/dockerhub-overview-mongodb.md
@@ -1,0 +1,78 @@
+# ExtendDB with MongoDB
+
+ExtendDB is a DynamoDB-compatible API server backed by your MongoDB deployment.
+This image contains ExtendDB only; MongoDB is a separate service that you
+choose and operate, including MongoDB Atlas or a self-managed replica set.
+
+## Quick start
+
+The MongoDB backend requires MongoDB 7.0 or newer configured as a replica set.
+Transactions and change streams are used by the backend, so a standalone
+MongoDB server is not sufficient.
+
+Pull the image:
+
+```console
+docker pull docker.io/extenddb/extenddb-mongodb:latest
+```
+
+Create a writable configuration directory and initialize it against your
+MongoDB deployment:
+
+```console
+mkdir -p ./extenddb-data
+cat > ./extenddb-data/bootstrap.toml <<'EOF'
+[storage]
+backend = "mongodb"
+
+[storage.mongodb]
+connection_string = "mongodb://mongodb.example.com:27017/?replicaSet=rs0"
+EOF
+
+docker run --rm \
+  -v "$PWD/extenddb-data:/var/lib/extenddb" \
+  docker.io/extenddb/extenddb-mongodb:latest \
+  init --backend mongodb \
+  --config /var/lib/extenddb/extenddb.toml \
+  --overwrite \
+  --bind-addr 0.0.0.0 \
+  --tls-san localhost
+```
+
+Start the server using the generated configuration:
+
+```console
+docker run --rm --name extenddb-mongodb \
+  -p 18443:18443 \
+  -v "$PWD/extenddb-data:/var/lib/extenddb:ro" \
+  --read-only \
+  --tmpfs /tmp:rw,noexec,nosuid,size=64m \
+  --cap-drop=ALL \
+  --security-opt=no-new-privileges:true \
+  docker.io/extenddb/extenddb-mongodb:latest \
+  serve --config /var/lib/extenddb/extenddb.toml
+```
+
+The server listens on `https://127.0.0.1:18443` by default. Initialization
+creates a self-signed certificate and the first administrator credentials;
+save the password printed by `init`.
+
+## Configuration and security
+
+Keep the MongoDB connection string in the generated configuration or provide
+it through the configuration mechanisms documented in the repository. Use
+MongoDB authentication and TLS for production deployments. The ExtendDB
+container runs as an unprivileged user and is designed to run with a
+read-only root filesystem.
+
+For a complete deployment guide and an optional local reference stack, see
+[`docker/README-mongodb.md`](https://github.com/ExtendDB/extenddb/blob/main/docker/README-mongodb.md).
+
+## Versioning
+
+The image tag is the ExtendDB release version. The MongoDB server version is
+independent and is selected in your MongoDB deployment. Pin an ExtendDB
+version in production instead of relying on `latest`.
+
+ExtendDB is licensed under the Apache License 2.0. The image includes the
+third-party license notices required by its dependencies.
--- a/docker/dockerhub-overview-mongodb.md
+++ b/docker/dockerhub-overview-mongodb.md
@@ -36,7 +36,8 @@ docker run --rm \
   --config /var/lib/extenddb/extenddb.toml \
   --overwrite \
   --bind-addr 0.0.0.0 \
-  --tls-san localhost
+  --tls-san localhost \
+  --tls-san 127.0.0.1
 ```
 
 Start the server using the generated configuration:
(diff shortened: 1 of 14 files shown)


LeeroyHannigan (commented):
The image itself is in good shape: Dockerfile.mongodb is line for line the hardened postgres pattern, dev-mode with the mongodb backend is a compile error so there is no credential-seeding risk, and the CI wiring rides the same triggers as postgres. Two Blocking items inline on the Docker Hub overview, and four more before this can queue:

**Should-fix:** neither destination repo exists yet: Docker Hub extenddb/extenddb-mongodb and ECR Public both return 404 today (the postgres twins return 200). ECR Public does not auto-create on push (promote-image.yml's own note from #283), and the dockerhub-description push trigger fires on merge of THIS PR, so it hits the missing repo immediately. Create both repos public before merge.

**Should-fix:** the Testing done section says the smoke test should be rerun after the final startup-order change; c43dfeff changed the script after the recorded pass and it only executes at release dispatch. Please run it once now and record the pass here; release-time-only failures have bitten us twice.

**Should-fix:** the overview is missing the trademark and independence note and the tag and signature-verification sections the postgres overview carries; the disclaimer belongs on every public image front.

**Should-fix:** the PR body dropped the template's Closes #, ADR / RFC line, and CONTRIBUTING checklist item; please restore them (n/a is fine for the ADR).

Nits: drop the empty ci: retrigger commit before the squash folds its message in; the dev compose interpolates $MONGO_CONNECTION_STRING into TOML unescaped; README-mongodb never names the file path the bootstrap snippet must be written to; and init passes --tls-san localhost while the docs give the endpoint as https://127.0.0.1:18443, so the cert will not cover the bare IP.

docker/dockerhub-overview-mongodb.md
```diff
@@ -0,0 +1,78 @@
...
+```console
+docker pull docker.io/extenddb/extenddb-mongodb:latest
+```
+
+Create a writable configuration directory and initialize it against your
+MongoDB deployment:
+
+```console
+mkdir -p ./extenddb-data
+cat > ./extenddb-data/bootstrap.toml <<'EOF'
```
LeeroyHannigan: **Blocking:** this writes the connection string to ./extenddb-data/bootstrap.toml but the init command below reads exactly its --config path (MongoBootstrapper::from_config), so the quick start fails with Cannot read config file. Write the snippet to ./extenddb-data/extenddb.toml, as docker-compose.mongodb.yml does.

docker/dockerhub-overview-mongodb.md
```diff
@@ -0,0 +1,78 @@
...
+```console
+docker run --rm --name extenddb-mongodb \
+  -p 18443:18443 \
+  -v "$PWD/extenddb-data:/var/lib/extenddb:ro" \
+  --read-only \
+  --tmpfs /tmp:rw,noexec,nosuid,size=64m \
+  --cap-drop=ALL \
+  --security-opt=no-new-privileges:true \
+  docker.io/extenddb/extenddb-mongodb:latest \
+  serve --config /var/lib/extenddb/extenddb.toml
```
LeeroyHannigan: **Blocking:** this serve example overrides the image CMD and drops --foreground; serve daemonizes by default, tini's child exits, and the container dies immediately (README.md documents this exact pitfall). Add --foreground, and reconsider the :ro state mount, which diverges from every other documented serve invocation.

docker/dockerhub-overview-mongodb.md
diegotoledano95: Addressed. The quick-start now writes ./extenddb-data/extenddb.toml, matching the /var/lib/extenddb/extenddb.toml path used by initialization, the image healthcheck, and the default container command.

docker/dockerhub-overview-mongodb.md
diegotoledano95: Addressed. The example now passes --foreground, so the server remains attached to Docker’s main process. The state volume is writable, while --read-only still protects the container root filesystem. The documentation now matches Dockerfile.mongodb.

diegotoledano95:
@LeeroyHannigan 

Addressed and verified the remaining review items.

The two blocking documentation issues are fixed:

- The quick-start now writes extenddb.toml, matching the configuration path used by initialization.
- The serve command now includes `--foreground`, and the state volume is writable while the container root filesystem remains read-only.

I ran the container smoke test locally with:

`ci/smoke-test-mongodb-container.sh`

The image built, the MongoDB stack initialized, the hardening checks passed, the DynamoDB API round trip passed, persistence was confirmed after restartingExtendDB, and the run ended with:

`=== PASS: MongoDB-backed ExtendDB container smoke test complete ===`

I also verified the Docker Compose escaping with:

`docker compose -f docker-compose.mongodb.yml config  | grep -F 'connection_string = "$$MONGO_CONNECTION_STRING"'`

Output:

`connection_string = "$$MONGO_CONNECTION_STRING"`

The empty commit `1fbb90e ci: retrigger pull request checks` was removed from the branch history.

I restored the Closes # field, the CONTRIBUTING checklist item, and the ADR / RFC line in the pull request description. I believe the description now matchesthe repository template, but please let me know if theres something else needed there.

I will create both public repositories before merge:

- Docker Hub: extenddb/extenddb-mongodb
- ECR Public: public.ecr.aws/extenddb/extenddb-mongodb

robinnsc (approved):
Looks good, thanks. One small change, with a bit out of it outside the scope of this PR, that we'll likely pickup as a followup: the MongoDB sidecar is ${MONGODB_IMAGE:-mongo:7.0}, unpinned. The postgres compose stack pins its sidecar by digest, which we'd ideally want to match. However the postgres release can currently fail intermittently, if its sidecar pull hits a RateExceeded error. This new stack pulls its sidecar image in two services, so it'd be exposed to the same issue. We'll likely want to add a retry around the compose pull in both smoke tests.

PR #335, opened 2026-09-10
fix(backup): report point-in-time recovery as unsupported

LeeroyHannigan:
## What

Reports point-in-time recovery honestly as unsupported, served by the engine for all three backends:

- `DescribeContinuousBackups` returns `ContinuousBackupsStatus: ENABLED` (the service reports this unconditionally) with `PointInTimeRecoveryStatus: DISABLED` and no `EarliestRestorableDateTime` or `LatestRestorableDateTime`.
- `UpdateContinuousBackups` with `PointInTimeRecoveryEnabled: true` returns `ContinuousBackupsUnavailableException` (HTTP 400, `__type com.amazonaws.dynamodb.v20120810#ContinuousBackupsUnavailableException`), the exception the service models on that operation. With `false` it returns the DISABLED description.
- `RestoreTableToPointInTime` resolves the source table first, returning `TableNotFoundException` for a missing table as the live service does, then returns `PointInTimeRecoveryUnavailableException` (HTTP 400, `__type com.amazonaws.dynamodb.v20120810#PointInTimeRecoveryUnavailableException`), the exception the service models on the restore operation, with the live service's message shape ("Point in time recovery is not enabled for table '<name>'").

The `describe_continuous_backups`, `update_continuous_backups`, and `restore_table_to_point_in_time` methods are removed from the `BackupEngine` storage trait and their implementations deleted from the PostgreSQL, SQLite, and MongoDB backends. The stored per-table `pitr_enabled` flag drove nothing, so its storage is removed with the catalog moving to 0.0.4: PostgreSQL migration `003_drop_continuous_backups.sql` drops the `continuous_backups` table, SQLite's schema replaces the CREATE TABLE with `DROP TABLE IF EXISTS`, and MongoDB no longer creates the collection (an existing deployment keeps an orphaned collection that is harmless and safe to drop by hand; the upgrade manual says so).

Doc corrections: `docs/differences-from-dynamodb.md` gains a Backup and Restore section pointing at on-demand backups; `docs/dynamodb-limits.md` fixes three rows that contradicted the code (backup/restore is implemented, the GetRecords 1,000-record cap is enforced, the 15-minute shard iterator expiry is enforced). The upgrade manual no longer states which catalog version is current or what each migration file does: a new section quotes the tooling that reports the installed and expected versions (`extenddb verify`, `extenddb migrate` without `--yes`, and the startup mismatch error), the migrations directory is the reference for per-file purposes, and the wire behavior change is recorded under Behavior Changes by Release. The architecture guide loses its catalog version literals the same way, and the upgrade manual's exemption from the `doc_version_literals` guard is removed so the guard now covers it.

## Why

Point-in-time recovery reported a capability that did not exist. `DescribeContinuousBackups` answered ENABLED with a 35-day window, `UpdateContinuousBackups` accepted an enable request, and `RestoreTableToPointInTime` either returned a generic `ValidationException` or, at the backend layer, fabricated a restore of current state ignoring the requested timestamp. A restore that silently returns the wrong data is worse than a refusal the caller can act on. This is phase 0 (T0.1) of the backup and restore design, which lands as its own PR.

## Testing done

- `cargo fmt --all -- --check`: clean.
- `cargo clippy --all-targets -- -D warnings`: clean.
- `cargo test --workspace`: 1160 passed, 0 failed across 27 test binaries. Includes unit tests pinning the description shape (restorable-time fields absent on the wire) and each refusal's status, `__type`, and message, plus the migration tripwire tests updated for catalog 0.0.4. The `doc_version_literals` guard passes with the upgrade manual now covered.
- `cargo +1.88.0 check --workspace --locked`: clean.
- New `tests/test_continuous_backups.py` pins the surface through boto3 on every backend: describe shape, enable refused with `ContinuousBackupsUnavailableException`, disable returns DISABLED, restore on an existing table refused with `PointInTimeRecoveryUnavailableException`, restore on a missing table returns `TableNotFoundException`. The rewritten Rust integration tests pin the same behavior through the Rust SDK.
- `devtools/run-tests --extenddb --rust-integration --pytest --backend sqlite --filter backup` against a fresh SQLite deployment (server started on catalog 0.0.4): rust integration 19 passed, 0 failed; pytest 5 passed, 3 skipped (MongoDB-gated).
- Same against a fresh PostgreSQL 16 deployment (fresh init ran migrations 001 through 003): rust integration 19 passed, 0 failed; pytest 5 passed, 3 skipped (MongoDB-gated).
- `devtools/run-mongodb-tests -- --rust-integration --pytest --filter backup` (mongo:7 replica set in Docker, `mongodb-test-hooks` release build): rust integration 19 passed, 0 failed; pytest 8 passed, 0 failed (the 5 new tests plus the MongoDB backup metadata suite).

## Checklist

ADR / RFC: the backup and restore design spec PR (section 6, Point-in-time recovery in v1)

## Breaking changes

- **Storage trait**: `BackupEngine` loses `describe_continuous_backups`, `update_continuous_backups`, and `restore_table_to_point_in_time`. Out-of-tree backends implementing these methods must drop them; the engine now answers the three operations without consulting storage.
- **Wire behavior**: `UpdateContinuousBackups` with `PointInTimeRecoveryEnabled: true` previously succeeded and `DescribeContinuousBackups` then reported ENABLED with a 35-day window; both now report DISABLED and the enable request fails with `ContinuousBackupsUnavailableException`. `RestoreTableToPointInTime` moves from a generic `ValidationException` to `TableNotFoundException` for a missing source table and `PointInTimeRecoveryUnavailableException` otherwise. Clients or templates that enable point-in-time recovery (for example Terraform `point_in_time_recovery = true`) will now receive an error and must stop requesting it.
- **Catalog**: version moves to 0.0.4 on PostgreSQL and SQLite (the unused `continuous_backups` table is dropped). Existing deployments must run `extenddb migrate` before the new binary serves; see the upgrade manual. Note for the sqlx migration-runner work: `003_drop_continuous_backups.sql` carries its own `BEGIN`/`COMMIT` and version `UPDATE` for the current runner; a port to a runner that owns transactions and version writes must strip both, the same treatment already prescribed for 002.

---

### DIFF
--- a/crates/engine/src/backup.rs
+++ b/crates/engine/src/backup.rs
@@ -190,7 +190,81 @@ pub(crate) async fn handle_restore_table_from_backup(
     serialize_output(&json!({ "TableDescription": desc }))
 }
 
+/// The one message every point-in-time recovery refusal carries.
+const PITR_UNSUPPORTED_MESSAGE: &str =
+    "Point-in-time recovery is not supported by this storage backend";
+
+/// The typed refusal for enabling point-in-time recovery.
+///
+/// `ContinuousBackupsUnavailableException` is the exception the service models
+/// on `UpdateContinuousBackups`, so every SDK surfaces it as a typed error a
+/// caller can match on, unlike a generic `ValidationException` whose meaning
+/// lives only in the message text.
+fn pitr_unsupported_error() -> DynamoDbError {
+    DynamoDbError::ContinuousBackupsUnavailableException(PITR_UNSUPPORTED_MESSAGE.to_owned())
+}
+
+/// The typed refusal for `RestoreTableToPointInTime`.
+///
+/// The service models `PointInTimeRecoveryUnavailableException` on this
+/// operation and returns it whenever recovery is not enabled on the source
+/// table, which for ExtendDB is always. The message mirrors the live service's
+/// shape ("Point in time recovery is not enabled for table '<name>'").
+fn pitr_restore_unavailable_error(table_name: &str) -> DynamoDbError {
+    DynamoDbError::PointInTimeRecoveryUnavailableException(format!(
+        "Point in time recovery is not enabled for table '{table_name}'"
+    ))
+}
+
+/// The continuous-backups description every table reports.
+///
+/// `ContinuousBackupsStatus` is always `ENABLED`, matching the service, which
+/// reports `ENABLED` unconditionally. `PointInTimeRecoveryStatus` is always
+/// `DISABLED` because no storage backend implements point-in-time recovery,
+/// and the restorable-time fields are omitted, as the service omits them for
+/// a table whose recovery is disabled. Owned by the engine rather than the
+/// backends: the answer does not depend on stored state, so no backend is
+/// consulted and all three report identically.
+fn disabled_continuous_backups_description() -> extenddb_core::types::ContinuousBackupsDescription {
+    extenddb_core::types::ContinuousBackupsDescription {
+        continuous_backups_status: "ENABLED".to_owned(),
+        point_in_time_recovery_description: Some(
+            extenddb_core::types::PointInTimeRecoveryDescription {
+                point_in_time_recovery_status: "DISABLED".to_owned(),
+                earliest_restorable_date_time: None,
+                latest_restorable_date_time: None,
+            },
+        ),
+    }
+}
+
+/// Verify a table exists in the caller's account, in any lifecycle status.
+///
+/// The continuous-backups operations answer for CREATING and DELETING tables
+/// too, so this goes through `describe_table` rather than `table_key_info`,
+/// which only resolves ACTIVE tables.
+async fn require_table_exists(
+    ctx: &OperationContext,
+    table_name: &str,
+) -> Result<(), DynamoDbError> {
+    ctx.storage
+        .describe_table(
+            &ctx.account_id,
+            extenddb_core::types::DescribeTableInput {
+                table_name: table_name.to_owned(),
+            },
+        )
+        .await
+        .map_err(storage_err_to_dynamo)?;
+    Ok(())
+}
+
 /// Handle `DescribeContinuousBackups`.
+///
+/// Point-in-time recovery is not supported by any storage backend, so the
+/// engine answers directly: continuous backups report `ENABLED` (the service
+/// reports this unconditionally) with `PointInTimeRecoveryStatus: DISABLED`
+/// and no restorable-time window.
 pub(crate) async fn handle_describe_continuous_backups(
     body: Value,
     ctx: &OperationContext,
@@ -206,16 +280,20 @@ pub(crate) async fn handle_describe_continuous_backups(
             )
         })?;
 
-    let desc = ctx
-        .storage
-        .describe_continuous_backups(&ctx.account_id, table_name)
-        .await
-        .map_err(storage_err_to_dynamo)?;
+    require_table_exists(ctx, table_name).await?;
 
-    serialize_output(&json!({ "ContinuousBackupsDescription": desc }))
+    serialize_output(
+        &json!({ "ContinuousBackupsDescription": disabled_continuous_backups_description() }),
+    )
 }
 
 /// Handle `UpdateContinuousBackups`.
+///
+/// Enabling point-in-time recovery is refused with
+/// `ContinuousBackupsUnavailableException`, because no storage backend
+/// implements it and reporting it as enabled would promise a restore that can
+/// never happen. Disabling it is a no-op that returns the description already
+/// reported by `DescribeContinuousBackups`.
 pub(crate) async fn handle_update_continuous_backups(
     body: Value,
     ctx: &OperationContext,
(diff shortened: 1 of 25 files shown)


crates/engine/src/backup.rs
```diff
@@ -190,7 +190,81 @@ pub(crate) async fn handle_restore_table_from_backup(
...
+/// The continuous-backups description every table reports.
+///
+/// `ContinuousBackupsStatus` is always `ENABLED`, matching the service, which
+/// reports `ENABLED` unconditionally. `PointInTimeRecoveryStatus` is always
+/// `DISABLED` because no storage backend implements point-in-time recovery,
+/// and the restorable-time fields are omitted, as the service omits them for
+/// a table whose recovery is disabled. Owned by the engine rather than the
+/// backends: the answer does not depend on stored state, so no backend is
+/// consulted and all three report identically.
+fn disabled_continuous_backups_description() -> extenddb_core::types::ContinuousBackupsDescription {
```
jcshepherd, line 228: Confusing name. default_continuous_backups_description?

crates/storage-postgres/src/migrations.rs
```diff
@@ -335,6 +359,19 @@ mod tests {
...
 
+    /// The convergence write must target the cell the startup gate reads.
+    ///
+    /// The runner converges the stored version on the compiled constant after
+    /// every walk, so a replay of an earlier file cannot leave the version
+    /// behind the schema actually present. That only holds while this statement
+    /// writes the same settings row the files write and the gate reads, so the
+    /// statement text is pinned against a rename of the key or the table.
+    #[test]
+    fn the_runner_convergence_write_targets_the_catalog_version_cell() {
```
jcshepherd, line 370: :-/  I'm not sure this test is worth the CPU cycles needed to compile it.

docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
+# Design Spec: Off-Host Backup and Restore for ExtendDB
```
jcshepherd, line 1: I'd prefer this be a separate PR.

docs/design/README.md
```diff
@@ -17,6 +17,7 @@ This directory contains the authoritative design documents for ExtendDB (extendd
 | 09 | [Testing](09-testing.md) | Test strategy, reference suites, golden files, multi-language test suites, coverage tracking |
```
jcshepherd, line 17: Same: separate PR.

docs/manuals/01-architecture-guide.md
```diff
@@ -172,7 +172,7 @@ extenddb uses a dual-database architecture:
 - **Catalog database** (e.g., `extenddb_catalog`): Stores table metadata, account/user/group/role/policy definitions, access keys, settings, stream metadata, and metrics. Shared across all accounts.
 - **Data database** (e.g., `extenddb_catalog_data`): Stores user items, GSI/LSI data, and stream records. Each table gets its own PostgreSQL table.
 
-The catalog version is 0.0.3 on PostgreSQL and SQLite, stored in the `settings` table under the key `catalog_version` and checked at startup.
+The catalog version is 0.0.4 on PostgreSQL and SQLite, stored in the `settings` table under the key `catalog_version` and checked at startup.
```
jcshepherd, line 175: Can we remove the explicit version here? "The catalog version is stored in the `settings` table under the key `catalog_version` and checked at startup." Reduce the number of places where the version has to be updated.

docs/manuals/07-upgrade-manual.md
```diff
@@ -4,9 +4,9 @@
 
 ## Current Status
 
-Catalog 0.0.3 is current. The 0.0.2 to 0.0.3 upgrade is the first in-place catalog upgrade ExtendDB has, and **every existing PostgreSQL deployment must run it**, including deployments that never use vector indexes: the server refuses to start against a catalog version it was not built for.
+Catalog 0.0.4 is current. **Every existing PostgreSQL and SQLite deployment must run the upgrade**: the server refuses to start against a catalog version it was not built for.
```
jcshepherd, line 7: Same. Our tooling should say what the current and installed versions are, not the docs.

docs/manuals/07-upgrade-manual.md
```diff
@@ -15,8 +15,9 @@ See [Catalog 0.0.3](#catalog-003-current) below for what changes and the exact s
 Migrations are SQL files in `crates/storage-postgres/migrations/`, applied in filename order:
 
 ```
-001_schema.sql            ← the complete initial schema
-002_vector_indexes.sql    ← vector index metadata, catalog 0.0.3
+001_schema.sql                    ← the complete initial schema
```
jcshepherd, line 18: Prefer to just let folks look in the migrations files, and let the files document their purpose rather than repeating it here.

jcshepherd (commented):
A few comments. I realize my complaining about having to update the catalog version in 10 different files is whining in the age of AI ... nonetheless, I'm not sure it's helping to expect to have real up-to-date catalog version numbers sprinkled through documentation. Teach a user to fish ...

crates/engine/src/backup.rs
LeeroyHannigan: Renamed to `default_continuous_backups_description` and updated the call sites.

crates/storage-postgres/src/migrations.rs
LeeroyHannigan: Removed

docs/design/15-backup-restore.md
LeeroyHannigan: Done

docs/design/README.md
LeeroyHannigan: Done

docs/manuals/01-architecture-guide.md
LeeroyHannigan: Done

docs/manuals/07-upgrade-manual.md
LeeroyHannigan: Done

docs/manuals/07-upgrade-manual.md
LeeroyHannigan: Done

PR #336, opened 2026-09-10
feat(backup): add backup store crate with filesystem and S3 implementations

LeeroyHannigan:
## What

Adds the backup store crate, `crates/backup-store` (`extenddb-backup-store`): the `BackupStore` trait the engine will write portable backups through, with a filesystem implementation and an S3 implementation. This is the store layer only; nothing is wired into the server or `crates/config` yet, and no API behavior changes.

- `BackupStore` trait: `put` (streaming body), `get` (streaming body), `head` returning `Option<ObjectMeta { size, last_modified }>`, `list(prefix)` streaming `ObjectMeta` in key order, `delete_prefix(prefix)` returning the deleted count, and `validate()` for startup checks. Every operation returns a `StoreError` (`NotFound`, `PermissionDenied`, `InvalidKey`, `Io`/`Transport` with sources, `Other`) whose display output never carries credentials.
- Shared key validation: keys are `/`-separated components, each non-empty, no `.` or `..`, no backslashes, no control characters, at most 1024 bytes total. Prefix matching is component-aligned, so `a/b` never matches `a/bc`; the S3 store filters listings client-side to the same rule.
- `FilesystemStore`: canonicalizes its root at construction and resolves every key one component at a time without following symlinks: a symlink anywhere in a key's path, pointing inside or outside the root, is refused, so a key never resolves through a link and put, get, head, list, and delete_prefix agree on what a key names. These checks defeat symlinks present when an operation starts; they do not defend against another local process mutating the tree during an operation, so the backup root must not be writable by less-trusted users. Because `create_dir_all` follows symlinks in existing components, `put` walks to the deepest existing ancestor of its destination, refusing any symlinked component before creating a directory, then creates the remaining components one at a time with `create_dir`, re-verifying each is a real directory and not a symlink. Writes go to a temporary sibling that is synced before the rename, and the parent directory is synced after it, so an Ok put means the object and its directory entry are on disk; a crash never leaves a truncated object under its final name, and temporary names are invisible to listings. Prefix deletion removes files and the directories the deletion emptied. Refusals name the offending path, never a link's target.
- `S3Store` on the official `aws-sdk-s3` crate: credentials from the SDK default chain, with `bucket`, `prefix`, `region`, optional `endpoint`, `force_path_style`, `upload_part_size_bytes` (default 8 MiB, enforced 5 MiB floor), and `max_concurrent_uploads` (default 4). Bodies above the part size upload as multipart with bounded concurrency; every failure after CreateMultipartUpload reaches the abort, a failed CompleteMultipartUpload included, so no orphaned parts accrue storage. Listing paginates; prefix deletion batches DeleteObjects requests of 1000. `validate()` calls HeadBucket and names the bucket in the failure message. The failure-injection interceptor seam the test suite uses is gated behind a `test-util` cargo feature and absent from the default API surface.
- `BackupStoreConfig` with serde `deny_unknown_fields`, modeling the store subset of the `[backup]` section (`store = "filesystem" | "s3"`, `[backup.filesystem] path`, `[backup.s3]` fields), plus `open(config) -> Result<Arc<dyn BackupStore>>`.

Dependencies: `aws-config = "=1.8.13"` and `aws-sdk-s3 = "=1.122.0"`, pinned exactly to the newest versions whose minimum supported rustc matches the workspace `rust-version` of 1.88 (later releases require a newer compiler and fail the 1.88.0 CI job); transitive `aws-smithy-*` crates are held in `Cargo.lock` to 1.88-compatible versions for the same reason, and `crc` moves from 3.4.0 to 3.3.0 so sqlx and the checksum path share one crc major. The license notices files are regenerated for the crc version change and all three checks pass.

## Why

Every backup today shares the failure domain of its data, backups cannot cross backends, and a lost catalog destroys the ability to restore. The backup and restore design (`docs/design/15-backup-restore.md`) closes that with one engine-owned path writing a portable format to a configured store. This PR is the store abstraction that path writes through (task T1.1); the manifest types, snapshot engine, and job workers build on it in the following PRs.

## Testing done

- `cargo test -p extenddb-backup-store`: 52 tests, 0 filtered. A shared test module runs identical checks against both stores: put/get/head round trip, overwrite, empty body, list with and without prefix (including trailing slash and the `a/b` vs `a/bc` boundary), delete_prefix counts and emptiness after, NotFound on missing get and None on missing head, every rejected key shape on every operation, a body larger than the part size verified byte-for-byte, and a put whose body fails midway leaving nothing under the key. Filesystem-only: traversal and symlink escapes refused, including multi-level keys under planted directory and file symlinks with the outside filesystem asserted untouched (no directory may be created outside the root); an in-root directory symlink refused on put, get, and head, with list of the aliased prefix empty, delete_prefix on it returning 0 and removing nothing, and the object under its real key surviving; an in-root file symlink refused on get, head, and put with the link target untouched; an ancestor collision with an existing object; no stray temporary file after a failed put; and the validate() probe on an unwritable root.
- The full suite also ran live against a local MinIO on both toolchains (stable and 1.88.0), 52 of 52 with zero skips: the multipart abort test checks ListMultipartUploads is empty afterward, and a complete-time failure injected through a client interceptor verifies the abort also runs when CompleteMultipartUpload itself fails. Without `EXTENDDB_TEST_S3_ENDPOINT` the S3 tests announce the skip on raw stderr, visible in default output; the crate README documents the docker and plain-binary MinIO invocations. CI wiring for a MinIO service is task T1.9.
- `cargo test --workspace`: 1233 passed, 0 failed.
- Both toolchains checked: `cargo +1.88.0 check --workspace --locked` and `cargo +1.88.0 test -p extenddb-backup-store` pass on the 1.88.0 toolchain (the CI minimum), and all gates below pass on stable.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings`: clean.
- `devtools/generate-software-license-notices --check`, `devtools/generate-software-license-notices --check --backend mongodb`, and `devtools/generate-dev-license-notices --check`: all three current after regeneration.

## Checklist

ADR / RFC: docs/design/15-backup-restore.md (landing in PR 1)

---

### DIFF
--- a/crates/backup-store/src/fs.rs
+++ b/crates/backup-store/src/fs.rs
@@ -1,0 +1,525 @@
+// Copyright 2026 ExtendDB contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Filesystem-backed store.
+//!
+//! Keys map to paths under a root directory that is canonicalized at
+//! construction. Every operation re-resolves its target and refuses to act
+//! when the resolved path escapes the root, so a symlink planted inside the
+//! root cannot redirect reads or writes elsewhere. Writes go to a temporary
+//! sibling file and rename into place, so a crash never leaves a truncated
+//! file under its final name.
+
+use std::path::{Path, PathBuf};
+
+use futures::future::BoxFuture;
+use futures::stream::BoxStream;
+use futures::{FutureExt, StreamExt, TryStreamExt};
+use tokio::io::AsyncWriteExt;
+
+use crate::key::{key_matches_prefix, normalize_prefix, validate_key, validate_prefix};
+use crate::{BackupStore, ByteStream, ObjectMeta, StoreError};
+
+/// Prefix and suffix marking in-flight temporary files, hidden from listings.
+const TEMP_PREFIX: &str = ".put-";
+const TEMP_SUFFIX: &str = ".tmp";
+
+/// Store writing objects under a local root directory.
+pub struct FilesystemStore {
+    /// Canonicalized root; every resolved target must stay under it.
+    root: PathBuf,
+}
+
+impl FilesystemStore {
+    /// Open a store rooted at `root`. The directory must already exist; it is
+    /// canonicalized here and the canonical form anchors every escape check.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when the root does not exist, cannot be resolved, or
+    /// is not a directory.
+    pub async fn open(root: impl AsRef<Path>) -> Result<Self, StoreError> {
+        let root = root.as_ref();
+        let canonical = tokio::fs::canonicalize(root).await.map_err(|e| {
+            StoreError::io(format!("backup root {} is not usable", root.display()), e)
+        })?;
+        let meta = tokio::fs::metadata(&canonical)
+            .await
+            .map_err(|e| StoreError::io(format!("stat {}", canonical.display()), e))?;
+        if !meta.is_dir() {
+            return Err(StoreError::Other(format!(
+                "backup root {} is not a directory",
+                canonical.display()
+            )));
+        }
+        Ok(Self { root: canonical })
+    }
+
+    /// Map a validated key to its path under the root. Key validation has
+    /// already refused `..`, empty components, and separators, so the join
+    /// cannot step outside the root without a symlink, which the resolve
+    /// checks catch.
+    fn key_path(&self, key: &str) -> PathBuf {
+        let mut path = self.root.clone();
+        for component in key.split('/') {
+            path.push(component);
+        }
+        path
+    }
+
+    /// Canonicalize `path` and refuse it when it resolves outside the root.
+    /// Missing paths surface as `NotFound`.
+    async fn resolve_within_root(&self, path: &Path) -> Result<PathBuf, StoreError> {
+        let canonical = tokio::fs::canonicalize(path).await.map_err(|e| {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                StoreError::NotFound
+            } else {
+                StoreError::io(format!("resolve {}", path.display()), e)
+            }
+        })?;
+        if canonical.starts_with(&self.root) {
+            Ok(canonical)
+        } else {
+            Err(StoreError::PermissionDenied(format!(
+                "{} resolves outside the backup root",
+                path.display()
+            )))
+        }
+    }
+
+    /// Prepare the directory a key's object will be written into, without
+    /// ever creating a directory outside the root.
+    ///
+    /// `create_dir_all` follows symlinks in existing path components, so
+    /// checking the parent only after creating it would already have created
+    /// directories on the far side of a planted symlink. Instead: walk to the
+    /// deepest existing ancestor of the destination, canonicalize it, and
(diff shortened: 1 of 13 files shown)


crates/backup-store/src/s3.rs
```diff
@@ -0,0 +1,596 @@
...
+            }
+            match next {
+                Some(next) => token = Some(next),
+                None => break,
+            }
+        }
+        Ok(deleted)
+    }
+
+    async fn validate_impl(&self) -> Result<(), StoreError> {
```
robinnsc, line 404: The filesystem store's `validate()` writes and removes a probe file, but this only calls HeadBucket, which succeeds with list-only permission. The trait doc says validate exists so a misconfigured store fails at startup rather than at the first CreateBackup — a bucket the server can see but not write to defeats exactly that. A probe put/delete under the configured prefix would make the two stores equivalent.

Related smaller thing that could fold in: `head`/`get` map a missing bucket (NoSuchBucket → 404) to NotFound/None, so after a bucket deletion reads look like "no such object" rather than "store is broken". Acceptable if validate is strengthened; worth a doc sentence either way.

robinnsc (commented):
Three comments inline, none blocking, plus one Cargo.lock question here since the lock diff is unwieldy to anchor: the diff changes `winapi-util` from `windows-sys 0.61.2` to `windows-sys 0.48.0` — a downgrade unrelated to the AWS additions, which looks like accidental churn from the `--precise` pinning of the smithy crates. Intentional? If not, a targeted `cargo update` should put it back.

crates/backup-store/src/fs.rs
```diff
@@ -0,0 +1,544 @@
...
+            return Ok(None);
+        }
+        Ok(Some(ObjectMeta {
+            key: key.to_owned(),
+            size: meta.len(),
+            last_modified: meta.modified().ok(),
+        }))
+    }
+
+    async fn list_impl(&self, prefix: &str) -> Result<Vec<ObjectMeta>, StoreError> {
```
robinnsc, line 290: `collect_objects` walks the whole prefix into a `Vec` inside `spawn_blocking` before wrapping it in a stream, while the S3 store streams page by page. For the backup layout (bounded per-backup prefixes) this is fine in practice, but the trait presents a stream, and a caller listing the root of a large store gets surprise memory behavior. Fine to keep as is — just worth a sentence in the module doc, or a follow-up note.

crates/backup-store/src/s3.rs
```diff
@@ -0,0 +1,596 @@
...
+                );
+            }
+            return Err(e);
+        }
+        Ok(())
+    }
+
+    /// Upload every part with bounded concurrency; returns the completed part
+    /// list sorted by part number.
+    async fn upload_parts(
```
robinnsc, line 212: With the default 8 MiB part size, S3's 10,000-part limit caps an object at about 78 GiB; past that the upload fails mid-flight with a service error rather than anything actionable. A cheap client-side check when the part number passes 10,000 (or a doc note telling callers to raise `upload_part_size_bytes` for huge objects) would turn that into a clear error.

robinnsc (approved):

PR #337, opened 2026-09-10
feat(core): add portable backup manifest and DynamoDB export layout types

LeeroyHannigan:
## What

Adds `crates/core/src/backup_format/`, the pure-type layer of the portable backup format: types, serde, and checksum helpers with no I/O and no storage dependency.

- `ExtenddbManifest` (`extenddb-manifest.json`): the backup commit record. `format_version` (current 1) gated by an explicit `check_supported()`; backup identity with `backup_creation_date_time` as f64 epoch seconds, the representation `BackupDetails` uses; a snapshot descriptor (`kind`, `marker`); a `source_table` section built from the existing wire types (`KeySchemaElement`, `AttributeDefinition`, `BillingMode`, `ProvisionedThroughput`, `OnDemandThroughput`, `SseDescription`, `GsiInput`, `LsiInput`, `VectorIndexDescription`); per-data-file entries (`key`, `item_count`, `sha256`, `bytes`); and `manifest_sha256`, computed over the manifest's compact JSON with the checksum field set to the empty string, with compute, seal, and verify helpers. Unknown fields are tolerated on read for forward compatibility; the version gate is the refusal point.
- DynamoDB export layout: `ExportSummaryManifest` and `ExportFileEntry` model `manifest-summary.json` and `manifest-files.json` exactly as Amazon DynamoDB writes them for `ExportTableToPointInTime` with `DYNAMODB_JSON` output, per the documented export output format (cited in the module doc). That includes the explicit `null` for `s3SseKmsKeyId`, `exportType` appearing only on incremental exports, and JSON-lines serialization for the files manifest. Serializer and parser both provided; the parser also accepts the documented incremental-export members it does not model.
- Data files: `encode_data_file` takes an iterator of `Item` and produces gzip bytes plus running item count, sha256 (hex) and md5 (base64) over the compressed bytes, and the uncompressed byte count. `DataFileDecoder` streams items back and verifies whichever expectations it is given. The line encoding is the same `json!({"Item": item})` wrapper the engine's export handler writes, so the two agree byte for byte.
- Key layout: `backup_prefix(account_id, table_name, backup_id)`, `data_file_key` (`data/000001.json.gz`), the fixed object names, and `parse_backup_arn` for the ARN shape the engine's backup handlers already use, unchanged.

Dependencies added to the workspace: `flate2` 1.1.9, `md-5` 0.10.6, `sha2` 0.10.9, the versions already resolved in `Cargo.lock`.

## Why

The backup and restore design replaces per-backend backup storage with one engine-owned portable format: the DynamoDB S3 export layout plus one ExtendDB manifest, so a backup restores into any backend, imports into Amazon DynamoDB with `ImportTable`, and a DynamoDB export restores into ExtendDB. This PR is the format layer that the store abstraction and the engine jobs build on.

## Testing done

- `cargo fmt --all -- --check` clean.
- `cargo clippy --all-targets -- -D warnings` clean.
- `cargo clippy -p extenddb-core -- -W clippy::pedantic`: the new `backup_format` module is clean; the remaining warnings are pre-existing elsewhere in the crate.
- `cargo test -p extenddb-core`: 546 passed, 54 of them in `backup_format`.
- `cargo test --workspace`: 1212 passed, 0 failed.
- `cargo +1.88.0 check --workspace --locked`: clean.
- Coverage includes: serde round trips for every type; `manifest_sha256` compute and verify with tamper detection; `check_supported` rejecting format_version 2; encoder and decoder round trips over items covering every attribute type (S, N at 38 significant digits plus negative and exponent forms, B, SS, NS, BS, M nested to depth 32, L, NULL, BOOL) with checksum and count tamper detection; a decompression-bound test that feeds a 64 MiB single-line gzip bomb and asserts the typed cap error fires with peak resident memory bounded (measured from `/proc/self/status` on Linux); multi-member gzip round trips; data-file key layout validation rejecting traversal, absolute, wrong-prefix, and extra-directory shapes; ARN parse and prefix round trips with malformed-ARN rejection; and golden-file tests against a checked-in export under `crates/core/testdata/ddb-export/` asserting the parser reads it and the serializer reproduces the `manifest-files.json` lines byte for byte. The fixture is constructed to the documented output format (provenance in its README).

## Checklist

- [x] Clippy is clean (`cargo clippy --all-targets -- -D warnings`); the new `backup_format` module is additionally `clippy::pedantic`-clean

ADR / RFC: docs/design/15-backup-restore.md (landing in PR 1)

---


crates/core/src/backup_format/data_file.rs
```diff
@@ -0,0 +1,727 @@
...
+            let available = self
+                .reader
+                .fill_buf()
+                .map_err(|e| FormatError::Decode(e.to_string()))?;
+            if available.is_empty() {
+                return Ok(line.len());
+            }
+            if let Some(nl) = available.iter().position(|&b| b == b'\n') {
+                let want = nl + 1;
+                if line.len() + want > MAX_LINE_BYTES {
```
robinnsc, line 289: Off-by-one against the doc: the comment on `MAX_LINE_BYTES` says "a line at or over the cap yields `LineTooLong`", but this check is `>`, so a line of exactly `MAX_LINE_BYTES` is accepted. Related: `line_exactly_at_cap_boundary_is_rejected` writes 1 MiB blocks while `written <= MAX_LINE_BYTES`, which ends at 5 MiB, so despite the name it tests one block past the cap rather than the boundary. Either behavior is defensible, but the doc, the check, and the test should agree, and the boundary test should feed exactly `MAX_LINE_BYTES` bytes.

crates/core/src/backup_format/manifest.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// How the consistent snapshot was taken.
+    pub snapshot: SnapshotDescriptor,
+    /// The source table definition.
+    pub source_table: SourceTableSchema,
+    /// The data files that make up the backup.
+    pub data_files: Vec<DataFileEntry>,
+    /// Lowercase hex sha256 of the manifest's canonical JSON, computed with
+    /// this field set to the empty string. See
+    /// [`Self::compute_manifest_sha256`].
+    pub manifest_sha256: String,
```
robinnsc, line 240: Design question, worth settling before format_version 1 freezes: the manifest checksums itself and records a sha256 per data file, but nothing commits `manifest-summary.json` or `manifest-files.json`. Those two can be tampered with or corrupted without any ExtendDB-side verification failing, and the files manifest carries the md5s a plain ImportTable-style consumer would trust. Since this manifest is the commit record, recording the sha256 of both DynamoDB manifests here would make the whole backup verifiable from one sealed document. Adding it later costs a version bump, so now is the cheap time.

crates/core/src/backup_format/manifest.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// of whether it ran [`ExtenddbManifest::validate_data_files`]. The
+    /// entry's recorded `sha256` and `item_count` are enforced against the
+    /// bytes.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FormatError::InvalidDataFileKey`] when the key is not a
+    /// single data-file entry under the prefix, and any error
+    /// [`crate::backup_format::decode_data_file`] can produce.
+    pub fn decode(&self, compressed: &[u8]) -> Result<Vec<crate::types::Item>, FormatError> {
```
robinnsc, line 167: The entry records `bytes` (compressed size) but `decode` never enforces it, only sha256 and item_count. A length check is nearly free and fails fast before hashing a wrong-sized object, so it seems worth passing through as an expectation (or checking directly here).

crates/core/src/backup_format/manifest.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// Callers run this when a manifest is parsed, and the decode path runs
+    /// the same per-entry [`DataFileEntry::validate_key`] again before it
+    /// resolves a key, so a key that slips past one check is caught at the
+    /// other.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FormatError::InvalidDataFileKey`] for the first offending
+    /// key.
+    pub fn validate_data_files(&self) -> Result<(), FormatError> {
```
robinnsc, line 272: This doesn't reject duplicate keys, so a manifest listing the same data file twice validates cleanly and would restore its items twice. A uniqueness check over `data_files` here would close that.

Cargo.toml
```diff
@@ -74,6 +74,10 @@ dashmap = "6"
 
 # Crypto & checksums
 crc32fast = "1"
+# Pinned to the versions already resolved in Cargo.lock.
```
robinnsc, line 77: The comment says "pinned", but these are caret requirements, so any compatible 1.x can resolve on a lockfile refresh. Either use `=` versions if pinning is the intent, or match the workspace's existing loose style (`crc32fast = "1"`) and drop the comment.

crates/core/src/backup_format/layout.rs
```diff
@@ -0,0 +1,173 @@
...
+/// The store key prefix for one backup: `<account_id>/<table_name>/<backup_id>/`.
+#[must_use]
+pub fn backup_prefix(account_id: &str, table_name: &str, backup_id: &str) -> String {
+    format!("{account_id}/{table_name}/{backup_id}/")
+}
+
+/// The key of the `index`-th data file within a backup prefix:
+/// `data/000001.json.gz` for index 1.
+#[must_use]
+pub fn data_file_key(index: u32) -> String {
```
robinnsc, line 35: Take or leave: zero-padding to six digits means file 1,000,000 breaks lexicographic ordering (the test even shows `data/1000000.json.gz`). If anything ever lists the data prefix and relies on name order matching index order, that's a trap. If order never matters, a one-line doc note saying so would close the question.

robinnsc (commented):
Read through the whole module. Overall this is in good shape: the bounded line reader, MultiGzDecoder, the key validation, and the golden fixtures are all solid, and the checksum canonicalization tradeoffs are documented rather than hidden. One real bug (the line cap boundary), one design question I'd like settled before format_version 1 freezes (committing the two DynamoDB manifests), and a few smaller things inline.

PR #338, opened 2026-09-10
Feat/storage cassandra in tree

jcshepherd:
## What
This is a full-featured Cassandra backend for ExtendDB. The only current ExtendDB feature that it does not support is vector indexes.

## Why

Refer to docs/rfcs/draft-cassandra-backend.md (in the PR) for background and design notes. Detailed design notes on specific aspects are available in the ADRs including in this PR.

Closes #

## Testing done

pytests, rust-integration tests, unit tests all pass. Clippy is a work in progress.

## Checklist

- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [X] All tests pass (`cargo test --workspace`)
- [X] Code is formatted (`cargo fmt --check`)
- [X] I have added or updated tests for new functionality
- [X] I have updated documentation if behavior changed
- [X] Breaking changes are noted below (if any)
- [X] If this changes the wire protocol, `Storage` trait, auth model, on-disk

ADR / RFC: docs/rfcs/draft-cassandra-backend.md 

## Breaking changes

---


amrith (commented):
Would you take a look at the failing checks - or are they related to 341?

crates/storage-cassandra/src/data/put_get_item.rs
```diff
@@ -0,0 +1,877 @@
...
+                    cap,
+                    &self.hlc,
+                    self.stream_retention_seconds,
+                )
+            });
+
+            if indexes.is_empty() && stream_stmt.is_none() && ttl_config.is_none() {
+                // Fast path: no batch needed.
+                let insert_query = format!(
+                    "INSERT INTO {data_keyspace}.{ddb_table} \
```
LeeroyHannigan: Conditional PutItem is not atomic. The row is read at 148 and the condition evaluated in Rust at 189, then this INSERT carries no IF and no version check, so two writers on the same key both pass and both write. Probe with `attribute_not_exists(guard)` on a non-key attribute: both succeeded in 30 of 30 trials, postgres 0 of 60. Needs the OCC treatment UpdateItem has (read version, write with IF version = ?, retry).

crates/storage-cassandra/src/data/put_get_item.rs
```diff
@@ -0,0 +1,877 @@
...
+}
+
+/// Returns `true` if `condition` is exactly `attribute_not_exists(<key_attr>)` where
+/// `<key_attr>` resolves to one of the names in `key_attr_names` after applying
+/// expression name substitutions from `maps`.
+///
+/// This is the only condition we map directly to a null-aware LWT without a
+/// prior read.
+pub(crate) fn is_attribute_not_exists_key(
+    condition: Option<&Expr>,
```
LeeroyHannigan, line 664: This never receives the expression name map, so `attribute_not_exists(#p)` with `{"#p": "pk"}` compares the literal `#p` against the key names, misses the LWT fast path, and falls into the read-then-write path above. Same request, two spellings, different atomicity: alias form both writers succeeded 20 of 20, plain form 0 of 20. The doc comment states this resolution happens; resolve `maps` before the comparison.

crates/storage-cassandra/src/data/put_get_item.rs
```diff
@@ -0,0 +1,877 @@
...
+                    cap,
+                    &self.hlc,
+                    self.stream_retention_seconds,
+                )
+            });
+
+            if indexes.is_empty() && stream_stmt.is_none() && ttl_config.is_none() {
+                // Fast path: no batch needed.
+                let insert_query = format!(
+                    "INSERT INTO {data_keyspace}.{ddb_table} \
```
LeeroyHannigan: This write leaves the `version` cell untouched, so a concurrent UpdateItem's `IF version = n` still applies and overwrites the put with a post-image computed from the pre-put row. Probe with PutItem racing `SET u = 1`: the put was lost in 17 of 20 trials (pre-put attributes present, put attributes gone) with both calls reporting success. Every write path needs to advance or clear `version` under the same protocol.

crates/storage-cassandra/src/data/delete_item.rs
```diff
@@ -0,0 +1,1375 @@
...
+            };
+            // Pinned here, immediately after the claim and before any further
+            // await, so it is strictly newer than the image this request owns
+            // yet strictly older than anything a later owner commits. A request
+            // suspended past its claim lifetime therefore loses to that later
+            // owner instead of overwriting it.
+            let mutation_timestamp = chrono::Utc::now().timestamp_micros();
+
+            // Delete the item (with index updates if needed).
+            let delete_cql = format!("DELETE FROM {data_keyspace}.{ddb_table} WHERE pk = ?");
```
LeeroyHannigan, line 587: The condition is evaluated at 469, then this DELETE has no IF and no version check. Probe: seed `{pk, v=1}`, DeleteItem `IF v = 1` racing `UpdateItem SET v = 2`: both calls succeeded in 30 of 30 trials and the item was gone in 16 of 30, which no serial order allows. Needs `DELETE ... IF version = ?` against the version read, with retry.

crates/storage-cassandra/src/data/transactions.rs
```diff
@@ -0,0 +1,1881 @@
...
+        let account_keyspace = self.account_keyspace(first_account);
+
+        // Check the account-scoped idempotency token before creating the ledger.
+        if let Some((idempotency_account, token, fingerprint)) = idempotency {
+            if idempotency_account != first_account {
+                return Err(StorageError::Validation(
+                    "Idempotency account does not match transaction account".to_owned(),
+                ));
+            }
+            self.check_idempotency_token(
```
LeeroyHannigan, line 248: The token is reserved durably at 861 before the ledger entry exists, and nothing deletes it when the transaction is cancelled. The retry then matches the stored fingerprint at 893 and returns IdempotentReplay, which the engine reports as a successful response with no writes. Executed: transaction cancelled by a failing condition, blocking item removed, identical call with the same token returned ok and wrote nothing. The SDKs generate this token and retry TransactWriteItems automatically. Reserve it inside the ledger commit, or release it on every cancellation path.

crates/storage-cassandra/src/update_table.rs
```diff
@@ -0,0 +1,552 @@
...
+                        .transpose()
+                        .map_err(|e| StorageError::Internal(e.to_string()))?
+                        .unwrap_or_default();
+
+                    batch = batch.add_query(
+                        format!(
+                            "INSERT INTO {catalog_ks}.indexes \
+                             (table_id, index_name, index_id, index_type, key_schema, \
+                              projection, index_status, provisioned_throughput) \
+                             VALUES (?, ?, ?, 'GSI', ?, ?, 'ACTIVE', ?)"
```
LeeroyHannigan, line 285: The index row commits as ACTIVE here, before the index data table exists (DDL runs after the batch) and before the backfill at 514. With one concurrent writer on a 10,000 item table: UpdateTable returned InternalServerError, DescribeTable reported ACTIVE immediately and never CREATING, the index held 7 items at first look and 35 of 10,200 four minutes later with no repair path (retry gets IndexAlreadyExists, the gsi_pending workers only apply post-creation writes), and 317 of 517 concurrent PutItems failed against the index table that did not exist yet. On a quiet table the same operation completes with 10,000 of 10,000. Write CREATING, fence writers until the table exists, backfill asynchronously, and reach ACTIVE only when the backfill commits.

crates/engine/src/streams.rs
```diff
@@ -111,13 +111,15 @@ pub async fn handle_get_shard_iterator(
...
             })?;
-            let n = raw.parse::<u64>().map_err(|_| {
+            let n = raw.parse::<u128>().map_err(|_| {
                 DynamoDbError::ValidationException("Invalid SequenceNumber".to_owned())
             })?;
             // n == 0: sequence 0 is the first possible record, so "at 0"
             // means "read from the beginning" — same as TRIM_HORIZON.
             if n > 0 {
-                format!("{:021}", n - 1)
+                // Pad to the same width as the input so lexicographic order
```
LeeroyHannigan: This changes shared engine behavior and breaks AT_SEQUENCE_NUMBER on postgres, sqlite, and mongodb. They store sequence numbers 21 digit zero padded and compare as text, so width-preserving padding turns an unpadded `5` into `4`, which sorts above every stored value: the iterator returns nothing where it previously returned records from sequence 5. It also does not fix the Cassandra case (an unpadded input against 23 digit HLC sequences fails the same way). This belongs in its own PR with regression tests for unpadded input on every backend.

crates/storage-cassandra/src/engine.rs
```diff
@@ -0,0 +1,261 @@
...
+            session: Arc::new(session),
+            region: region.to_string(),
+            keyspace_prefix: config.keyspace_prefix.clone(),
+            replication_factor: config.replication_factor,
+            datacenter: config.datacenter.clone(),
+            control_plane_notify: Arc::new(tokio::sync::Notify::new()),
+            gsi_default_delay_ms: Arc::new(std::sync::atomic::AtomicU64::new(10)), // Default 10ms
+            gsi_queue: crate::gsi_queue::GsiQueue::new(),
+            hlc: crate::stream_util::new_shared_hlc(
+                config.instance_id.as_deref().unwrap_or("default"),
```
LeeroyHannigan, line 124: `instance_id` is always None here, so every node falls back to "default" and derives the same HLC node id (startup logs `node_id=137 instance_id="default"`). `set_instance_id` has no caller anywhere in the tree. `stream_records` keys on (shard_id, sequence_number), so two nodes in the same millisecond with the same counter and shard overwrite each other's record. ADR 0018's uniqueness argument depends on this being wired.

crates/storage/src/config.rs
```diff
@@ -23,6 +23,19 @@ pub trait StorageConfig: Send + Sync + std::fmt::Debug {
     /// Clone this config into a boxed trait object.
     fn clone_box(&self) -> Box<dyn StorageConfig>;
 
+    /// Set the server instance identifier for this backend.
+    ///
+    /// Called by the server before constructing the storage engine, passing
+    /// a stable unique identifier for this server process (typically the
+    /// bind address, e.g. "192.168.1.1:18443"). Backends that need per-instance
```
LeeroyHannigan: Added to the shared trait for one backend's need and never called: `grep -rn set_instance_id crates/` finds only the definitions, this default, the Cassandra impl, and the wrapper in extenddb-config. Wire it in the serve path or derive the node id from something already unique.

crates/storage-cassandra/src/bootstrapper.rs
```diff
@@ -0,0 +1,704 @@
...
+
+        // CLI args override config (or use config values if no CLI arg provided)
+        let resolved_contact_points = cassandra_contact_points.map_or(contact_points, |cp| {
+            cp.split(',')
+                .map(std::string::ToString::to_string)
+                .collect()
+        });
+        let resolved_admin_user = cassandra_user
+            .unwrap_or_else(|| std::env::var("USER").unwrap_or_else(|_| "cassandra".to_owned()));
+        let resolved_keyspace_prefix = keyspace_prefix.unwrap_or(prefix);
```
LeeroyHannigan, line 190: Re-running init against a cluster where the `extenddb` role already exists produces a config the server cannot start with. Executed with `--keyspace-prefix r_probe --cassandra-user cassandra`: init created `r_probe_catalog` and `r_probe_account_<id>`, printed "User 'extenddb' already exists", then wrote `keyspace_prefix = "extenddb"` and no username or password; serve failed with "Connection timeout after 10s". Persist the resolved prefix, and either fail loudly or rotate the password when the role exists.

crates/storage-cassandra/tests/common/mod.rs
```diff
@@ -0,0 +1,778 @@
...
+use extenddb_storage::TableEngine;
+use extenddb_storage_cassandra::{
+    CassandraCatalogStore, CassandraEngine, CassandraSession, CassandraStorageConfig,
+};
+use std::sync::Arc;
+
+/// Returns a standard test configuration for Cassandra.
+pub fn test_config() -> CassandraStorageConfig {
+    let mut config = CassandraStorageConfig {
+        contact_points: vec!["127.0.0.1:9042".to_string()],
```
LeeroyHannigan, line 20: Hardcoded contact point with no environment override and no skip gate, so the 25 tests in this crate's integration files fail `cargo test --workspace` unless a Cassandra node happens to be on 9042. That is the red `test` job and the first checklist item in the PR template. The other backends keep live service tests out of the workspace suite or gate them.

tests/test_transaction_operations.py
```diff
@@ -326,6 +326,26 @@ def test_transact_write_conditional_put_fail(dynamodb_client, hash_table):
     resp = dynamodb_client.get_item(TableName=hash_table, Key={"pk": {"S": "cp-2"}})
     assert resp["Item"]["v"]["S"] == "old"
 
+def test_transact_write_conditional_put_fail_diag(dynamodb_client, hash_table):
```
LeeroyHannigan: This test prints and asserts nothing, so it passes on every backend regardless of behavior. Its printed branch ("NO EXCEPTION - transaction succeeded (wrong)") describes a conditional Put inside TransactWriteItems succeeding when it must fail, which is the same defect class as the put and delete comments above. Drop it from this PR; the asserting test directly above covers the case once the backend is fixed.

LeeroyHannigan (changes requested):
Requesting changes. Twelve blocking items are commented inline; the full analysis with the probes, the should-fix list, and a proposed acceptance path comes separately.

Four items are not code and have no line to sit on:

1. CI on this head is red on clippy (three `too_many_arguments` and one `from_str` naming in the crate), test, notices-current (Cargo.lock is stale under `--locked`), launcher-tests, CodeQL (account_id written to a log), and both MongoDB legs. The 1.88.0 leg was cancelled, so MSRV never completed.
2. No workflow in this PR runs anything against Cassandra, so none of the crate tests, the pytest suite, or the Rust integration suite cover this backend in CI. RFC-0002 requires conformance results tracked in CI, and this backend implements every trait, so that obligation is total. The workflow in #339 targets this branch rather than main.

For context on what holds up: 111 differential write probes and every read probe matched the postgres backend exactly, the shared pytest suite passes 860 tests against Cassandra, the tenant boundary held on every cross-account probe, and throughput at 32 clients is within roughly 10 percent of postgres on writes. The defects are concentrated in the write paths that depend on what they read. UpdateItem already has a sound protocol; applying it to conditional PutItem, conditional DeleteItem, PutItem versioning, and the transaction token would close five of the twelve.

PR #339, opened 2026-09-11
feat(cassandra): TTL production-readiness follow-ups

robinnsc:
## What

Follow-up work to take the merged Cassandra TTL implementation to production readiness, in the order we agreed after #330: functionality first, then performance. One squashed commit, four areas.

**A self-healing audit.** A queue registration is only ever created by a write or by the enable backfill. That means any registration lost through a path the design didn't anticipate is permanent for an item that's never written again, and nothing alarms — this was the failure shape behind most of the bugs found during the #330 review rounds. A new worker rescans every ready TTL table every six hours and re-registers whatever's missing. Healthy items cost one quorum read each, pages are paced at 100ms, and the pass stands aside if the table's TTL config moves mid-scan. Repairs are counted in a new `TtlAuditRepairedEntryCount` metric, which doubles as the drift signal: any nonzero value means some loss path fired and is worth investigating.

**Transaction recovery does full reconciliation.** PREPARE now captures each row's pre-commit image into the ledger, so COMMITTING recovery retires the queue entry the item had before the transaction as well as registering the committed one — the same transition the live commit path performs, and it now covers DELETE ops too. Old ledgers without the field still parse (serde default) and just stay insert-only, which is the old behavior.

**Cassandra CI, and a green board.** A workflow runs the unit, metadata, and TTL integration suites against a `cassandra:4.1` service container. Until now none of these tests ran anywhere but a developer machine. The repair worker also emits `TtlRepairMarkerCount`, the unresolved-marker gauge ADR-0010 asks operators to watch.

Two smaller commits make the rest of the checks pass, since both failures predate this PR and hit every PR against this branch. The clippy job (`-D warnings`, workspace-wide) had nine pre-existing findings, fixed in dd4c563 with the narrowest changes that hold: targeted allows matching existing repo usage for the structural lints, and a mechanical `chunks_exact` → `as_chunks` in sqlite vector_search. And the bare workspace `cargo test` job has no services, so the Cassandra integration suites failed it on connection timeout; cc30965 has them probe 127.0.0.1:9042 and skip themselves when nothing is listening, following the postgres suites' skip-when-unconfigured pattern. Since integration.yml documents exactly how that pattern goes wrong (batch_transact_authz "reported ok WITHOUT EXECUTING" for months), the guard is hardened: the dedicated Cassandra workflow sets `EXTENDDB_TEST_CASSANDRA=required`, under which the suites never skip and an unreachable Cassandra fails the job loudly rather than producing a green run with zero tests executed. All three paths verified by hand (reachable → 26+1 run; absent → skip with message; absent+required → timeout failure). Both commits are separable if you'd rather take either on the base branch.

**Performance.** Three changes, each measured with a new benchmark harness (`tests/ttl_bench.rs`, ignored by default, run manually around changes):

- The sweep processes its batch concurrently (16 in flight) across the queue's shard partitions, with the lease renewed per 64-row wave. Batch goes from 100 to 1,000 per cycle. A 500-item backlog drains in one pass at ~156 items/sec of wall time, versus four passes at 65/sec before. The production ceiling moves from ~100 to ~1,000 expirations per table per minute.
- Per-table TTL config is served from a 5-second in-process cache, cutting a catalog read from every put/update/delete cluster-wide. Client latency is unchanged (that read was already overlapped with the index fetch) — this is load reduction, not a latency win. The background outbox pass also skips keyspaces of accounts with no TTL tables, widening every tenth cycle so leftovers still drain.
- TTL claims fence on the OCC `version` column instead of shipping the whole serialized item as an LWT condition value (the "unify the two concurrency mechanisms" suggestion from the #330 review). Every image write now bumps `version`: puts pre-read it and write old+1, transactions bump server-side. Bench numbers with 256-byte items: plain put p50 730µs, TTL put p50 ~1.46ms — the remaining tax is the claim LWT round itself.

## Why

The #330 review established the operating envelope but left the failure economics asymmetric: prevention had several layers, repair had none, and the observed bug class (silent non-expiration) is invisible without a repair pass to count what it fixes. The audit closes that. The rest is throughput and per-write overhead, which were the explicitly deferred items from the review — including two things Joel named directly (metadata caching, version-based fencing).

## Design decisions worth reviewing

- **The audit is deliberately leaseless.** An earlier draft held the table's control lease for the whole scan, which would have suspended expiration and lifecycle changes for hours on a large table — "the audit is rare" bounds how often that happens, not how long. The lease bought nothing: the audit's registrations are idempotent conditional inserts that defer to claimed work, so it's safe under concurrent sweeps and backfills, and a per-page quorum config check catches lifecycle movement. Two hosts auditing the same table just duplicate idempotent reads.
- **The exact base delete keeps all three conditions (owner AND version AND full image).** Version alone turned out not to be a unique image identity: a transaction's server-side `version + 1` and a stale writer's explicit `old + 1` can stamp two different images with the same version if app and coordinator clocks diverge past the claim lifetime, and version restarts across delete/recreate. Claims tolerate a mis-admitted fence (it only affects contention); the one statement that destroys data keeps the image condition and rests on no clock assumption. That payload is paid once per actual expiration, not per write, so the point of the change — nothing item-sized on the write path — holds.
- **CQL arithmetic quirk that shaped the transaction path:** `version = version + 1` is valid on regular columns in Cassandra 4.x, but `null + 1` stays null. Prepare's new-item INSERT initializes `version` to 0 so the commit yields 1; without that, transactional-only rows would stay unversioned forever. Legacy null-version rows fall back to the old `item_data` fence.
- **Sweep concurrency instead of per-shard leases.** Per-(table, shard) leases can't express the lifecycle exclusion the single lease provides (an LWT can't couple across rows), so the sweep keeps one lease and parallelizes under it. Same safety story, no protocol change — every queue transition was already conditional on its exact work id. The per-row pre-effects gate is a quorum config read, not a lease renewal, so 16 concurrent rows don't contend on one catalog partition's Paxos.
- **The config cache is only safe because the audit exists.** A stale `None` makes a write skip its queue mutation; a stale generation enqueues into a retired one. Both are now eventually repaired instead of silently permanent, which is what made a 5s cache acceptable. An invalidation epoch prevents a lifecycle change from being shadowed by a concurrent cache-miss repopulation, and `update_table`'s GSI-fence decision bypasses the cache at quorum. Index metadata is deliberately not cached — a stale-empty index list affects non-TTL correctness and belongs to the general metadata-cache work.

## Risks

- ~~The CI workflow has never executed — this PR's checks are its first run.~~ Resolved: `cassandra-storage` passed on its first execution.
- The skip guard means a developer running the integration suites locally without Cassandra gets a quiet pass (with an eprintln, visible under `--nocapture`). CI is immune via `required`, but locally it's on you to have the container up.
- The version fence changes LWT conditions on the write path. It's covered by direct claim refusal/admission probes plus an end-to-end changed-image recovery test, but it's the highest-blast-radius change here and deserves the closest read.
- Sweep batch and concurrency (1,000 / 16) raise Cassandra load per cycle on backlogged tables. Numbers are from a single-node bench; a real cluster under sustained load is untested (see follow-ups).
- The audit adds a paced full-table scan per TTL table every six hours. On very large tables that's a sustained background read load; the pacing constant is a first guess.
- The docs commit that landed on this branch mid-flight was rebased over; the conflict resolution kept both (their clippy fixes were re-applied where my rewrites had superseded them — clippy count matches the base exactly).

## Planned follow-ups (not in this PR)

- **`ENABLING`/`DISABLING` status + a checkpointed, non-blocking enable backfill.** These belong together: the status is only honest once the backfill runs in the background instead of blocking `UpdateTimeToLive`. Touches the shared `TimeToLiveStatus` enum, so all backends.
- **A real load test** (sustained writes + expiration backlog + node restart) before any production performance claim beyond this bench.
- **Retiring the `item_data` fallback** once no unversioned rows can exist, which removes the canonical-JSON coupling entirely.
- Async-GSI + TTL remains out of scope per ADR-0010; it's the one functional capability the PostgreSQL backend has that this one doesn't.

## Testing done

Real Cassandra 4.1, single node, `127.0.0.1:9042`, after rebasing onto the current branch head:

- `cargo test -p extenddb-storage-cassandra --test ttl_integration -- --test-threads=1` — **26 passed, 0 failed**
- `cargo test -p extenddb-storage-cassandra --test metadata_integration` — 1 passed
- `cargo test -p extenddb-storage-cassandra --lib` — 43 passed
- `cargo clippy -p extenddb-storage-cassandra --all-targets` — 4 warnings, byte-identical to the branch head before this PR
- `cargo fmt --all -- --check`, `git diff --check` — clean
- Bench (relative numbers, this machine only): drain 156 items/sec with all 500 deletions verified; put p50 plain 730µs / TTL 1.46ms
- Skip-guard paths exercised for real: with Cassandra up all 27 tests execute; with the container stopped both suites skip and report ok with a message; with the container stopped and `EXTENDDB_TEST_CASSANDRA=required` the suite fails on connection timeout in 10s
- Workspace `cargo clippy --all-targets -- -D warnings` clean

New tests: audit repairs a lost registration (and a second pass is a no-op), version-fence claim refusal and admission probes, end-to-end changed-image recovery under the version fence, and the earlier suite's 22 tests all still pass unmodified except where signatures changed.

Everything here went through adversarial review before commit; the two findings that survived it (version non-uniqueness, the audit's lease hold) are the first two design callouts above.

## Checklist

- [x] The affected Cassandra package builds and all applicable tests pass against real Cassandra
- [x] This branch introduces no new Clippy warnings
- [x] Tests cover the new behavior
- [x] Documentation (ADR-0010, differences-from-dynamodb) updated to match the code
- [x] No public wire-protocol, `Storage` trait, auth, or CLI breaking change is introduced


robinnsc:
Note on the second commit: the clippy job runs `-D warnings` workspace-wide and the base branch had nine pre-existing findings (server auth, sqlite vector_search, and four in storage-cassandra), so the job was red for any PR against it. dd4c563 fixes them with the narrowest changes that hold — targeted allows matching existing repo usage, plus one mechanical `chunks_exact` → `as_chunks` swap. It's split out so it can be dropped here and taken on the base branch instead if that's preferred. Also flagging: the `cassandra-storage` check in this PR is the new workflow's first-ever execution.

robinnsc:
On the two red checks: `test` / `run-tests` run the workspace suite with no Cassandra service, so the storage-cassandra integration tests fail on connection timeout. This is pre-existing — the base branch's own runs show the same jobs red (e.g. run 34534880070) — and the fix is a branch-level decision (gate the suites behind an env var, or give that job a Cassandra container like the dedicated `cassandra-storage` workflow, which is green here), so I left it out of this PR rather than deciding it unilaterally.

.github/workflows/integration-cassandra.yml
```diff
@@ -0,0 +1,62 @@
...
+  contents: read
+
+# `required` disables the suites' skip-when-unreachable guard (they otherwise
+# skip themselves in the bare workspace test job, which has no services): in
+# THIS workflow an unreachable Cassandra must fail the job, never report a
+# green run with zero tests executed.
+env:
+  EXTENDDB_TEST_CASSANDRA: required
+
+# The Cassandra backend's integration tests (crates/storage-cassandra/tests)
```
jcshepherd, line 21: Hey: question. These look similar to the "direct integration tests" in the original Cassandra plug-in repo. Would you agree? Do you think there is value in porting those tests over?

crates/storage-cassandra/src/ttl_worker.rs
```diff
@@ -1128,14 +1309,35 @@ pub async fn sweep_once(storage: &CassandraEngine, metrics: &MetricsCollector) {
...
+                            metrics.record_ttl_staleness(
+                                table_name,
+                                now_epoch.saturating_sub(expires_at) as f64,
+                            );
+                        }
+                        Ok(false) => {}
+                        // One row's failure must not abandon the wave or the
+                        // table: its durable queue state drives its own retry.
+                        Err(error) => {
+                            tracing::warn!("TTL worker: row failed in {table_name}: {error}");
```
jcshepherd, line 1339: This might be worth emitting a metric for, to give it a better chance of getting noticed if it's happening chronically.

jcshepherd (approved):
One minor suggestion below, but I don't see anything blocking here. Thanks!

crates/storage-cassandra/src/ttl_worker.rs
robinnsc: Good call — added in b918638 as `TtlSweepRowErrorCount`, dimensioned per table from the wave outcome loop. Isolated blips are expected (contention, restarts); the signal is a sustained rate on one table, which means rows are erroring and retrying without ever expiring.

.github/workflows/integration-cassandra.yml
robinnsc: Agreed, they're the same kind of test — real node, real suites, no mocks. I haven't gone through the plug-in repo's direct integration tests case by case, so I can't say how much overlap there already is with what's in-tree now, but porting whatever doesn't overlap seems clearly worth doing. I'd take it as a follow-up rather than growing this PR further — happy to pick it up next.

PR #347, opened 2026-09-14
fix(core): validate attribute name sizes inside nested maps (F-17)

ssbytedev:
Notice: I used AI to help me generate this summary.

## What

`validate_attribute_name_sizes` (`crates/core/src/validation/mod.rs`) now recurses into `M` and `L` values via a new `check_nested_attribute_name_sizes` helper, applying the existing per-attribute-name byte limit (`LimitsConfig::max_attribute_name_bytes`) to map keys at every nesting depth, not just the top level.

Adds five unit tests covering:
- A nested map key over the limit
- A map key nested inside a list
- A key four levels deep (`M → L → M → M`)
- Keys at/under the limit passing
- UTF-8 byte-length (not char-length) counting

Also updates:
- `docs/technical-debt.md` (F-17 struck through)
- `docs/dynamodb-limits.md` (Attribute name size row now notes nested coverage)

## Why

`validate_attribute_name_sizes` only checked top-level attribute names, so an oversized key nested inside a map or list attribute was silently accepted where real DynamoDB rejects it. This affects every call site that runs item validation: `PutItem`, `BatchWriteItem` (put), `TransactWriteItems` (Put), and `ImportTable`.

## Testing done

```
cargo fmt --all -- --check
cargo clippy -p extenddb-core --all-targets -- -D warnings
cargo test -p extenddb-core validation::tests     # 68 passed, 5 new
cargo test -p extenddb-core -p extenddb-engine    # 497 + 91 passed
```

Pure Rust change in the core crate — no database or server needed to verify.

**Note for reviewers:** the nested-name error wording is not separately measured against real DynamoDB; AWS's stated naming rules give the byte limit without a top-level qualifier, so the existing top-level message text is reused for nested violations rather than a new measured string.

**Out of scope:** `UpdateItem`'s `ExpressionAttributeValues` has no attribute-name-size check at all (top-level or nested) — flagged as a follow-up, not fixed here.

## Checklist

- [x] All tests pass (`cargo test --workspace`) — ran targeted `-p extenddb-core -p extenddb-engine`; not run for backends requiring Postgres/Mongo
- [x] If this changes the wire protocol, Storage trait, auth model, on-disk format, or public CLI surface, an RFC has been accepted or is linked below. Otherwise, an ADR captures the decision (link below).

**ADR / RFC:** n/a — validation-logic fidelity fix, no wire/trait/auth/CLI/schema surface change.

## Breaking changes

Behavior-tightening, not a breaking API/wire change: requests that previously succeeded with an oversized attribute name nested inside a map or list will now be rejected with `ValidationException`, matching AWS DynamoDB. No callers should depend on the old (permissive) behavior.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache License 2.0 and I agree to the Developer Certificate of Origin (DCO). See [[CONTRIBUTING.md](https://claude.ai/CONTRIBUTING.md)](../CONTRIBUTING.md) for details.


ssbytedev:
Hello! I was previously an intern at DynamoDB, and I recently came across this repository. I noticed a few low-hanging fruit opportunities and thought I’d contribute.

If you could take a look at this change and provide any feedback, I’d really appreciate it. Hopefully, we can get it merged soon. Thanks!

LeeroyHannigan (commented):
Code and tests look right. I built the branch and ran it against a live server: the four paths it names now reject an oversized nested key, and fmt, clippy, the workspace tests, and the MSRV check are clean. One documentation edit requested inline, then this can go in.

docs/dynamodb-limits.md
```diff
@@ -41,7 +41,7 @@ Source: [AWS DynamoDB Service Quotas](https://docs.aws.amazon.com/amazondynamodb
 | Maximum item size | 400 KB (409,600 bytes) | Enforced | `LimitsConfig::max_item_size_bytes`, validated on PutItem and post-UpdateItem |
 | Partition key size | 1–2,048 bytes | Enforced | `validate_key_sizes`, `LimitsConfig::max_partition_key_size_bytes`            |
 | Sort key size | 1–1,024 bytes | Enforced | `validate_key_sizes`, `LimitsConfig::max_sort_key_size_bytes`                 |
-| Attribute name size | 1–64 KB (65,535 bytes) | Enforced | `validate_attribute_name_sizes`, `LimitsConfig::max_attribute_name_bytes`     |
+| Attribute name size | 1–64 KB (65,535 bytes) | Enforced | `validate_attribute_name_sizes`, `LimitsConfig::max_attribute_name_bytes`; applies to nested map keys at every depth as well as top-level names |
```
LeeroyHannigan: Please name the operations here, as the nesting depth row below does. On a server built from this branch, UpdateItem and TransactWriteItems Update accept a 65,536-byte name at top level (`SET #n = :v`) and nested (`SET doc.#n = :v`, and `SET doc = :v` with the key inside the map), so the row should say where the limit is enforced and where it is not. Suggested notes text:

`validate_attribute_name_sizes`, `LimitsConfig::max_attribute_name_bytes`; top-level names and nested map keys at every depth, on PutItem, BatchWriteItem.PutRequest, TransactWriteItems.Put, and ImportTable. Not enforced on UpdateItem or TransactWriteItems.Update at any level.

ssbytedev: Good catch. I confirmed in the code.

`validate_attribute_name_sizes` is only wired into:

- `validate_put_item` (`PutItem`)
- `batch_write_item.rs` (`PutRequest`)
- The `Put` arm of `transact_write_items.rs`
- `import_export.rs`

It’s **not** called anywhere in `update_item.rs` or the `Update` arm of `transact_write_items.rs`, so oversized names (top-level or nested) go through unchecked on those paths.

Updated the row to **Partial** and used your suggested notes text verbatim:

> `validate_attribute_name_sizes`, `LimitsConfig::max_attribute_name_bytes`; top-level names and nested map keys at every depth, on `PutItem`, `BatchWriteItem.PutRequest`, `TransactWriteItems.Put`, and `ImportTable`. Not enforced on `UpdateItem` or `TransactWriteItems.Update` at any level.

Also adjusted the **Summary/Enforcement Coverage** table so it stays consistent:

- **Items:** `6/0 → 5/1` enforced/partial
- **Total:** `29/0 → 28/1`

Scoping this PR to the **docs fix only**. The actual `UpdateItem`/`TransactWriteItems.Update` enforcement gap is a separate code change. Happy to file a follow-up issue for it if useful.

LeeroyHannigan:
Thanks, the row reads right now.

One more thing, since the description flags the nested message as unmeasured: I measured it against the service on 2026-09-16. For a 65,536 byte attribute name, top-level or nested, on PutItem and inside an UpdateItem value, the service returns

`ValidationException`: `1 validation error detected: Attribute name is too large, must be less than 65536 bytes`

and through `ExpressionAttributeNames` the same text followed by ` for key #n`. A 65,535 byte nested name is accepted. So the existing top-level message in `check_attribute_name_size` ("One or more parameter values were invalid: Size of attribute name '...' has exceeded the maximum size limit of 65535 bytes") is not the service's either; it predates your change. If you are willing, replace the message in this PR with the measured text (one string plus the test that checks it) and drop the "not separately measured" clause from the F-17 row. If you would rather keep this PR to what it is, that is fine too and I will carry the message into the follow-up.

Yes to the follow-up issue for UpdateItem and TransactWriteItems Update; please include the ExpressionAttributeNames variant with its ` for key #n` suffix, since that path has its own message.

LeeroyHannigan (approved):
Thank you for this, appreciate your contribution. And if you wanted to do the same for Updates that would be very welcome. 

🚀

PR #348, opened 2026-09-14
docs(design): add the backup and restore design

LeeroyHannigan:
## What

Adds `docs/design/15-backup-restore.md`, the design document for off-host backup and restore, and its index line in `docs/design/README.md`.

The design: one portable backup format for all backends, the DynamoDB S3 export layout plus a single ExtendDB manifest, written to a configured backup store (a local directory or an S3 bucket) and driven by the engine through a consistent snapshot iterator each backend provides. The document covers the store abstraction, the manifest, the snapshot engine, the restore lifecycle with deferred activation, the point-in-time recovery posture for v1 (unsupported, reported with the typed exceptions the service models), configuration, migration from the current per-backend backup rows, and the compatibility goal that a backup restores into any backend, imports into Amazon DynamoDB with ImportTable, and a DynamoDB export restores into ExtendDB.

## Why

Backups today are rows inside the same database that holds the data, in a shape private to each storage backend. Every backup shares the failure domain of its data: host, disk, or database loss takes the backups with it, and loss of the catalog destroys the ability to find or restore them. The three backends disagree on what a restore produces, backups cannot cross backends or versions, and point-in-time recovery reported a window it could never restore. Per-backend fixes would leave the failure-domain, portability, and catalog-loss problems open; one engine-owned path with a portable format closes all of them.

## Testing done

Documentation only. The document contains no markdown links, so there are no internal anchors to break; the README index line's relative link resolves to the added file. `cargo test -p extenddb-storage-postgres --test doc_version_literals` passes with the new document in the scanned set.

## Checklist

ADR / RFC: this document is the design record.

---


docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
+# Design Spec: Off-Host Backup and Restore for ExtendDB
+
+Status: Draft for maintainer review, 2026-09-10. Target: v1.0. Baseline: main at 4fd2c5f8.
+
+## Summary
+
+ExtendDB backups today are rows inside the same database that holds the data, in a shape private to each storage backend. This spec replaces that with a single portable backup format written to a configured backup store, either a local directory or an S3 bucket, driven by the engine rather than by each backend. The DynamoDB API contract does not change: CreateBackup, DescribeBackup, ListBackups, DeleteBackup, and RestoreTableFromBackup keep their request and response shapes. Backends gain one new obligation, a consistent snapshot iterator over a table, and lose the obligation to implement backup storage themselves. The format is the DynamoDB S3 export layout plus one ExtendDB manifest, so a backup restores into any backend, imports into Amazon DynamoDB with ImportTable, and a DynamoDB export restores into ExtendDB.
```
jcshepherd, line 7: A "consistent snapshot iterator over a table" may not be supportable by all backends (e.g., Cassandra). We should decide what are options are for those backends. E.g., simply do not support backup/restore for those backends and expect users to use the backend database's native backup/restore features, or do a best effort backup/restore with the published caveat that they are likely inconsistent if any write activity occurs during the activity.

docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
...
+
+The spec also settles point-in-time recovery for v1: unsupported, reported honestly, with the format designed so a later change-log archive can add it.
```
jcshepherd, line 9: I'm actually thinking that we should pursue PITR first, because it'll force us towards a single standard implementation for GSI maintenance, streams, TTL, backups, etc. instead of bespoke implementations for these things (depending on whatever limitations a given engine imposes).

docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Motivation
+
+The v1 readiness analysis (v1gaps-backup-dr.md, v1gaps-data-integrity.md) found:
```
jcshepherd, line 13: Backup/restore operations are executed synchronously with the front-end API request.

docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
...
+    fn head(&self, key: &str) -> BoxFuture<Result<Option<ObjectMeta>, StoreError>>;
+    fn list(&self, prefix: &str) -> BoxStream<Result<ObjectMeta, StoreError>>;
+    fn delete_prefix(&self, prefix: &str) -> BoxFuture<Result<u64, StoreError>>;
+}
+```
+
+- `FilesystemStore`: root directory from config. Keys map to paths under the root; the existing export-path sandboxing rules apply (canonicalize, refuse traversal, refuse symlinks that resolve outside the root). Writes go to a temporary name and rename into place so a crash never leaves a truncated file under its final name.
+- `S3Store`: bucket, prefix, region, optional endpoint and path-style flag for S3-compatible stores, credentials from the standard AWS SDK chain (environment, shared config, IMDS, IRSA, container credentials). Uploads over `upload_part_size_bytes` use multipart; failed multipart uploads are aborted. Dependency: the official `aws-sdk-s3` crate, chosen over `object_store` because it inherits the SDK credential chain, IRSA support, and retry behavior that operators already configure for other AWS tooling. The dependency adds to the licenses check (Apache-2.0, no notice change expected).
```
jcshepherd, line 108: At first glance, these objections to object_store look invalid: it looks like all of those behaviors, etc., are supported by object_store already for S3. I'd take a second look: it'd be great to provide users that flexibility.

docs/design/15-backup-restore.md
```diff
@@ -0,0 +1,311 @@
...
+source_table_details_json
+```
+
+The store is the source of truth; the row is a cache. `extenddb backup sync` lists `<root>/<account>/` prefixes, reads every `extenddb-manifest.json`, and upserts rows, marking rows whose prefix is gone as DELETED and prefixes with no row as AVAILABLE. This is the recovery path when the catalog is lost and the store survives, and it is how a fresh ExtendDB adopts a bucket of backups from another instance.
+
+The ARN encodes account, table, and backup id, and the key prefix is derived from those three fields deterministically, so `describe_backup(arn)` works from the store alone after a sync.
+
+### 5. Operations
+
+#### 5.1 CreateBackup
```
jcshepherd, line 158: When/how does the catalog data itself get backed up?

jcshepherd (commented):
A few questions below. I might take a shot at writing a proposal for a log-driven approach to PITR.

PR #350, opened 2026-09-15
test: Porting over the Cassandra Repo integration suite

robinnsc:
## What

Porting over the original extenddb-cassandra-plugin repository's direct integration suit. 136 tests that exercise the storage traits directly against a real Cassandra node (139 ported, minus the four plug-in tag tests deduplicated into the consolidated in-tree lifecycle test when the metadata suite was folded in — see the second commit) — no HTTP server in between — of which the in-tree port had previously picked up only the four tag tests.

Coverage the tree gains: CRUD and key-type handling (decimal, binary), Query/Scan semantics, pagination, segments and index reads, transactions (TransactGet, TransactWrite rollback and idempotency, ledger internals, lock protection), GSI/LSI physical schema and the async propagation workers, streams (write, read, pagination, shards), table lifecycle, and the whole management plane: access keys, accounts, users, groups, roles, policies, admin, settings, backup/restore state. None of that had automated coverage in-tree; the TTL suites use some of these operations but never assert their general semantics.

Structure: one test binary (`tests/direct_integration.rs`), one module per area under `tests/direct/`, all sharing `tests/common/mod.rs` — which is the in-tree descendant of the plug-in's `helpers.rs`, so the port is mostly mechanical: keyspace-prefix adaptation, two `CreateTableInput` fields added since the fork, and the skip-without-Cassandra guard on every test so the serviceless workspace `cargo test` job stays green (same pattern the existing suites use, `EXTENDDB_TEST_CASSANDRA=required` in the dedicated workflow).

## Why

135 of 139 tests had no in-tree equivalent (~8,600 LOC of coverage). This suite is the safety net under everything else the in-tree port changes.

## Not copied verbatim

Four tests the porting inventory flagged as weak were strengthened, and one of them promptly caught something:

- **The sync-GSI test's TODO was hiding a real semantic difference.** The plug-in engine wrote GSI rows synchronously by default; in-tree the default is 10ms *async* propagation through the GSI queue, which no test worker drains. The old test "verified the write path executes without crashing" — it would never have caught a lost index write, and when I added the missing assertion it failed immediately. It now pins the index to synchronous propagation (the same catalog knob production routing reads) and asserts the row is visible through the index-scan path.
- The table-lifecycle and settings tests printed ✓/✗ outcomes instead of asserting them; every outcome is now an assertion. The settings test also mutated `control_plane_delay_seconds` — a live control-plane knob in the shared catalog — and left the test value behind; it now restores what it found (or the reader default, 0.25, when the key was absent).
- The pk-only delete-timestamp test carried dead bindings from an inspection it never did; it now verifies the delete actually removed the item, with a comment explaining why the timestamp itself is uninspectable on that table shape.

## Parallelism and CI

The suite runs with default parallelism — every test provisions its own account and keyspace. Full run on a local single node: ~17–19 minutes wall, 139/139. Wired into the Cassandra CI workflow as a step after the existing serial suites; this will grow that job's runtime, and if it becomes the long pole the next step is splitting it into its own job in the same workflow.

One observed interaction worth knowing: running the *other* TTL suite immediately after this one on the same node hit one flaky claim-contention failure while Cassandra was still dropping this suite's 139 keyspaces in the background (test-fixture Drop is async fire-and-forget). It passes reliably in isolation and CI runs the suites in the opposite order, so nothing is changed here — flagging it in case it ever shows up in someone's local run.

## Testing done

Real Cassandra 4.1, single node:

- `cargo test -p extenddb-storage-cassandra --test direct_integration` — **139 passed, 0 failed** (two full runs, plus targeted reruns of every strengthened test)
- Existing suites unaffected: `ttl_integration` 26/26 serial, `metadata_integration` 1/1, lib 43/43
- `cargo clippy --all-targets -- -D warnings` — clean (the ported code carried ten lint findings; fixed)
- Skip-guard behavior inherited from the existing suites' helper (probe 9042, `required` in CI)

## Checklist

- [x] Test-only change; the one non-test file touched is the CI workflow
- [x] No new Clippy warnings (workspace `-D warnings` clean)
- [x] No wire, trait, schema, or behavior change


robinnsc:
Rebuilt on the current base after the branch reconstruction dropped merged PR #339 (see #349 for the full story; restored as 439ca5b, which this suite depends on for the skip-guard helper and CI workflow). Worth highlighting: this suite already paid for itself before merging — the delete_item tests caught the rebuilt base's OCC delete fast path silently skipping partition_max_delete_timestamp, which disarmed the transaction prepare path's new-item-vs-delete ordering check for plain deletes. Fixed on the base as 0aebbe4; 139/139 green on top of it.

crates/storage-cassandra/tests/direct/accounts.rs
```diff
@@ -0,0 +1,144 @@
...
+//! Integration tests for Account management operations for `CassandraCatalogStore`.
+
+#[cfg(test)]
+mod tests {
+    use crate::helpers::{setup_engine, test_config, unique_test_account, unique_test_id};
+    use extenddb_storage::management_store::ManagementStore;
+
+    #[tokio::test]
+    async fn test_create_account() {
+        if crate::helpers::skip_without_cassandra() {
```
jcshepherd, line 13: Just checking: is this function already pushed?

jcshepherd (approved):
Overall, LGTM. Two questions:
* is helpers::skip_without_cassandra() already checked in?
* Do you plan to migrate storage-cassandra/tests/metadata_integration.rs and ttl_integration.rs to this suite?

PR #351, opened 2026-09-15
fix(auth): compute the SigV4 payload hash from the received body

LeeroyHannigan:
## What

`canonical_request()` in `crates/auth/src/sigv4/canonical.rs` now computes the hashed-payload line as the SHA-256 of the body the server received. Previously it took the value from the client's `x-amz-content-sha256` header when that header was present and hashed the body only when it was absent.

A signed `x-amz-content-sha256` header is still covered as an ordinary canonical header, so clients that send it keep working. The `UNSIGNED-PAYLOAD` rejection in `verify_signature()` is unchanged.

Also:
- `docs/manuals/10-security-model.md`: step 6 states where the payload hash comes from; steps 8 and 9 now give the error names and HTTP status codes the server returns (`InvalidSignatureException` 400, `UnrecognizedClientException` 400).

## Why

A request signed for one body and transmitted with another, carrying the hash of the signed body in `x-amz-content-sha256`, verified and executed. On a server built from main, `PutItem` signed for `{pk: signed}` and sent with the body `{pk: tampered, injected: yes}` returned HTTP 200 and stored the second item. The header let the sender choose the value the signature was checked against, so anyone able to alter traffic between a client and the server could rewrite any signed request into another request of the same operation.

## Testing done

Unit tests (`extenddb-auth`): the canonical request's payload line comes from the body when the header claims a different hash; a correctly signed body verifies; a body swapped after signing fails with `InvalidSignatureException`, with and without the header carrying the signed hash; an unsigned stray header does not affect verification. The two new tests that exercise the header path fail on the previous `canonical.rs` and pass with this change.

Integration (`tests/test_auth_error_fidelity.py::TestSignedPayloadIntegrity`): signs one `PutItem` body, sends another with the signed hash in the header, expects 400 `InvalidSignatureException`, and reads both keys back to confirm nothing was stored; a control request with an unchanged body and the header succeeds. Against a SQLite server built from main the header variant fails (`assert 200 == 400`); against this branch all three pass. The class runs against the service as well when `EXTENDDB_TEST_ENDPOINT` is unset.

```
cargo fmt --all -- --check
cargo test --workspace                       # 1,186 passed
devtools/run-tests --extenddb --pytest --rust-integration --parallel --backend sqlite
  # pytest 981 passed, 23 skipped, 1 xfailed; rust integration 511 passed
  # (import/export excluded as in CI: the ad hoc server has no [import]/[export] paths)
```

## Checklist

ADR / RFC: n/a. The verification steps in `docs/manuals/10-security-model.md` and REQ-AUTH-003 already describe the canonical request as built over the body hash; this change makes the code match them.

## Breaking changes

None for conforming clients. A client that signed a request over a payload hash different from its transmitted body, and relied on the header to make the signature verify, is now rejected with `InvalidSignatureException`, the same outcome the service gives such a request.

---


jcshepherd (approved):
LGTM but leaves me wondering ... what is the x-amz-content-sha256 header used for, other than fodder for the request signing function?

PR #354, opened 2026-09-16
fix(engine): read Query and Scan pages from storage in chunks

LeeroyHannigan:
## What

`Query` and `Scan` now read from storage in chunks and stop as soon as the page is full. `read_page()` in `crates/engine/src/read_helpers.rs` asks storage for 128 items after the cursor, feeds them one at a time through the post-read pipeline (`FilterExpression`, `ProjectionExpression`, `Select`, the 1 MB page budget), and asks for another chunk only while the page is not full, `Limit` has not been reached, and storage reports more. Later chunks are sized from the average item size seen so far to fill the remaining budget, clamped to 32..1,024 items. The cursor for each chunk is the same key shape a client sends as `ExclusiveStartKey`, so there is no `Storage` trait change: each backend answers with the keyset pagination it already implements. A storage page that does not advance the cursor is an `InternalServerError` rather than a loop, and a backend that returns more items than asked cannot push a page past `Limit`.

The MongoDB `Scan` path (`crates/storage-mongodb/src/data_engine.rs`) now sizes its cursor's network batches to the requested count. It streamed with the driver's default batching (101 documents, then up to 16 MB per `getMore`), so every chunk pulled and discarded a multi-megabyte batch; with the hint a chunk costs one batch. The segment filter still runs client-side, so this is a network hint and not a bound.

`scan.rs` and `query.rs` call `read_page()` in place of one unbounded storage read followed by `apply_post_read()`. `ConsumedCapacity` and the `returned_bytes` metric are computed from the evaluated items.

Two client-visible behaviors change at the page boundary:
- A page whose items all fail `FilterExpression` ends at 1 MB with `Count: 0` and a `LastEvaluatedKey`. Before, the engine kept scanning until an item passed, which made a selective filter over a large table an unbounded single request. DynamoDB documents the empty page with `LastEvaluatedKey`.
- `ConsumedCapacity` reflects the page's evaluated bytes rather than every row storage returned.

Docs: `docs/dynamodb-limits.md` (page limit row) and the Query walkthrough in `docs/design/02-high-level-design.md`.

## Why

A `Query` or `Scan` without `Limit` (the SDK default for `Scan`) asked storage for every matching row (capped at 1,000,001 on postgres and sqlite, uncapped on mongodb), held the whole result in memory, then cut the page. Measured on a 100,000 item, 4 KB per item table against servers built from main 0a2b3bd3:

| backend | page 1 time | page 1 server RSS growth | 20 pages time | server RSS after 20 pages |
|---|---|---|---|---|
| postgres | 1.67 s | 56 to 778 MB | 29.8 s | 2,667 MB |
| sqlite | 1.07 s | 72 to 608 MB | 19.0 s | 2,508 MB |
| mongodb | 1.09 s | 58 to 542 MB | 15.9 s | 1,143 MB |

Each later page repeated the fetch from its cursor to the end of the table, so a full pagination (about 381 pages here) was quadratic in table size, and one client with a default `Scan` could push the server toward memory exhaustion.

## Testing done

Same table, same harness, servers built from this branch:

| backend | page 1 time | page 1 server RSS growth | 20 pages time | server RSS after 20 pages |
|---|---|---|---|---|
| postgres | 0.02 s | 49 to 50 MB | 0.50 s | 52 MB |
| sqlite | 0.01 s | 49 to 50 MB | 0.47 s | 78 MB |
| mongodb | 0.06 s | 63 to 79 MB | 0.58 s | 191 MB |

Page contents are identical before and after on every backend: 263 items and 1,045,425 bytes for the `Scan`; for a `Query` over a 30,000 item, 1 KB per item partition, 1,077 items and 1,047,888 bytes, page 1 in 0.03 to 0.08 s (from 0.26 to 0.35 s).

New tests:
- `crates/engine/src/read_helpers.rs`: ten unit tests drive `read_page()` against an in-memory store that records every read. They cover a 100,000 item table read in at most three chunks, the budget cut inside a chunk with the key of the last evaluated item, the filtered empty page, `Limit` paging (7, 7, 6 and exactly-the-limit), tiny items read in growing chunks, exhaustion, `Select: COUNT`, an empty or repeated storage page (error, not a loop), and a backend that returns more than asked.
- `tests/test_read_paging.py`: 15 integration tests against a live server, in the style of `test_query_scan.py`: 1.5 MB tables paginated exactly once with and without `Segment`/`TotalSegments`, on the base table and a GSI; the filtered empty page; `Limit` interaction; mixed 100 B and 300 KB items; page-based `ConsumedCapacity`; 3,000 item partition queries in both orders on the base table and a GSI. On servers built from main, 13 pass and the two boundary-behavior tests fail; on this branch all 15 pass on postgres, sqlite, and mongodb.

Further runs against servers built from this branch and from main: pagination under concurrent inserts, deletes ahead of and behind the cursor, and 1 KB to 300 KB item growth (plain, single-partition Query, GSI, and 4-segment parallel Scan; three rounds, zero lost or duplicated stable items, every page within 1 MB plus one item, identical to main); 31 cursor-shape probes per backend (N, B, and S sort keys at their extremes in both orders, non-unique and sparse GSIs, an LSI, `Select COUNT` page sums, projections that drop key attributes, `Limit` at and beside the chunk boundaries, malformed `ExclusiveStartKey`) with identical item sequences to main; and workloads chosen to favor main: 200,000 items of 60 B (this branch 4.5x faster on page 1), 32 concurrent clients paginating the 100,000 item table for 30 s (116 pages per second against 2 to 3, with main's server reaching 8 to 9 GB resident), a 5,000 item table that fits one page (this branch within 6 percent of main on postgres and sqlite, 74 against 68 ms on mongodb), and a never-matching filter paginated to completion (381 requests against main's one, 1.2 to 1.9 s against 0.9 to 1.3 s, with main's single request driving the server to 1.0 to 1.5 GB resident).

Per-backend read plans for the chunk shape (`limit` plus `ExclusiveStartKey`) were captured while designing this: sqlite and mongodb answer every shape with an index seek whose examined rows equal the chunk size. Postgres seeks on `pk` but not on `sk_s`, because the tables' text columns carry the database collation while the engine compares with `COLLATE "C"`, so a chunk whose cursor sits inside a large partition with S sort keys re-reads that partition. That is pre-existing (main's single read did the same once per page), the fix is still faster on such tables, and the index change is filed separately.

```
cargo fmt --all -- --check
cargo test --workspace                       # 1,188 passed
devtools/run-tests --extenddb --pytest --rust-integration --parallel --backend sqlite     # 986 passed; rust 511 passed
devtools/run-tests --extenddb --pytest --rust-integration --parallel --backend postgres   # 987 passed; rust 511 passed
```
Remaining failures in those runs were environment: import/export tests against an ad hoc server whose allowed path is spelled through a symlink (excluded in CI), and on postgres `test_list_tables_pagination`, which fails on main as well when table names sort differently under byte order and the database collation (filed separately).

## Checklist

ADR / RFC: n/a. No `Storage` trait or wire change; the engine uses the existing `limit` and `exclusive_start_key` parameters.

## Breaking changes

None to the API shape. Two responses change content: a page whose items all fail the filter now returns `Count: 0` with a `LastEvaluatedKey` where it previously scanned on; and `ConsumedCapacity` on a truncated page is now the page's, not the table remainder's. Both match the service.

---


LeeroyHannigan:
Second commit, 1c43db35: a page that reaches `Limit` now always carries a `LastEvaluatedKey`, and the page after it is empty with none. Measured against the service on 2026-09-16: Query Limit 5 over exactly 5 items and Scan Limit 6 over exactly 6 both return a key. Before (on main as well), the key appeared only when storage reported another row, because every backend peeks one row past the requested count. The two paging tests that recorded the old behavior as an observation now assert this one; unit tests updated. Gates re-run: fmt, clippy, 1,191 workspace tests, MSRV, the 15 paging tests and the query and scan suites (135) against a SQLite server built from this head.

crates/engine/src/read_helpers.rs
```diff
@@ -61,54 +63,101 @@ pub fn validate_projection_unused_names(
...
-    let mut last_processed_key: Option<Item> = None;
-    let is_count = matches!(select, Some(Select::Count));
-
-    for item in raw_items {
+/// The `DynamoDB` page budget: a Query or Scan evaluates at most this many bytes
+/// of items before it returns a page.
+pub const PAGE_BYTE_BUDGET: usize = 1_048_576;
+
+/// Items requested from storage on the first chunk of a page.
+const FIRST_CHUNK: i64 = 128;
```
robinnsc, line 79: The first chunk is byte-blind: later chunks are sized from the observed average item size, and the `MAX_CHUNK` comment explicitly justifies the 1,024 cap by the 400 MB worst case — but this first chunk of 128 items has no size information, so on a table of 400 KB items it pulls roughly 50 MB from storage to fill a 1 MB page (~3 items), discarding the rest, on every page-1 request. Same class of risk the MAX_CHUNK comment guards against, an order of magnitude smaller but reachable by any client in a single request. Either start at 32 and let the growth estimate take over, or document here why 128 is an acceptable ceiling.

robinnsc (commented):
One sequencing point that isn't about this diff itself: the chunk cursor is an ExclusiveStartKey-shaped key issued per chunk, so backend cursor defects that used to fire once per page now fire once per chunk — including the reverse hash-only GSI repeat/skip bug fixed in #362 and the Postgres partition re-read fixed in #361 (which this PR's own description measures). #361/#362 should land before or together with this. Once they're in, `test_read_paging.py` could add a reverse hash-only-GSI pagination case, which is the one shape the current suite doesn't cover.

robinnsc (approved):

PR #355, opened 2026-09-16
fix(auth): check the scope and session token, canonicalize repeated headers, and time out stalled requests

LeeroyHannigan:
## What

Four request checks in the SigV4 path brought in line with the service, and request timeouts on every connection. Each service behavior cited below was measured against `dynamodb.us-east-1.amazonaws.com` on 2026-09-16 with a hand-rolled signer, and each is reproduced by a test in this PR.

- Credential scope (`crates/auth/src/sigv4/verify.rs::check_scope`, called from the provider): the region must equal `[server] region` and the service must be `dynamodb`; every failing part is reported in one `InvalidSignatureException` message in the service's order, `"Credential should be scoped to a valid region. "` then `"Credential should be scoped to correct service: 'dynamodb'. "` (the region is not named; the trailing spaces are the service's). Before, the signing key was derived from whatever region the client wrote, so a request signed for one deployment verified against another that shared its credentials. A malformed scope now reports the service's `IncompleteSignatureException` wire name (was `IncompleteSignature`) and its five-element message.
- Session token with a long-term key (`lib.rs`): rejected after signature verification with `UnrecognizedClientException: "The security token included in the request is invalid"`. Before, the token was ignored. The message for a presented token that does not resolve has no trailing period; the unknown-key and missing-token messages keep theirs, as measured.
- Repeated headers (`crates/auth/src/sigv4/canonical.rs`): values of a header sent more than once are joined with commas in the order sent. Before, only the first value was read, so a client that sent a signed header twice could never authenticate; botocore signs such headers comma-joined and the service accepts them. A value that is not visible ASCII contributes an empty value rather than vanishing, so a request carrying such a header cannot canonicalize the same as one without it. A request with more than one `host` or `authorization` header is refused with a plain 400 and `Connection: close` before anything else, as the service's front end does (`crates/server/src/handler.rs`).
- `UNSIGNED-PAYLOAD` (`crates/auth/src/sigv4/verify.rs`): the special rejection is removed. The service answers a client that signs that literal with the plain signature mismatch, and since #351 the payload line is always the hash of the received body, so the mismatch falls out of verification with the measured text.
- Request timeouts (`crates/server`, `crates/config`): new `[server] request_timeout_secs`, default 30, sets the HTTP/1 header read timeout on the TLS acceptor, the HTTP/2 keep-alive PING interval and timeout (both servers negotiate `h2` by ALPN), and a body inactivity timeout on every route (`tower_http::timeout::RequestBodyTimeoutLayer`). A stalled body is answered with HTTP 408 and `Connection: close`; a stalled HTTP/1 header phase is closed at the timeout and a stalled HTTP/2 connection within twice it (measured: 10.0 s at a 5 s setting). The header timeout also bounds how long an idle keep-alive connection waits for its next request, which is hyper's own 30 second default (never active before because no timer was installed). `request_timeout_secs = 0` is refused at startup. Hardening, not fidelity: nothing bounded either before.
- Docs: `docs/manuals/10-security-model.md` lists every verification check in the order the code runs them, and its credential storage section now describes the in-process cache (`[auth.cache]`, on by default) in place of a sentence that denied it. `docs/manuals/05-admin-guide.md` documents `region` as the accepted scope and the new timeout key; the `init` template mentions both.

Two candidate changes were measured and deliberately not made: the service parses the request body before verifying the signature (a malformed body with a wrong secret or no Authorization header returns `SerializationException`), which is the handler's existing order; and the service accepts unsigned `x-amz-*` headers, including `x-amz-target`, so no signed-header requirement beyond `host` is added.

## Why

Findings from the review of #351. Two are correctness against the service (region and token), one is a compatibility defect (a client repeating a signed header could not authenticate), one is a message divergence, and the timeouts close a denial-of-service hole in which a `Content-Length` the client never delivers held a connection open indefinitely (reproduced: still open after 20 s on main; closed at the configured 5 s on this branch).

## Testing done

Unit (`extenddb-auth`, 160 passed): comma-joined repeated header in the canonical request; `UNSIGNED-PAYLOAD` signed as the payload line fails as a plain mismatch with no special wording; end-to-end provider tests with a correctly signed request for the wrong region, a token with a long-term key, a matching session token, a bogus session token, and a missing session token, each asserting the exact measured message; exact (case- and whitespace-sensitive) region comparison; an empty token header on a long-term key; an unsignable duplicate header value changing the canonical form.

Integration (`tests/test_auth_error_fidelity.py::TestRequestChecks`): wrong-region scope, scope wrong in both parts, token with a long-term key, repeated signed header sent as two lines, `UNSIGNED-PAYLOAD` literal, duplicated `host`, and a stalled body. Against a SQLite server built from this branch with `request_timeout_secs = 5`: 14 of 14 in the file pass. Against a server built from main: six of the seven new cases fail (main already answered a duplicated `host` with a 400, as a signature mismatch). Against the service itself (endpoint unset): region, both-parts scope, repeated header, `UNSIGNED-PAYLOAD`, and duplicated `host` pass, so those five assertions are measured, not inferred; the token case skips with temporary credentials and the timeout case is server-only by design.

Stalls probed by hand at a 5 s setting: HTTP/1 header stall closed at 5.0 s (main holds it past 20 s); HTTP/2 preface-then-stall closed at 10.0 s (main holds it past 15 s); a body delivered one byte every 4 s succeeds (inactivity, not total time); boto3 and the Node SDK reuse pooled connections across the idle close without a caller-visible error.

```
cargo fmt --all -- --check
cargo test --workspace                       # 1,196 passed
cargo +1.88.0 check --workspace --locked     # Cargo.lock: hyper-util added to the server crate's list (already in the tree)
devtools/run-tests --extenddb --pytest --rust-integration --parallel --backend sqlite   # 988 passed; rust 511 passed
```
The only integration failures were the six import/export tests, which need `[import]`/`[export]` paths the ad hoc server lacks (excluded in CI).

## Checklist

ADR / RFC: n/a. No new surface; the checks bring existing verification in line with the service, and the timeout is a new optional `[server]` key with a default.

## Breaking changes

Region scope enforcement. A client whose SDK region differs from the server's `[server] region` (default `us-east-1`) is now rejected with `InvalidSignatureException`, where it was silently accepted. This is what the service does for a scope other than the endpoint's, but deployments may have relied on the old behavior without knowing. Remedy: set `[server] region` to the region clients sign for, or configure clients with the server's region. Release notes should carry this line.

Timeouts. A client that pauses for more than `request_timeout_secs` (default 30) while sending headers or a body is disconnected. No SDK pauses that long; the value is configurable.

---


crates/server/src/lib.rs
```diff
@@ -250,9 +259,27 @@ pub async fn start_server(
...
-            .handle(handle)
+            .handle(handle);
+        // HTTP/1: a connection that has not finished sending its request
+        // headers within `request_timeout` is closed; the same bound applies
+        // to the idle wait for the next request on a keep-alive connection.
+        server
+            .http_builder()
+            .http1()
+            .timer(hyper_util::rt::TokioTimer::new())
+            .header_read_timeout(request_timeout);
```
robinnsc, line 272: These connection timeouts only reach the TLS path: the plain-HTTP `else` arm below calls `axum::serve` directly, which installs no timer, so the `header_read_timeout` and HTTP/2 keep-alive settings never apply there. A stalled HTTP/1 header phase — the exact hole this PR closes — stays open on that branch; only the body-timeout layer covers it. If that branch is unreachable in production because TLS is mandatory, a comment saying so (and why dev-mode accepts the gap) closes the question; otherwise the same config needs to reach it.

crates/server/src/handler.rs
```diff
@@ -47,14 +47,43 @@ pub(crate) async fn handle_request(
...
-            &DynamoDbError::RequestEntityTooLargeException(
-                "Request size has exceeded the maximum allowed size".to_owned(),
-            ),
-            &request_id,
-        );
+    // A request carrying more than one `host` or `authorization` header is
+    // refused outright, as the service's front end does (400, connection
+    // closed). Either duplicate is a header-confusion shape a proxy in front
+    // of this server could resolve differently from the signature check.
+    if headers.get_all("host").iter().count() > 1
```
robinnsc, line 54: This refusal only covers the DynamoDB route: it lives in `handle_request`, so `/console` and `/management` never see it. The stated rationale — a proxy in front of the server resolving header confusion differently — applies at least as much to the session-authed console as to the SigV4 route, where the signature already binds the host. A thin layer applied alongside the other security layers would cover every route with the same code.

PR #361, opened 2026-09-17
fix(postgres): COLLATE "C" key columns and row-comparison page cursors

### LINKED ISSUE #357: [Bug] PostgreSQL: ListTables pagination skips names because the filter and the ordering use different collations
## Describe the bug

On the PostgreSQL backend, paginating `ListTables` with `ExclusiveStartTableName` can skip tables. The page query in `crates/storage-postgres/src/table_engine.rs` (lines 63 to 75) orders with `ORDER BY table_name COLLATE "C"` (byte order, which is what Amazon DynamoDB uses) but filters with `WHERE table_name > $2`, and that comparison uses the column's own collation, which is the database default (`en_US.utf8` in the CI and default container setups). The two collations disagree whenever a name contains punctuation: `en_US.utf8` ignores `_` at the first comparison level, so `ab` sorts after `a_z` in byte order but before it in `en_US.utf8`. A page ending on `a_z` therefore asks for names `> 'a_z'` under `en_US.utf8`, and `ab` does not qualify.

Table names with underscores are common (`orders_2026`, `user_sessions`), so any deployment that pages `ListTables` with a `Limit` can miss tables.

## To Reproduce

Against a server built from main `9e400357` on PostgreSQL 16 (database collation `en_US.utf8`):

1. Create two tables named `lt9f6e92_a_z` and `lt9f6e92_ab`.
2. `ListTables(Limit=1)`, then follow `LastEvaluatedTableName` until it is absent.
3. The pages return `['lt9f6e92_a_z']` only; `lt9f6e92_ab` never appears.

The same steps on the SQLite and MongoDB backends return both names, because their comparison and ordering agree.

## Expected behavior

Every table appears on exactly one page, and pages come in byte order, as the service returns them.

## Actual behavior

Tables whose names sort differently under the two collations are skipped (or, in the mirror case, repeated).

## Suggested fix

Qualify the comparison the same way as the ordering: `WHERE account_id = $1 AND table_name COLLATE "C" > $2`. The `tables` table is small, so the planner change does not matter here. The same mismatch affects the data tables in a different way (query cost rather than correctness); that is #358.

## Environment

- ExtendDB main at 9e400357
- PostgreSQL 16 in a container, database collation `en_US.utf8`
- boto3 1.40, Python 3.12

### LINKED ISSUE #358: [Bug] PostgreSQL: Query and Scan cursors seek on pk only because sort-key columns are not COLLATE "C"
## Describe the bug

On the PostgreSQL backend, a Query or Scan that resumes from a cursor inside a partition with a string sort key re-reads the partition from its start on every page (and, since #354, on every chunk). The engine compares and orders `sk_s`, `base_pk`, and `base_sk_s` with `COLLATE "C"` to get DynamoDB's byte order (`crates/storage-postgres/src/data/query_scan.rs`, lines 35 to 99), but the data tables declare those columns as plain `TEXT` (`crates/storage-postgres/src/data/ddl.rs`, lines 63, 78, 159, 174, 298), so they and their btrees carry the database default collation (`en_US.utf8`). PostgreSQL cannot use an index built under one collation to answer a range predicate under another, so only the leading `pk` comparison, which has no collation qualifier, is served by the index.

## To Reproduce

`EXPLAIN (ANALYZE, BUFFERS)` on the statements the code builds, against a 50,000 item table (`pk S`, `sk S`; partition `hot` holds 10,000 items, the rest have unique partition keys) on PostgreSQL 16 with `en_US.utf8`:

- Query within `hot` resuming at `sk > 's00005000'`: Bitmap Heap Scan with `Recheck Cond: pk = 'hot'` and `Filter: sk_s > 's00005000'`; all 10,000 rows of the partition examined, 5,001 removed by the filter, 451 buffers, 3.3 ms. Every page or chunk pays this regardless of where the cursor sits.
- Base-table Scan resuming at a cursor inside `hot`: Incremental Sort presorted on `pk` only, Index Scan with `Index Cond: pk >= 'hot'` and the row comparison demoted to a Filter.
- GSI Scan resuming inside a 1,250 row GSI partition: 1,251 rows read to return 625.

With a probe index `CREATE INDEX ON <data_table> (pk, sk_s COLLATE "C")` in place, the Query becomes `Index Cond: ((pk = 'hot') AND (sk_s > 's00005000'))`, 129 rows examined, 112 buffers, 0.11 ms, and the Scan resumes as a pure index seek on `ROW(pk, sk_s) > ROW('hot', 's00005000')`, 0.06 ms.

SQLite and MongoDB answer every chunk shape with an index seek whose examined-row count equals the chunk size.

## Expected behavior

Resuming a page or chunk costs work proportional to the rows returned, not to the size of the partition the cursor sits in.

## Actual behavior

Cost per page or chunk is O(partition size) for string sort keys. Numeric and binary sort keys are unaffected: `sk_n NUMERIC` has no collation and `BYTEA` compares bytewise, so those cursors seek on the existing primary key.

## Suggested fix

Declare the text key columns `COLLATE "C"` in the DDL (`sk_s`, `base_pk`, `base_sk_s`, and `pk` for consistency), so the primary key and the GSI ordering index are built in byte order and every `COLLATE "C"` predicate is sargable. Existing deployments need a migration that rebuilds the indexes (or adds the collated index alongside). The `ListTables` comparison has the same root cause with a correctness rather than a cost effect; that is #357.

## Environment

- ExtendDB main at 9e400357 (measured on 0a2b3bd3 and the #354 branch; the DDL is unchanged)
- PostgreSQL 16 in a container, database collation `en_US.utf8`

LeeroyHannigan:
## What

PostgreSQL backend: the text key columns of data, index, and vector tables are declared `TEXT COLLATE "C"`, the forward GSI, hash-only GSI, and LSI page cursors are row comparisons, and the `ListTables` page filter uses the same collation as its ordering.

- `crates/storage-postgres/src/data/ddl.rs`: `pk`, `sk_s`, `base_pk`, `base_sk_s`, and their numbered variants take `TEXT_KEY` (`TEXT COLLATE "C"`). The primary key and the ordering indexes are therefore built in UTF-8 byte order, which is the order the query builders already ask for with `COLLATE "C"` on every string key predicate and `ORDER BY`. Before, the columns inherited the database collation (`en_US.utf8` in the default setups), and PostgreSQL cannot use an index built under one collation for a range predicate under another, so a cursor resuming inside a partition with a string sort key seeked on `pk` alone and filtered or sorted the rest of the partition.
- `crates/storage-postgres/src/data/query_scan.rs`: where every cursor column runs in the direction of the `ORDER BY`, the cursor is `(sk, base_pk, base_sk) > ($1, $2, $3)`. The expanded `a > $1 OR (a = $1 AND b > $2 ...)` returns the same rows (also with NULL binds and NULL columns, which both forms evaluate the same way: a NULL only matters once every earlier pair is equal, and then both yield NULL) but is planned as a bitmap union that collects every row past the cursor before the `LIMIT` applies, or as an index scan filtered from the start of the partition. The reverse GSI cursor keeps the expanded form because its sort key runs backwards while its base-key tie-breaker stays ascending.
- `crates/storage-postgres/src/table_engine.rs`: `WHERE table_name COLLATE "C" > $2`, matching `ORDER BY table_name COLLATE "C"`.
- No migration. Tables created by earlier binaries keep the database collation on those columns; their results stay byte-ordered because the predicates carry the collation explicitly, and only the index use differs. Recorded as technical debt C-11 with the two migration shapes and their costs.
- Docs: `docs/design/04-component-storage.md` key storage bullets and the GSI cursor example.

## Why

Closes #357
Closes #358

Amazon DynamoDB orders string keys by UTF-8 byte value. Measured 2026-09-16 on a (pk S, sk S) table with keys chosen so byte order and locale order disagree: forward order `A B Z _ a "a b" a-b a0 a_z ab a~ b z é 中`, reverse the mirror, `BETWEEN '_' AND 'b'` returning `_` through `b`, `begins_with 'a'` returning the seven keys starting with lower-case `a`, and `ListTables` returning names in byte order with no name skipped across `Limit=1` pages. Main's query order already matched; what did not was the cost of getting it, and `ListTables`, whose filter ran under `en_US.utf8` while its order ran under `COLLATE "C"`, so a page ending on `..._a_z` skipped `..._ab` (reproduced on a server built from main; SQLite and MongoDB return both).

The cost matters more with #354, which reads pages in chunks: a per-page partition re-read becomes a per-chunk one. Measured with chunked reads on 200,000 items of about 100 bytes in one partition with a GSI (#354 alone versus #354 plus this change, same layout, same container): full Query pagination 10.27 s to 6.50 s, GSI Query 2.55 s to 1.61 s, Scan 11.22 s to 6.49 s. With main's page reads the two builds are within noise of each other (18.5 s and 19.7 s), because main reads every row after the cursor on each page and that dominates. `EXPLAIN (ANALYZE, BUFFERS)` on a 50,000-row partition: the forward GSI page costs 29 buffers with the row comparison inside the Index Cond, against 25,301 buffers for the expanded form on the same data; base, LSI, hash-only GSI, Scan, N and B sort key, `begins_with`, and `BETWEEN` cursors are all index seeks bounded by the `LIMIT`. Load time was the same on both builds once run order was controlled for (the second load on the shared container is slower whichever build runs it).

## Testing done

Unit (`extenddb-storage-postgres`, 6 new): the fragment `build_pagination_where` emits for every shape (base, forward GSI over the full base key, reverse GSI expanded form, GSI on a hash-only base, LSI both directions, hash-only index). These exist because the two forms return the same rows and only a plan tells them apart, so a revert to the expanded form would otherwise pass every test.

Storage-level (`crates/storage-postgres/tests/key_collation.rs`, needs `EXTENDDB_TEST_PG_CONNECTION_STRING` as `vector_control_plane.rs` does): every text key column of the data table, the GSI table, and a hash-only table reports collation `C` in `pg_attribute`; `EXPLAIN` of the Query resume, Scan resume, and GSI resume shapes shows an index scan with `sk_s` in an Index Cond and not in a Filter; `ListTables` with `Limit=1` over `lt_a_z`, `lt_ab`, `lt_b` returns all three in that order. All three fail with main's source (collation `default`; plan is a Sort with no index scan; `lt_ab` skipped) and pass with this branch.

Wire (all backends): `tests/test_query_scan.py::TestStringSortKeyByteOrder` (forward, reverse, `BETWEEN`, `begins_with`, `Limit=1` cursor walk over the measured key set) and `tests/test_table_operations.py::TestListTablesOrder::test_list_tables_pages_in_byte_order_without_skipping`. Against the service: 7 passed. Against servers built from main: SQLite 7, MongoDB 7, PostgreSQL 6 with the `ListTables` case failing. Against a PostgreSQL server built from this branch: 7.

Mixed schema: a deployment created and loaded by the main binary, then served by this branch's binary: base and old-GSI cursors byte-ordered with no skip or repeat over 315 mixed-case, underscore, and multi-byte keys; a GSI added by `UpdateTable` to the old table backfilled and paged correctly (collated index table fed from old-collation base columns); new tables get `C`, old ones keep `default`; no `could not determine which collation` error in any log. A database whose default collation is already `C` behaves the same.

```
cargo fmt --all -- --check
cargo test --workspace                                 # 1,200 passed
devtools/run-tests --extenddb --pytest --rust-integration --parallel --backend postgres   # rust 511 passed; pytest failures were the import/export allowed-path environment, the console suites under the sandbox CA variables (41 pass with them unset), and the CLI migration suite's role lacking CREATEDB on this container
```

Found on the way and not changed here: reverse pagination on a hash-only GSI returns a repeated item on both PostgreSQL and SQLite (the cursor compares with `>` in both directions; six items, `Limit=2`, `ScanIndexForward=false` yields `p5, p4, p5`); a key containing a NUL byte returns `InternalServerError` on PostgreSQL where SQLite stores it; the reverse GSI cursor's expanded form reads from the top of the partition down to the cursor on each page. Each is pre-existing and reproduces on main.

## Checklist

ADR / RFC: n/a. The on-disk change is the collation attribute of key columns on newly created PostgreSQL tables; no wire, trait, auth, or CLI change.

## Breaking changes

None for clients. Tables created after this change have their key columns collated `C`; tables created before keep the database collation and are served unchanged (technical debt C-11 describes the migration that would bring them over).

---


crates/storage-postgres/src/table_engine.rs
```diff
@@ -62,7 +62,7 @@ impl TableEngine for PostgresEngine {
 
             let rows: Vec<(String,)> = if let Some(ref start) = input.exclusive_start_table_name {
                 sqlx::query_as(
-                    "SELECT table_name FROM tables WHERE account_id = $1 AND table_name > $2 ORDER BY table_name COLLATE \"C\" LIMIT $3",
+                    "SELECT table_name FROM tables WHERE account_id = $1 AND table_name COLLATE \"C\" > $2 ORDER BY table_name COLLATE \"C\" LIMIT $3",
```
robinnsc, line 65: The catalog `table_name` column has the same gap this PR fixes for data tables. Making the filter and order agree is the correctness fix, but both now run under `COLLATE "C"` while the column and its primary-key index stay on the database collation, so every page sorts the account's table names instead of seeking. Cardinality makes this harmless in practice, but it's exactly the pattern `TEXT_KEY` addresses, and fresh installs could get `table_name TEXT COLLATE "C"` in 001_schema.sql the same way. Either that, or extend C-11 to mention the catalog column so it isn't lost.

crates/storage-postgres/tests/key_collation.rs
```diff
@@ -0,0 +1,416 @@
...
+        let Scratch {
+            engine,
+            db,
+            admin,
+            db_name,
+        } = self;
+        drop(engine);
+        db.close().await;
+        sqlx::query(&format!(
+            "DROP DATABASE IF EXISTS \"{db_name}\" WITH (FORCE)"
```
robinnsc, line 54: `DROP DATABASE ... WITH (FORCE)` needs PostgreSQL 13+. If the backend's supported floor is below 13, the scratch harness fails on exactly the setups that need testing most. If 13+ is already the project's minimum, ignore; otherwise terminate backends first and drop without FORCE.

crates/storage-postgres/src/data/ddl.rs
```diff
@@ -11,6 +11,24 @@ use extenddb_storage::error::StorageError;
...
+/// `ORDER BY`. Declaring the columns with the same collation makes the
+/// primary key and the ordering indexes serve those predicates; a column left
+/// on the database collation (`en_US.utf8` in the default setups) forces
+/// PostgreSQL to seek on `pk` alone and filter the sort key afterwards,
+/// which re-reads a partition from its start on every page.
+///
+/// Tables created before this constant existed keep the database collation
+/// on these columns. Their results are still ordered correctly, because the
+/// predicates carry the collation explicitly; only the index use differs.
+const TEXT_KEY: &str = "TEXT COLLATE \"C\"";
```
robinnsc, line 31: Nit: this const is declared in the middle of the `use` block — between `use super::...` and `use crate::PostgresEngine`. Moving it below the imports keeps the header conventional.

robinnsc (approved):

PR #362, opened 2026-09-17
Index page cursors follow ScanIndexForward on every column

LeeroyHannigan:
## What

PostgreSQL and SQLite index page cursors follow `ScanIndexForward` on every column, and the `ORDER BY` and cursor columns are derived from one list per backend.

- `crates/storage-postgres/src/data/query_scan.rs`: new `order_by_columns` returns the ordered, collated column list for a shape (base table: sort key; LSI: index sort key, base sort key; GSI with a sort key: index sort key, `base_pk`, base sort key; hash-only index: `base_pk`, base sort key). `build_pagination_where` emits `(cols) > ($n..)` or `(cols) < ($n..)` from that list; the `ORDER BY` block applies `ASC` or `DESC` to every column of the same list. Placeholder positions are unchanged, so the bind construction is untouched.
- `crates/storage-sqlite/src/data/query_scan.rs`: the same shape as `query_order_columns` and `append_query_pagination`, with the binds collected in column order and one row comparison emitted. A unit test module is added (the file had none).
- `tests/test_query_scan.py`: `TestIndexPaginationMirrorsAndNeverSkips`.
- MongoDB is unchanged: its sort document and cursor comparator already flipped together on every column.

Stacked on #361 (`fix/postgres-key-collation`), which rewrote the same builder; GitHub retargets to `main` when #361 merges.

## Why

Two defects, one cause. A reverse Query on a GSI with only a partition key repeated the first page and skipped the rest: the page was ordered `base_pk DESC` while its cursor compared `base_pk > cursor`, so with six items and `Limit 2` a client saw `p5, p4`, then `p5`, then nothing. Reproduced on PostgreSQL and SQLite servers built from the parent (Limit 1 over 13 items: 1 delivered, 12 skipped; Limit 7: 6 duplicates, 6 skipped).

On a GSI with a sort key the two SQL backends kept the base-key tie-breaker ascending while the sort key descended. That paginated without skips but was not the mirror of the forward read. Measured 2026-09-17 against DynamoDB: a reverse Query is the exact mirror of the forward one at every level, including among items that share an index sort key and among the items of a hash-only index partition; the order inside a tie run or a hash-only partition is deterministic (three identical reads per direction on three tables) and is neither UTF-8 byte order nor insertion order, so it is unspecified. ExtendDB keeps its byte-order tie-breaker and now mirrors it; the three backends return the same forward and reverse order for the same data.

Applying the direction to every column also makes the reverse GSI cursor a row comparison, closing the follow-up noted on #358. `EXPLAIN (ANALYZE, BUFFERS)` on a 20,000-row index partition: the reverse GSI page is an `Index Scan Backward` with the row comparison in the `Index Cond`, 105 buffers for a 101-row page; the parent's OR form read from the partition top through a filter (10,000 rows removed, 10,263 buffers) into an incremental sort. On a 50,000-row partition the review measured 104 buffers against 25,131 and 2.5 ms against 24.9 ms per page. Every cursor shape on both backends plans as a single index seek with no sort node, except the LSI reverse cursor, which is unchanged by this PR and sorts a LIMIT-bounded group on both builds (the physical index carries `base_pk` between the two ordered columns).

## Behavior change

A client holding a `LastEvaluatedKey` from a reverse GSI Query issued by an earlier build resumes at a different position inside the current tie run after this deploys: no error, and at most the items of that one tie run are re-seen or missed once. Forward reads, LSI reads, base-table reads, Scan, and MongoDB are byte-identical to the parent on every probe.

## Testing done

Unit (`extenddb-storage-postgres` 8, `extenddb-storage-sqlite` 7): the fragment and bind order for every shape in both directions, and a check that the `ORDER BY` and cursor column lists agree for all seven shapes. Restoring the parent's builder bodies under these tests fails 4 (PostgreSQL) and 6 (SQLite).

Wire (`tests/test_query_scan.py::TestIndexPaginationMirrorsAndNeverSkips`, 28 cases): forward read deterministic, reverse the exact mirror, cursor walks at Limit 1, 2, 3, 5 in both directions equal to the unpaged read, on a hash-only GSI over a hash-only base, a hash-only GSI over a composite base, and a GSI whose sort key has five-way ties. Against DynamoDB: 28 passed. Against servers built from this branch: 28 on each backend. Against servers built from the parent: 10 failed on PostgreSQL and on SQLite (8 reverse hash-only walks, 2 tie-order mirrors), 28 passed on MongoDB. A build that fixed only the hash-only shape would fail exactly the 2 tie-order cases.

Further probes: S, N, and B base sort keys, N base partition keys, N index sort keys with negative, fractional, and 1E+30 values, LSI tie runs, range conditions (`BETWEEN`, `begins_with`, `<`, `>=`) with reverse cursor walks at Limit 1, mid-tie resumes in both directions, all three backends bit-identical and skip-free; reverse pagination under concurrent inserts before, after, and inside the current tie run, deletion of the live `LastEvaluatedKey` item, and sort keys moved across the cursor (48 stable items seen exactly once on every backend); malformed `ExclusiveStartKey` responses identical to the parent; row-comparison equivalence to the directed OR expansion over a NULL-including domain (4,480 combinations on PostgreSQL, 160 on SQLite, 0 disagreements); 20-page reverse walks of a 50,000-row partition 11 percent faster on PostgreSQL, 7 percent on SQLite, unchanged on MongoDB.

Gates: fmt, clippy, 1,209 workspace tests, MSRV 1.88, `tests/test_query_scan.py` 116 passed on each backend, the GSI and index integration files 142 passed on the SQL backends and 130 on MongoDB (remaining failures are environment: the `./target/release/extenddb` path and the CLI tests' connection string).

Closes the reverse-cursor follow-up on #358.


crates/storage-sqlite/src/data/query_scan.rs
```diff
@@ -352,91 +391,51 @@ fn append_query_pagination(
...
 ) -> Result<(), StorageError> {
-    let cmp = if forward { ">" } else { "<" };
-
+    let cols = query_order_columns(sk_info_val, base_sk_info, is_index, is_lsi);
+    if cols.is_empty() {
+        return Ok(());
+    }
+    // A missing sort-key attribute in the start key binds as the empty
+    // string, which sorts before every other text value.
+    let sk_bind = |name: &str, ty: ScalarAttributeType| -> Result<BoundValue, StorageError> {
```
robinnsc, line 400: Non-blocking question: a start key missing a sort-key attribute binds as `Text("")`. That's the pre-existing convention, but the new `missing_sort_key_attributes_bind_as_the_empty_string` test now pins it as intended, so worth settling: is this path reachable past ExclusiveStartKey validation? If it is, binding empty text against an N-typed column has a sharp edge in SQLite — text sorts above every numeric, so a forward resume from such a cursor skips all remaining numeric ties rather than starting before them. If validation upstream guarantees the attributes are present, a comment saying the fallback only preserves placeholder count for malformed-input paths would close the question.

robinnsc (changes requested):
One blocking item that can't be anchored inline because the file isn't in this diff: `docs/design/04-component-storage.md`. The parent PR (#361) updated it to say the reverse GSI cursor keeps the expanded OR form with an ascending tie-breaker, and this PR makes exactly that cursor a row comparison with every column following the direction. Nothing in this diff touches the doc, so when the stack merges the design doc will describe the form this PR deleted. Please update the reverse-query example and the surrounding prose here.

crates/storage-sqlite/src/data/query_scan.rs
jcshepherd: Tracking as https://github.com/ExtendDB/extenddb/issues/370

jcshepherd (approved):
Approving, with related document change required in https://github.com/ExtendDB/extenddb/issues/370

